PR #4, opened 2024-07-19
Add Context - resolves #1

### LINKED ISSUE #1: Add Context
Let's add a smallest viable `Context` to the repo, ported from eosim. We're going to need to discuss a few things like whether we want this components/plugin distinction etc., which I figure we can do in the PR.

jasonasher:
This imports the code from `eosim` for Context.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -174,30 +174,32 @@ impl Context {
     }
 
     pub fn execute(&mut self) {
-        // Execute callbacks if there are any in the queue
-        loop {
-            let callback = self.callback_queue.pop_front();
-            match callback {
-                Some(callback) => callback(self),
-                None => break,
-            }
-        }
         // Start plan loop
         loop {
-            let timed_plan = self.plan_queue.get_next_timed_plan();
-            match timed_plan {
-                Some(timed_plan) => {
-                    self.current_time = timed_plan.time;
-                    (timed_plan.callback)(self);
-                    loop {
-                        let callback = self.callback_queue.pop_front();
-                        match callback {
-                            Some(callback) => callback(self),
-                            None => break,
-                        }
-                    }
+            let mut executed = false;
+
+            // First execute callbacks.
+            loop {
+                let callback = self.callback_queue.pop_front();
+                match callback {
+                    Some(callback) => {
+                        executed = true;
+                        callback(self)
+                    },
+                    None => break,
                 }
-                None => break,
+            }
+
+            // Now execute the first timed plan.
+            if let Some(timed_plan) = self.plan_queue.get_next_timed_plan() {
+                executed = true;
+                self.current_time = timed_plan.time;
+                (timed_plan.callback)(self);
+            }
+
+            // If nothing happened, then we are done.
+            if !executed {
+                break;
             }
         }
     }
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,0 +1,354 @@
+use std::{
+    any::{Any, TypeId},
+    cmp::Ordering,
+    collections::{BinaryHeap, HashMap, HashSet, VecDeque},
+    rc::Rc,
+};
+
+use derivative::Derivative;
+
+pub trait Component: Any {
+    fn init(context: &mut Context);
+}
+
+pub trait Plugin: Any {
+    type DataContainer;
+
+    fn get_data_container() -> Self::DataContainer;
+}
+
+#[macro_export]
+macro_rules! define_plugin {
+    ($plugin:ident, $data_container:ty, $default: expr) => {
+        struct $plugin {}
+
+        impl $crate::context::Plugin for $plugin {
+            type DataContainer = $data_container;
+
+            fn get_data_container() -> Self::DataContainer {
+                $default
+            }
+        }
+    };
+}
+pub use define_plugin;
+
+pub struct PlanId {
+    pub id: u64,
+}
+
+#[derive(Derivative)]
+#[derivative(Eq, PartialEq, Debug)]
+pub struct TimedPlan {
+    pub time: f64,
+    plan_id: u64,
+    #[derivative(PartialEq = "ignore", Debug = "ignore")]
+    pub callback: Box<dyn FnOnce(&mut Context)>,
+}
+
+impl Ord for TimedPlan {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let time_ordering = self.time.partial_cmp(&other.time).unwrap().reverse();
+        if time_ordering == Ordering::Equal {
+            // Break time ties in order of plan id
+            self.plan_id.cmp(&other.plan_id).reverse()
+        } else {
+            time_ordering
+        }
+    }
+}
+
+impl PartialOrd for TimedPlan {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+#[derive(Debug)]
+struct PlanQueue {
+    queue: BinaryHeap<TimedPlan>,
+    invalid_set: HashSet<u64>,
+    plan_counter: u64,
+}
+
+impl PlanQueue {
+    pub fn new() -> PlanQueue {
+        PlanQueue {
+            queue: BinaryHeap::new(),
+            invalid_set: HashSet::new(),
+            plan_counter: 0,
+        }
+    }
+
+    pub fn add_plan(&mut self, time: f64, callback: impl FnOnce(&mut Context) + 'static) -> PlanId {
+        // Add plan to queue and increment counter
+        let plan_id = self.plan_counter;
+        self.queue.push(TimedPlan {
+            time,
+            plan_id,
+            callback: Box::new(callback),
+        });
+        self.plan_counter += 1;
+        PlanId { id: plan_id }
+    }
+
+    pub fn cancel_plan(&mut self, id: PlanId) {
+        self.invalid_set.insert(id.id);
+    }
+
+    pub fn get_next_timed_plan(&mut self) -> Option<TimedPlan> {
+        loop {
+            let next_timed_plan = self.queue.pop();
+            match next_timed_plan {
+                Some(timed_plan) => {
+                    if self.invalid_set.contains(&timed_plan.plan_id) {
+                        self.invalid_set.remove(&timed_plan.plan_id);
+                    } else {
+                        return Some(timed_plan);
+                    }
+                }
+                None => {
+                    return None;
+                }
+            }
+        }
+    }
+}
+
+type Callback = dyn FnOnce(&mut Context);
+type EventHandler<E> = dyn Fn(&mut Context, E);
+pub struct Context {
+    plan_queue: PlanQueue,
+    callback_queue: VecDeque<Box<Callback>>,
+    event_handlers: HashMap<TypeId, Box<dyn Any>>,
+    immediate_event_handlers: HashMap<TypeId, Box<dyn Any>>,
+    plugin_data: HashMap<TypeId, Box<dyn Any>>,
+    time: f64,
+}
+
+impl Context {
+    pub fn new() -> Context {
+        Context {
+            plan_queue: PlanQueue::new(),
+            callback_queue: VecDeque::new(),
+            event_handlers: HashMap::new(),
+            immediate_event_handlers: HashMap::new(),
+            plugin_data: HashMap::new(),
+            time: 0.0,
+        }
+    }
+
+    pub fn add_plan(&mut self, time: f64, callback: impl FnOnce(&mut Context) + 'static) -> PlanId {
+        // TODO: Handle invalid times (past, NAN, etc)
+        self.plan_queue.add_plan(time, callback)
+    }
+
+    pub fn cancel_plan(&mut self, id: PlanId) {
+        self.plan_queue.cancel_plan(id);
+    }
+
+    pub fn queue_callback(&mut self, callback: impl FnOnce(&mut Context) + 'static) {
+        self.callback_queue.push_back(Box::new(callback));
+    }
+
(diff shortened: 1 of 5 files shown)


src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+                    let internal_callback = Rc::clone(callback);
+                    callbacks_to_return
+                        .push(Box::new(move |context| internal_callback(context, event)));
+                }
+            }
+        }
+        callbacks_to_return
+    }
+
+    pub fn release_event<E: Copy + 'static>(&mut self, event: E) {
```
k88hudson-cfa, line 243: Is the event emitter stuff new? We may way to split this out

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+        // Execute callbacks if there are any in the queue
+        loop {
+            let callback = self.callback_queue.pop_front();
+            match callback {
+                Some(callback) => callback(self),
+                None => break,
+            }
+        }
+        // Start plan loop
+        loop {
```
k88hudson-cfa, line 264: I think this could use tests/we could maybe find a way to do the callback draining better, but probably better as a follow up

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    pub fn cancel_plan(&mut self, id: PlanId) {
+        self.invalid_set.insert(id.id);
+    }
+
+    pub fn get_next_timed_plan(&mut self) -> Option<TimedPlan> {
+        loop {
+            let next_timed_plan = self.queue.pop();
+            match next_timed_plan {
```
bbbruce, line 102: Can .map be used here?

k88hudson-cfa (changes requested):
Can you rebase on main (which should remove the precommit changes) and also split out the event emitter stuff for now? I think after discussing it would be good to merge that separately.

Will do a more thorough pass after that's done

src/context.rs
jasonasher: Pulled out in latest commit

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+            Some(x) => Some(x),
+            None => panic!("Plugin data container of incorrect type"),
+        }
+    }
+
+    pub fn get_time(&self) -> f64 {
+        self.time
+    }
+
+    pub fn add_component<T: Component>(&mut self) {
```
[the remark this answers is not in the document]
jasonasher: Yes, removed for now as it's vestigial. The original use was that adding a Component did create a namespace for plans and other things related to the Component but with successive simplification it's nothing more than some common structure for the end user.

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+
+type Callback = dyn FnOnce(&mut Context);
+type EventHandler<E> = dyn Fn(&mut Context, E);
+pub struct Context {
+    plan_queue: PlanQueue,
+    callback_queue: VecDeque<Box<Callback>>,
+    event_handlers: HashMap<TypeId, Box<dyn Any>>,
+    immediate_event_handlers: HashMap<TypeId, Box<dyn Any>>,
+    plugin_data: HashMap<TypeId, Box<dyn Any>>,
+    time: f64,
```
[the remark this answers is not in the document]
jasonasher: Yes, done.

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+            .get_mut(type_id)
+            .unwrap()
+            .downcast_mut::<T::DataContainer>();
+        match data_container {
+            Some(x) => x,
+            None => panic!("Plugin data container of incorrect type"),
+        }
+    }
+
+    pub fn get_data_container<T: Plugin>(&self) -> Option<&T::DataContainer> {
```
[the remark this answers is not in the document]
jasonasher: Unfortunately, no. The standard library has this same kind of pattern because &T and &mut T are very different things (the code just makes it look like they could be handled similarly).

src/context.rs
```diff
@@ -0,0 +1,354 @@
...
+        let type_id = &TypeId::of::<T>();
+        if !self.plugin_data.contains_key(type_id) {
+            self.add_plugin::<T>();
+        }
+        let data_container = self
+            .plugin_data
+            .get_mut(type_id)
+            .unwrap()
+            .downcast_mut::<T::DataContainer>();
+        match data_container {
```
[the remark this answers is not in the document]
jasonasher: Yes, and I just moved to unwrap, even.

src/context.rs
```diff
@@ -0,0 +1,244 @@
...
+pub struct PlanId {
+    pub id: u64,
+}
+
+#[derive(Derivative)]
+#[derivative(Eq, PartialEq, Debug)]
+pub struct TimedPlan {
+    pub time: f64,
+    plan_id: u64,
+    #[derivative(PartialEq = "ignore", Debug = "ignore")]
```
k88hudson-cfa, line 40: Maybe add a comment about why this is needed

src/context.rs
```diff
@@ -0,0 +1,244 @@
...
+            fn get_data_container() -> Self::DataContainer {
+                $default
+            }
+        }
+    };
+}
+pub use define_plugin;
+
+pub struct PlanId {
+    pub id: u64,
```
jasonasher, line 32: Maybe this should not have pub visibility? People using Context should only work with PlanIds that are returned and not construct them directly.

k88hudson-cfa (changes requested):

src/context.rs
```diff
@@ -0,0 +1,244 @@
...
+            time,
+            plan_id,
+            callback: Box::new(callback),
+        });
+        self.plan_counter += 1;
+        PlanId { id: plan_id }
+    }
+
+    pub fn cancel_plan(&mut self, id: PlanId) {
+        self.invalid_set.insert(id.id);
```
k88hudson-cfa, line 91: As we discussed live, this should be structured in some way that it will panic or at least warn if you try to cancel a plan that has already run. I think you don't want to be in a scenario where there's a bug in your logic where you do the cancel too late but it just sort of fails silently.

jasonasher: Yes, good observation. I implemented this in one relatively easy way and added tests.

src/plan.rs
```diff
@@ -0,0 +1,139 @@
+use std::{
+    cmp::Ordering,
+    collections::{BinaryHeap, HashMap},
+};
+
+pub struct Id {
+    id: usize,
```
ekr-cfa, line 7: Why is this `usize`? I would probably use `u64`.

src/plan.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+pub struct Plan<T> {
+    pub time: f64,
+    pub data: T,
+}
+
+#[derive(PartialEq, Debug)]
+pub struct Record {
```
ekr-cfa, line 16:
```suggestion
struct Record {
```

I would also consider calling this `QueueEntry` or `Entry`

src/plan.rs
```diff
@@ -0,0 +1,139 @@
...
+impl Ord for Record {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let time_ordering = self.time.partial_cmp(&other.time).unwrap().reverse();
+        if time_ordering == Ordering::Equal {
+            // Break time ties in order of plan id
+            self.id.cmp(&other.id).reverse()
+        } else {
+            time_ordering
+        }
+    }
```
ekr-cfa, line 38: I think this would be cleaner as:
```suggestion
        match time_ordering { 
            Ordering::Equal => self.id.cmp(&other.id).reverse(),
            _ => time_ordering
        }
```

src/plan.rs
```diff
@@ -0,0 +1,139 @@
...
+        self.plan_counter += 1;
+        Id { id }
+    }
+
+    /// # Panics
+    ///
+    /// This function panics if you cancel a plan which has already
+    /// been cancelled or executed.
+    pub fn cancel_plan(&mut self, id: &Id) {
+        self.data_map.remove(&id.id).expect("Plan does not exist");
```
ekr-cfa, line 78: I would add a comment, as in:

`Delete the plan from the map, but leave it in the queue. It will be skipped
when we get to the deadline`.

src/plan.rs
```diff
@@ -0,0 +1,139 @@
+use std::{
```
ekr-cfa, line 1: I think this could use a block comment at the top describing the architecture.

src/plan.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// been cancelled or executed.
+    pub fn cancel_plan(&mut self, id: &Id) {
+        self.data_map.remove(&id.id).expect("Plan does not exist");
+    }
+
+    pub fn get_next_plan(&mut self) -> Option<Plan<T>> {
+        loop {
+            match self.queue.pop() {
+                Some(plan_record) => {
+                    if let Some(data) = self.data_map.remove(&plan_record.id) {
```
ekr-cfa, line 85: I would also add a comment here to explain what is happening.

src/plan.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// This function panics if you cancel a plan which has already
+    /// been cancelled or executed.
+    pub fn cancel_plan(&mut self, id: &Id) {
+        self.data_map.remove(&id.id).expect("Plan does not exist");
+    }
+
+    pub fn get_next_plan(&mut self) -> Option<Plan<T>> {
+        loop {
+            match self.queue.pop() {
+                Some(plan_record) => {
```
ekr-cfa, line 84: If you take my advice above, I would call this `entry`

ekr-cfa (changes requested):

ekr-cfa (changes requested):
I think this is basically ready, except for my comment about `get_data_container()` which I think requires either a change or an argument about why I'm wrong.

Either way, we should have a test that exercises calling `get_data_container()` when one doesn't exist.

src/context.rs
```diff
@@ -0,0 +1,336 @@
...
+        }
+    };
+}
+pub use define_data_plugin;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    define_data_plugin!(ComponentA, Vec<u32>, vec![]);
```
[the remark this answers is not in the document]
jasonasher: Thanks, I worked hard to get those right.

src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+//! together to provide all of the functions of the simulation. For instance,
+//! For instance, a simple disease transmission model might consist of the
+//! following modules:
+//! * A population loader that initializes the set of people represented
+//!   by the simulation.
+//! * An infection seeder that introduces the pathogen into the population.
+//! * A disease progression manager that transitions infected people through
+//!   stages of disease until recovery.
+//! * A transmission manager that models the process of an infected
+//!   person trying to infect susceptible people in the population.
```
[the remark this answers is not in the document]
jasonasher: I put it here so it would show up in the generated Rust docs for the library. Will remove if you feel strongly.

src/context.rs
```diff
@@ -0,0 +1,336 @@
...
+            .downcast_mut::<T::DataContainer>()
+            .unwrap() // Will never panic as data container has the matching type
+    }
+
+    /// Retrieve a reference to the data container associated with a
+    /// `DataPlugin`
+    ///
+    /// Returns a reference to the data container if it exists or else `None`
+    #[must_use]
+    pub fn get_data_container<T: DataPlugin>(&self) -> Option<&T::DataContainer> {
```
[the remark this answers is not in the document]
jasonasher: The idea is that we want this method to return a reference to the data container in the current `Context` if it exists. If it doesn't there isn't a way to return a reference as nothing would own the data container. Circumventing this would entail interior mutability which I think for now we want to avoid.

src/context.rs
```diff
@@ -0,0 +1,336 @@
...
+                (plan.data)(self);
+            } else {
+                // OK, there aren't any plans, so we're done.
+                break;
+            }
+        }
+    }
+}
+
+impl Default for Context {
```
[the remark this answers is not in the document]
jasonasher: I think it's idiomatic to follow the struct with its impl before doing other traits, so I don't think there is a way to move this up more and keep with that.

src/context.rs
```diff
@@ -0,0 +1,336 @@
...
+        impl $crate::context::DataPlugin for $plugin {
+            type DataContainer = $data_container;
+
+            fn create_data_container() -> Self::DataContainer {
+                $default
+            }
+        }
+    };
+}
+pub use define_data_plugin;
```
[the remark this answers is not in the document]
jasonasher: I wanted to define Context first as it was the main focus of this source file. It seemed natural to have the DataPlugin defined after that together with its construction macro.

src/plan.rs
```diff
@@ -204,6 +204,29 @@ mod tests {
...
+    fn add_and_get_plans() {
+        let mut plan_queue = Queue::new();
+        plan_queue.add_plan(1.0, 1);
+        plan_queue.add_plan(2.0, 2);
+
+        let next_plan = plan_queue.get_next_plan().unwrap();
+        assert_eq!(next_plan.time, 1.0);
+        assert_eq!(next_plan.data, 1);
+
+        plan_queue.add_plan(3.0, 3);
```
ekr-cfa, line 217: I would consider adding it with 1.5 to demonstrate that sort works properly.

ekr-cfa (approved):
LGTM, but I added one comment below.

src/context.rs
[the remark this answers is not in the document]
ekr-cfa: I think there is a philosophical question here. In general, I think that things should appear before things that depend on them, or in some separate location and be imported in. In this case, neither is true. The question is whether it's possible to infer what DataPlugin does from the code that uses it, and I am skeptical of that.

k88hudson-cfa (approved):
🚢 it

PR #7, opened 2024-07-29
Run pedantic clippy in CI

natemcintosh:
I often find the clippy `pedantic` argument to be helpful. Feel free to ignore and close this PR.

This *would* be better as a pre-commit, but I could not for the life of me figure out how to pass in the `-W clippy::pedantic` argument to the pre-commit config; it just kept skipping over it, or saying it got an unexpected argument. I figured "better late (in the PR process) than never" applied here.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -18,3 +18,5 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Run pedantic clippy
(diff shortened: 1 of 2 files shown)


k88hudson-cfa (changes requested):
I took a look at the Rust hooks repo, did you try `args: ["--", "-D", "warnings", "-W", "clippy::pedantic"]`?

natemcintosh:
That did it!

k88hudson-cfa (approved):
Thanks for the PR! I think we can give this a try, we can always get rid of it if it's too much

PR #18, opened 2024-08-08
Added random module

k88hudson-cfa:
This PR adds a module to support creating and using random generators. It's implemented as a trait on context, with a `DataPlugin` to store a base random seed and a hashmap of rng instances. The design supports any rng implementation that implements `rand::SeedableRandom`.

The basic idea is you initialize the utility with a base random seed when setting up a `context`:
```rust
context.set_base_random_seed(12345678);
```
In a module that needs an independent random number generator, you use a macro to define an independent type, and then call `context.get_rng::<YourType>()` to retrieve it. For example:

```rust
define_rng!(FooRng)

let mut rng = context.get_rng::<FooRng>();
let value = rng.next_u64();
```
Note that an `rng` needs a mutable reference because it has to store internal state about previous calls in order to be reproducible.

One difference from the eosim implementation here is that I destroy existing rngs if `base_seed` is changed rather than using a flag on the holder.

### DIFF
--- a/src/random.rs
+++ b/src/random.rs
@@ -1,5 +1,6 @@
 use crate::context::Context;
-use rand::SeedableRng;
+use rand::prelude::Distribution;
+use rand::{Rng, SeedableRng};
 use std::any::{Any, TypeId};
 use std::cell::{RefCell, RefMut};
 use std::collections::HashMap;
@@ -58,11 +59,41 @@ crate::context::define_data_plugin!(
     }
 );
 
+/// Gets a mutable reference to the random number generator associated with the given
+/// `RngId`. If the Rng has not been used before, one will be created with the base seed
+/// you defined in `init`. Note that this will panic if `init` was not called yet.
+fn get_rng<R: RngId + 'static>(context: &Context) -> RefMut<R::RngType> {
+    let data_container = context
+        .get_data_container::<RngPlugin>()
+        .expect("You must initialize the random number generator with a base seed");
+
+    let rng_holders = data_container.rng_holders.try_borrow_mut().unwrap();
+    RefMut::map(rng_holders, |holders| {
+        holders
+            .entry(TypeId::of::<R>())
+            // Create a new rng holder if it doesn't exist yet
+            .or_insert_with(|| {
+                let base_seed = data_container.base_seed;
+                let seed_offset = fxhash::hash64(R::get_name());
+                RngHolder {
+                    rng: Box::new(R::RngType::seed_from_u64(base_seed + seed_offset)),
+                }
+            })
+            .rng
+            .downcast_mut::<R::RngType>()
+            .unwrap()
+    })
+}
+
 // This is a trait exension on Context
 pub trait ContextRandomExt {
     fn init_random(&mut self, base_seed: u64);
 
-    fn get_rng<R: RngId + 'static>(&self) -> RefMut<R::RngType>;
+    fn sample<R: RngId + 'static, T>(&self, sampler: impl FnOnce(&mut R::RngType) -> T) -> T;
+
+    fn sample_distr<R: RngId + 'static, T>(&self, distribution: impl Distribution<T>) -> T
+    where
+        R::RngType: Rng;
 }
 
 impl ContextRandomExt for Context {
@@ -77,39 +108,35 @@ impl ContextRandomExt for Context {
         rng_map.clear();
     }
 
-    /// Gets a mutable reference to the random number generator associated with the given
-    /// `RngId`. If the Rng has not been used before, one will be created with the base seed
-    /// you defined in `init`. Note that this will panic if `init` was not called yet.
-    fn get_rng<R: RngId + 'static>(&self) -> RefMut<R::RngType> {
-        let data_container = self
-            .get_data_container::<RngPlugin>()
-            .expect("You must initialize the random number generator with a base seed");
+    /// Gets a random sample from the random number generator associated with the given
+    /// `RngId` by applying the specified sampler function. If the Rng has not been used
+    /// before, one will be created with the base seed you defined in `set_base_random_seed`.
+    /// Note that this will panic if `set_base_random_seed` was not called yet.
+    fn sample<R: RngId + 'static, T>(&self, sampler: impl FnOnce(&mut R::RngType) -> T) -> T {
+        let mut rng = get_rng::<R>(self);
+        sampler(&mut rng)
+    }
 
-        let rng_holders = data_container.rng_holders.try_borrow_mut().unwrap();
-        RefMut::map(rng_holders, |holders| {
-            holders
-                .entry(TypeId::of::<R>())
-                // Create a new rng holder if it doesn't exist yet
-                .or_insert_with(|| {
-                    let base_seed = data_container.base_seed;
-                    let seed_offset = fxhash::hash64(R::get_name());
-                    RngHolder {
-                        rng: Box::new(R::RngType::seed_from_u64(base_seed + seed_offset)),
-                    }
-                })
-                .rng
-                .downcast_mut::<R::RngType>()
-                .unwrap()
-        })
+    /// Gets a random sample from the specified distribution using a random number generator
+    /// associated with the given `RngId`. If the Rng has not been used before, one will be
+    /// created with the base seed you defined in `set_base_random_seed`.
+    /// Note that this will panic if `set_base_random_seed` was not called yet.
+    fn sample_distr<R: RngId + 'static, T>(&self, distribution: impl Distribution<T>) -> T
+    where
+        R::RngType: Rng,
+    {
+        let mut rng = get_rng::<R>(self);
+        distribution.sample::<R::RngType>(&mut rng)
     }
 }
 
 #[cfg(test)]
 mod test {
     use crate::context::Context;
+    use crate::define_data_plugin;
     use crate::random::ContextRandomExt;
     use rand::RngCore;
-    use rand_distr::{Distribution, Exp};
+    use rand::{distributions::WeightedIndex, prelude::Distribution};
 
     define_rng!(FooRng);
     define_rng!(BarRng);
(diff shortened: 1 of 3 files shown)


k88hudson-cfa:
So I gave explicit initialization a try (implementation here if you're curious: [k88hudson_rng_explcit_init](https://github.com/CDCgov/ixa/compare/k88hudson_rng_explicit_init)), which ends up looking something like this:

In your init function of the module, you call `create_rng`:
```rust
init() {
  // ...
  context.create_rng::<FooRng>();
}
```
And then later you call `get_rng` as normal:
```rust
some_other_method() {
  let mut foo_rng = context.get_rng::<FooRng>();

  foo_rng.next_u64();
  // ...
  foo_rng.next_u64();
}
```

The thing I don't like about this is that if you want to reset the base random seed, you have to deal with reinitialization (like maybe iterating through existing ones and replacing them?), which seems kind of not great

k88hudson-cfa:
I spent a bunch of time but I couldn't figure out how to get this to work with interior mutability for the `RngHolder`s which also having the whole `HashMap` in a `RefCell`, so submitting as is

ekr-cfa:
> I spent a bunch of time but I couldn't figure out how to get this to work with interior mutability for the `RngHolder`s which also having the whole `HashMap` in a `RefCell`, so I think I'm just gonna leave it as is for now

Should this say "without also"?

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+        base_seed: 0,
+        rng_holders: RefCell::new(HashMap::new()),
+    }
+);
+
+#[allow(clippy::module_name_repetitions)]
+pub trait RandomContext {
+    fn set_base_random_seed(&mut self, base_seed: u64);
+
+    fn get_rng<R: RngId>(&self) -> RefMut<'_, R::RngType>;
```
ekr-cfa, line 55: Do you definitely need the <'_>. Not all RefMut uses do:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=08fbd68290076f81c190a8e6ebcb091e

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+impl RandomContext for Context {
+    /// Initializes the `RngPlugin` data container to store rngs as well as a base
+    /// seed. Note that rngs are created lazily when `get_rng` is called.
+    fn set_base_random_seed(&mut self, base_seed: u64) {
```
ekr-cfa, line 61: I think it's a bit ambiguous from the name of this function whether reseeding with the same value resets the PRNG state (I see from your tests you expect it to, and that's what it does). Perhaps we should call it `seed()` or even `init()` to make clear that reseeding with the same value isn't a noop.

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Gets a mutable reference to the random number generator associated with the given
+    /// `RngId`. If the Rng has not been used before, one will be created with the base seed
+    /// you defined in `set_base_random_seed`. Note that this will panic if `set_base_random_seed` was not called yet.
+    fn get_rng<R: RngId>(&self) -> RefMut<'_, R::RngType> {
+        let data_container = self
+            .get_data_container::<RngPlugin>()
+            .expect("You must initialize the random number generator with a base seed");
+
+        let rng_holders = data_container.rng_holders.try_borrow_mut().unwrap();
+
```
ekr-cfa, line 79: Extra blank line.

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[test]
+    fn get_rng_basic() {
+        let mut context = Context::new();
+        context.set_base_random_seed(42);
+
+        let mut foo_rng = context.get_rng::<FooRng>();
+        assert_eq!(foo_rng.next_u64(), 5113542052170610017);
+        assert_eq!(foo_rng.next_u64(), 8640506012583485895);
+        assert_eq!(foo_rng.next_u64(), 16699691489468094833);
```
ekr-cfa, line 115: Do you really want to hardcode these values? This will break if Rust changes the algorithm, which they say they might.

image

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[test]
+    #[should_panic(expected = "You must initialize the random number generator with a base seed")]
+    fn panic_if_not_initialized() {
+        let context = Context::new();
+        context.get_rng::<FooRng>();
+    }
+
+    #[test]
+    #[should_panic]
+    fn get_rng_one_ref_per_rng_id() {
```
ekr-cfa, line 127: Isn't this test misnamed because the rule it is testing is that you can't have two references to any two Rngs, as you are getting BarRng.

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+macro_rules! define_rng {
+    ($random_id:ident) => {
+        struct $random_id {}
+
+        impl $crate::random::RngId for $random_id {
+            // TODO: This is hardcoded to StdRng; we should replace this
+            type RngType = rand::rngs::StdRng;
+
+            fn get_name() -> &'static str {
+                stringify!($random_id)
```
ekr-cfa, line 19: Note that there are several ways in which one could end up having duplicate names. I wonder if we should store the type_id in the box and assert at runtime if someone tries to do `get_rng<T>` when `get_rng<U>` has already been called and T.get_name() == U.get_name()

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    #[test]
+    fn get_rng_two_types() {
+        let mut context = Context::new();
+        context.set_base_random_seed(42);
+
+        let mut foo_rng = context.get_rng::<FooRng>();
+        foo_rng.next_u64();
+        drop(foo_rng);
```
ekr-cfa, line 145: Maybe comment this drop to explain why it's different from the previous test.

ekr-cfa (changes requested):

k88hudson-cfa:
To summarize some discussion: one thing we should consider here is that we will mostly be using `rng`s in combination with some kind of distribution/sample function. These may be created in place by a module (like the `Exp` example below), and sometimes they might need to be stored as a global variable / in a data container (because they are pretty large / used in multiple places or whatever).

The design in this PR returns a `RefMut` you have to deal with that when you pass it to a sample function, which is maybe not ideal:

```rust
let mut foo_rng = context.get_rng::<FooRng>();
let dist = Exp::new(1 / bar).unwrap();
let value = dist.sample(&mut *foo_rng);
```
 
@jasonasher did some prototyping of an alternative design where instead of the context extension providing a `get_rng` method, it implements a `sample` that retrieves the rng internally:

```rust
let dist = context.get_data_container::<SamplerData>().unwrap();
let value = context.sample::<FooRng, usize>(|rng| dist.sample(rng));
// Aternatively, pass in the whole distribution
let value2 = context.sample_distr::<FooRng, usize>(dist);
```

I don't this hate honestly, if you're pretty much always using rngs this way

jasonasher:
In my experience the most common ways we use random sampling are either as distributions like this or by grabbing random f64/f32s to make future decisions, such as setting individual propensities to engage in various behaviors. So having a general `sample` method and then `sample_distr` and `sample_xyz` for various primitive `xyz` would cover all of the common use cases. I'd be interested to know if @confunguido agrees.

src/random.rs
```diff
@@ -0,0 +1,174 @@
...
+
+            fn get_name() -> &'static str {
+                stringify!($random_id)
+            }
+        }
+    };
+}
+pub use define_rng;
+
+pub trait RngId: Any {
```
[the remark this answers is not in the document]
jasonasher: I think you could get away with just `RngId: 'static` so that it has a TypeId, or move the static lifetime requirement into the `get_rng` / `sample` method signatures.

confunguido:
> In my experience the most common way we use random sampling is either with distributions like this or by grabbing random f64/f32s to make future decisions, such as setting individual propensities to engage in various behaviors. So having a general `sample` method and then `sample_distr` and `sample_xyz` for various primitive `xyz` would cover all of the common use cases. I'd be interested to know if @confunguido agrees.

@jasonasher. I think I agree. Sampling from distribution, making decisions, and grabbing a random individual from a group. As a side note,  I think it would be nice to have the ability to specify and sample from custom probability distributions.

src/random.rs
k88hudson-cfa: Alright, I added this to the macro which should ensure this actually fails at compile time:

```rust
pub static $random_id: () = ();
```

k88hudson-cfa:
Alright, I think this is ready: updated this to expose `sample` and `sample_distr` – also added something to the macro to prevent name collisions.

jasonasher:
> @jasonasher. I think I agree. Sampling from distribution, making decisions, and grabbing a random individual from a group. As a side note, I think it would be nice to have the ability to specify and sample from custom probability distributions.

@confunguido These would be supported right out of the box with sample_distr: [rand_distr](https://docs.rs/rand_distr/latest/rand_distr/). And, by implementing the Distribution<T> trait we can support custom distributions.

src/random.rs
jasonasher: The downside to the static variable solution is potential collisions with other types (which we could reduce by mangling the name a bit), and that it only works if people use the macro (which of course they should). Maybe the runtime solution is better as we would only check when `get_rng` is called below and when the generator needs to be created (basically on first sample after seeding/reseeing)?

k88hudson-cfa: I mean, do you actually want two types across to have the same name? It's true that runtime check is safer without the macro though, that's probably enough of a reason to prefer it

k88hudson-cfa: On second thought after implementing the runtime check, it is quite a bit more code / requires storing the name etc. and I lean towards leaving it as is – 

I pushed a version that namespaces the type name a bit (I couldn't figure out how to do it without importing `paste`)

k88hudson-cfa:
So @jasonasher came up with a solution in https://github.com/CDCgov/ixa/tree/jma_rand_runtime_collision for runtime collision checking  – the thing I don't like about it is that it's easy to forget to update the field that caches name collisions given the current architecture of data containers, so I think i'd still leave this as is (unless we really don't want to require `paste`)

jasonasher:
I don't think requiring `paste` is much of an issue - I was more reacting to handling people potentially not using the macro. My instinct is we don't have much of an update problem here for the name field caching, as it's handled in one place and not obviously likely to change much in the future, but I could be missing something and defer to your and @ekr-cfa's judgment here.

ekr-cfa:
I would prefer a compile time check even if it's imperfect. I think if people color this far outside the lines, then they can suffer.

ekr-cfa (approved):
LGTM

PR #19, opened 2024-08-08
Gce ixa basic example

confunguido:
This PR describes a basic infection model based on constant force of infection in a homogeneous population.


k88hudson-cfa:
This seems like a much better level of complexity on what's included in the model. I think what would be good to do now is work on fleshing out that section about the simulation overview into a description of the actual architecture. So for example, you could describe how the entry-point sets up the parameters/main execution loop for each simulation, which loads the following modules... (a population loader, a transmission manager, etc.) and then describe what each one does. Of course the division of what each module does is a matter of discussion and there are a lot of options here – start with what you'd expect based working on other models, and we'll go from there.

Once you've described that you can then get into the interactions around a person getting infected / the events getting fired and like diagram that out in terms of what happens in each module, rather than just describing the events in generic terms. Also for stochasticity – that would also be good to address in terms of how it shows up in the architecture (like that modules need their own independent rng or whatever)

examples/basic_infection/README.md
```diff
@@ -25,6 +22,20 @@ Infected individuals schedule their recovery at time `t + infected period`. The
...
+
+### Initialization and modules
+
+ - *parameters module*: this module manages data for parameter values for population size, force of infection, and infection duration. At initialization, all parameter variables are set to a specific value.
+ - *person module*: contains a unique ID for each person that identifies each person in the simulation.
+ - *person_infection_status module*: this module connects each person ID with a specific value for a person's infection status, which could be one of Susceptible, Infected, or Recovered.
+ - *population manager module*: The population manager module is in charge of setting up the population. At initialization, the module reads the population size (N) parameter and creates N persons with unique ids (1..N) and sets their initial infection status to Susceptible.
+ - *transmission module*: This module is in charge of spreading the infection through the population. At initialization, it schedules an infection attempt for time 0. An infection attempt consists of choosing at random a susceptible individual from the population and setting their infection status to **infected**.
```
jasonasher, line 34: Infection attempt events also schedule follow-on infection attempt events with exponential inter-arrival times (assuming you mean a Poisson point process here), so there is another parameter at play here too.

examples/basic_infection/README.md
```diff
@@ -25,6 +22,20 @@ Infected individuals schedule their recovery at time `t + infected period`. The
 flowchart LR
 S(Susceptible) --FoI--> I(Infected) --inf. period--> R(Recovered)
 ```
+## Simulation architecture
+The simulation executaion first initializes the model parameters and prepares the simulation modules.
+
+### Initialization and modules
```
jasonasher, line 28: I think it would be good for each of these modules to flesh out some more detail - enough so an engineer could assess what features Context needs to have, and to even be able to make good progress translating your outline to code.

So, you could describe what each module does upon initialization (if anything) in terms of registering for events, making plans, etc. Then for each of the function/closure handlers for those events and plans you should describe what happens in the body of that function/closure.

ekr-cfa: To follow on here, I think it's useful to understand the flow of data.

So, for instance, the transmission manager infects person X, and then the infection manager has to do something, right? How does that happen? I would also add reporting, as it illustrates some other pieces, especially if you want to do status reporting, not event reporting.

examples/basic_infection/README.md
```diff
@@ -25,6 +22,20 @@ Infected individuals schedule their recovery at time `t + infected period`. The
...
+ - *person module*: contains a unique ID for each person that identifies each person in the simulation.
+ - *person_infection_status module*: this module connects each person ID with a specific value for a person's infection status, which could be one of Susceptible, Infected, or Recovered.
+ - *population manager module*: The population manager module is in charge of setting up the population. At initialization, the module reads the population size (N) parameter and creates N persons with unique ids (1..N) and sets their initial infection status to Susceptible.
+ - *transmission module*: This module is in charge of spreading the infection through the population. At initialization, it schedules an infection attempt for time 0. An infection attempt consists of choosing at random a susceptible individual from the population and setting their infection status to **infected**.
+ - *infection manager module*: This module controls the infection status of each person. Specifically, it handles the progression of newly infected persons by scheduling recovery, which is drawn from an exponential distribution based on the infection duration parameter from the parameters module .
```
ekr-cfa, line 35: As I understand it, the infection manager manages the state of the infected person but does not try to infect others. I know you don't say otherwise, but this would be helpful to sharpen.

examples/basic_infection/README.md
confunguido: I elaborated a bit more with transmission and infection managers. I think they are a bit crude in the sense that some of the interactions with the context are still unclear. I will add more detail to the other modules but pushed for now to get some feedback.

confunguido: Also, yeah, I will draft something for the reporting module soon.

examples/basic_infection/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Infected individuals schedule their recovery at time `t + infected period`. The infection status of recovered individuals remains as recovered for the rest of the simulation.
+
+### Infection state transitions
+```mermaid
+flowchart LR
+S(Susceptible) --FoI--> I(Infected) --inf. period--> R(Recovered)
+```
+## Simulation architecture
+The simulation executaion first initializes the model parameters and prepares the simulation modules.
```
k88hudson-cfa, line 26: Let's discuss in person let's work through moving this from a list structure to something like: at the entry point of the model we read model parameters, initialize each module, set up loop etc. – then what gets kicked off first?Which module schedules infections (transmission manager )? How do they executed? What happens next / which data flows where? 

So the structure is like talking through how things flow through the system so you can put some of the details (like the methods you're describing, stochasticity etc.) in context. Also it's very acceptable to hand wave things you're nto sure about :)

examples/basic_infection/README.md
```diff
@@ -0,0 +1,103 @@
...
+- Module data: This modules doesn't hold any specific data
+- Dependencies:
+```
+    context, parameters, random number generator, person infection status, person, population manager
+```
+- Initialization:
+```
+    init(context) {
+        context.add_rng(id = transmission);
+        context.add_plan(attempt_infection(context), time = 0);
```
k88hudson-cfa, line 50: Ah, so basically the first thing that happens is you try to infect someone?

confunguido: Yes. Although, not sure that I put it in the best place. In this example, there isn't any transmission through contacts, but here it assumes that everything else has been setup, meaning the population has been set properly with the infection status as susceptible.

examples/basic_infection/README.md
```diff
@@ -0,0 +1,223 @@
...
+    context.observe_person_property_event(update_property_counter);
+    context.add_plan(report_periodic_item, time = 0);
+}
+```
+Methods are implemented for person property counter to increment and decrement the counters. For the periodic report, one method will release the periodic items to a report based on the period defined in parameters. Changes in the person properties are observed and the callback function `update_property_counter` updates the counts for each property.
+```
+fn update_property_counter(context, person_id,
+                           person_property_type,
+                           previous_person_property_value) {
+    person_property_counter.increment(context.get_person_property(person_property_type, person_id));
```
confunguido, line 178: I'm not specifying here how to implement the methods: increment, decrement, or `get_counter` for the `person_property_counter` because I think the important thing is that we want a counter that increments values for each person property type/value, but I can elaborate if necessary.

PR #26, opened 2024-08-20
Report Example for Multi-Threading

ranya-cfa:
This PR adds a report writing example for multi-threading, which will implement the report writing module.

You can run examples with cargo run --example reports-multi-threaded.

### DIFF
--- a/examples/reports-multi-threaded/main.rs
+++ b/examples/reports-multi-threaded/main.rs
@@ -18,16 +18,18 @@ fn main() {
             let people = vec!["1", "2", "3"];
             for person in people {
                 let person = person.to_string();
+                let scenario = scenario.clone();
                 context.add_plan(1.0, {
-                    let person = person.clone();
                     move |context| {
                         #[cfg(feature = "reports")]
                         context.send_report(Incidence {
+                            scenario: scenario,
                             person_id: person.clone(),
                             t: context.get_current_time(),
                         });
                         println!(
-                            "Person {} was infected at time {}.",
+                            "Scenario: {}, Person {} was infected at time {}.",
+                            scenario,
                             person,
                             context.get_current_time()
                         );
--- a/examples/reports-multi-threaded/main.rs
+++ b/examples/reports-multi-threaded/main.rs
@@ -1,6 +1,12 @@
 use ixa::context::Context;
 use std::thread;
 
+struct Incidence {
+    scenario: String,
+    person_id: String,
+    t: f64,
+}
+
 fn main() {
     let scenarios = vec!["Illinois", "Wisconsin", "Arizona", "California"];
     let mut handles = vec![];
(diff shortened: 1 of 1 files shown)


examples/reports-multi-threaded/main.rs
```diff
@@ -8,11 +8,33 @@ fn main() {
...
+            #[cfg(feature = "reports")]
+            context.add_report::<Incidence>("Incidence");
+            
             println!("Scenario: {}", scenario);
 
+            let people = vec!["1", "2", "3"];
+            for person in people {
+                let person = person.to_string();
+                context.add_plan(1.0, {
+                    let person = person.clone();
```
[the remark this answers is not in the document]
ranya-cfa: The clone is from when I was adding multiple plans. Since there's only one now, person doesn't need to be cloned. Removing that now

examples/reports-multi-threaded/main.rs
```diff
@@ -8,11 +8,33 @@ fn main() {
...
+                    move |context| {
+                        #[cfg(feature = "reports")]
+                        context.send_report(Incidence {
+                            person_id: person.clone(),
+                            t: context.get_current_time(),
+                        });
+                        println!(
+                            "Person {} was infected at time {}",
+                            person,
+                            context.get_current_time()
```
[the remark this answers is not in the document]
ranya-cfa: Is it better now?

examples/reports-multi-threaded/main.rs
```diff
@@ -8,11 +8,35 @@ fn main() {
...
             println!("Scenario: {}", scenario);
 
+            let people = vec!["1", "2", "3"];
+            for person in people {
+                let person = person.to_string();
+                let scenario = scenario.clone();
+                context.add_plan(1.0, {
+                    move |context| {
+                        #[cfg(feature = "reports")]
+                        context.send_report(Incidence {
```
confunguido, line 25: Do I understand correctly that the "Incidence " label here indicates that context can have multiple types of reports and this is the incidence version which expects `scenario, person_id, and t`?

k88hudson-cfa: Actually can you define the `struct` for this ranya? Yeah, I was thinking maybe like one report with a `disease_status` for this example (obviously you could do two a different report for each disease status instead like the other example)

ranya-cfa: Yes, that's what I was thinking that this is passing the incidence struct. What should the struct look like? In eosim, Jason defined incidence as this
`#[derive(Serialize)]
pub struct Infection {
    pub time: f64,
}`
and disease status was an enum
`pub enum DiseaseStatus {
    S,
    I,
    R,
}`

k88hudson-cfa: Actually I really don't think this matters too much for this example, so let's just assume `Incidence` is infections only for now and includes scenario, person_id, and t

ranya-cfa: Sounds good, I added that. Let me know what you think!

k88hudson-cfa (approved):
LGTM

PR #27, opened 2024-08-21
Update the reports explanation

k88hudson-cfa:
A few minor updates on the reports explanation for the basic-infections example, highlighting the need for previous/current snapshots and the timing of when periodic reporting must occur (i.e. at the end of all plans for a given t)


examples/basic-infection/README.md
```diff
@@ -183,82 +186,130 @@ infection status:
...
+1,Recovered,0
+...
 ```
-At initialization, the property counter is initialized with the current state of all the population. The report also observes changes in person properties which are handled by a function that updates counts in the property counter. The first report day is scheduled for t = 0.
+
+In this case, we could actually compute each daily summary in
+post-processing from the instantaneous reports instead of generating a second
+set of periodic reports. However, we want a summary of properties
+which are not otherwise recorded, such as perhaps whether an individual is
+hospitalized.
```
confunguido, line 253: Yes, currently hospitalized people would be great. I also think maybe we want to know how many susceptible people are at some point. If we have a model where there are births and deaths, it might be difficult to keep track of the actual number of susceptible individuals in post-processing, and maybe more difficult if we include categories (e.g., age).

examples/basic-infection/README.md
```diff
@@ -183,82 +186,130 @@ infection status:
...
-                person_property_counter = person_property_counter.get_counter(person_property, person_property_value),
-                }
-                context.print_report_data(report_item);
-            }
-        }
+When all plans have executed for a given time t, `ixa` calls the on_period_end
+callback to write the report row and schedule the next periodic report:
+
+```rust
+fn report_periodic_item(day, context) {
```
confunguido, line 299: Can day be t? I guess we don't really care about 0.5 days, but I'm trying to think whether there will be a case when someone may want a periodic report for less than a day. Perhaps if we have a vector-borne disease model where we want to keep track of number of infectious bites per hour or something like that?

examples/basic-infection/README.md
```diff
@@ -183,82 +186,130 @@ infection status:
...
+        ));
     }
 
     // Schedule next report day
-    next_report_time = context.get_time() + paramters.get_parameter(reporting_period);
-    if next_report_time < parameters.get_parameter(max_days) {
-        context.add_plan(report_periodic_item, next_report_time);
+    next_report_time = context::get_time() + parameters::get_parameter(reporting_period);
+    if next_report_time < parameters::get_parameter(max_days) {
+        context::plans_did_execute(report_periodic_item, report_periodic_item);
```
confunguido, line 312: what does `plan_did_excute` do?

k88hudson-cfa: Ah that's supposed to be `on_period_end`

examples/basic-infection/README.md
k88hudson-cfa: yeah for sure, or it could be longer

examples/basic-infection/README.md
```diff
@@ -296,20 +297,20 @@ When all plans have executed for a given time t, `ixa` calls the on_period_end
...
             count: counter::get_count(infection_status)
         ));
     }
 
-    // Schedule next report day
+    // Schedule next report
     next_report_time = context::get_time() + parameters::get_parameter(reporting_period);
-    if next_report_time < parameters::get_parameter(max_days) {
-        context::plans_did_execute(report_periodic_item, report_periodic_item);
+    if next_report_time < parameters::get_parameter(max_periods) {
```
confunguido, line 312: I don't think `max_periods` makes sense here.  My initial thought here was that if there is a maximum time specified by the user (e.g., 100 days), then reports should stop scheduling a next report time after 100 days. But that would mean that every time a periodic report is included, a `max_time` should also be specified.   

My understanding is that the context will execute until there are no more plans scheduled, but if `max_time` isn't specified, reports continue to be scheduled indefinitely. I didn't address this in my example, but I think it's important to determine **when should reports stop?**

confunguido (approved):

PR #29, opened 2024-08-27
Events

k88hudson-cfa:


src/context.rs
```diff
@@ -352,4 +400,28 @@ mod tests {
...
     }
+
+    #[derive(Copy, Clone)]
+    struct Event {
+        pub data: usize,
+    }
+
+    #[test]
+    fn test_events() {
+        let mut context = Context::new();
```
ekr-cfa, line 411: This code looks fine. I think, however, we probably need a few more tests. Specifically:

1. That if you have two event types, and subscribe to each, different handlers fire
2. That if you subscribe twice to the same event, both handlers fire
3. That if you do (a) queue a callback to fire event (b) queue a callback to register to an event (c) execute. The handler never fires.

ekr-cfa (approved):

PR #30, opened 2024-08-27
Simplify event handling using Option::take pattern

jasonasher:


ekr-cfa:
Merging so I can review it all at once.

PR #31, opened 2024-08-27
More tests for events

ekr-cfa:
I added the tests from my review.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -5,12 +5,16 @@
 use std::{
     any::{Any, TypeId},
     collections::{HashMap, VecDeque},
+    rc::Rc,
 };
 
 use crate::plan::{Id, Queue};
 
 /// The common callback used by multiple `Context` methods for future events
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -38,6 +42,7 @@ type Callback = dyn FnOnce(&mut Context);
 pub struct Context {
     plan_queue: Queue<Box<Callback>>,
     callback_queue: VecDeque<Box<Callback>>,
+    event_handlers: Option<HashMap<TypeId, Box<dyn Any>>>,
```
k88hudson-cfa, line 45: Can you improve the comments about why we need an Option

k88hudson-cfa (approved):
LGTM

src/context.rs
jasonasher: I replaced the use of Option here with destructuring.

ekr-cfa (commented):
Jason's changes look good.

PR #36, opened 2024-08-28
Test examples in CI

jasonasher:
This will have cargo run/test the examples in CI. We should probably also consider having clippy run on these examples too.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,4 +13,5 @@ repos:
     hooks:
     -   id: fmt
     -   id: clippy
+        # Use 'all-targets' to run on all code, including tests and examples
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -13,4 +13,4 @@ repos:
     hooks:
     -   id: fmt
     -   id: clippy
-        args: ["--", "-D", "warnings", "-W", "clippy::pedantic"]
+        args: ["--all-targets", "--", "-D", "warnings", "-W", "clippy::pedantic"]
```
k88hudson-cfa, line 16: Can you add a comment that clippy:pedantic won't apply to tests/examples unless you include `--all-targets`

k88hudson-cfa (approved):
R+ with nits

PR #39, opened 2024-08-28
Add priorities for plans

jasonasher:


ChiragKumar9:
At this point, we are holding off from making first and last plans unable to modify context. This is because doing so requires dividing context's information up into "info context uses internally", like for reports, and info that impacts the simulation. While a report (done with priority `Last`), may want to modify its internal report data, it shouldn't necessarily modify other data. However, we are holding off from implementing this division of data in context right now.

src/context.rs
```diff
@@ -16,6 +16,20 @@ type Callback = dyn FnOnce(&mut Context);
...
 
+/// An enum to indicate the priority for plans at a given time.
+///
+/// Most plans will have `Normal` priority. Plans with priority `First` are
+/// handled before all `Normal` plans, and those with priority `Last` are
+/// handled after all `Normal` plans. In all cases ties between plans at the
+/// same time and with the same priority are handled in the order of scheduling.
+///
+#[derive(PartialEq, Eq, Ord, PartialOrd)]
+pub enum PlanPriority {
```
k88hudson-cfa, line 27: I'm sorry to nitpick about words but I really don't like the term "Priority" here which implies like some unbounded arbitrary numbering system (which I don't think we want to generally encourage).  Can we do maybe `PlanPhase` or `ExecutionPhase` and refer to them in docs/comments as "phases"

src/context.rs
```diff
@@ -104,19 +118,38 @@ impl Context {
...
+    /// Add a plan to the future event list at the specified time and with the
+    /// specified priority (first, normal, or last among plans at the
+    /// specified time)
+    ///
+    /// Returns an `Id` for the newly-added plan that can be used to cancel it
+    /// if needed.
+    /// # Panics
+    ///
+    /// Panics if time is in the past, infinite, or NaN.
+    pub fn add_plan_with_priority(
```
k88hudson-cfa, line 142: with_phase

k88hudson-cfa (changes requested):
This looks pretty much good to me with the renaming. Sweet!

src/context.rs
ChiragKumar9: Yes! Totally fair. Switched to `ExecutionPhase` -- thanks!

ChiragKumar9:
Wait oops, I have to fix some stuff here before re-review.

k88hudson-cfa:
Looks good after a rebase 👍

ChiragKumar9:
I feel the power of the rebase coursing through my veins! Ready for re-review @k88hudson-cfa, thanks!

k88hudson-cfa (approved):
Nice work, go ahead and merge when you're ready 👍

Issue #2, opened 2024-07-19
Add hello world example


k88hudson-cfa:
Let's add little hello world demo in an sub-directory `examples/` directory once the basics are in place. 

Each demo in `examples/` should be in its own sub-directory with a little README describing what it's demonstrating.

PR #41, opened 2024-08-28
Example #2: Group-dependent time-varying force of infection

ChiragKumar9:
`README.md` specifying model and `groups.md` specifying my conception of the `ixa` `groups` component for a food-borne illness that has a time-varying force of infection (i.e., seasonality) that varies by region/group. Rust pseudocode for the model. Any and all feedback appreciated -- the goal here is to not just think about `ixa` functionality but also how `ixa` structure can drive modularity -- and others are also welcome to review.


examples/infection-by-group/README.md
```diff
@@ -0,0 +1,380 @@
...
+In `regions.rs`:
+
+```rust
+use strum_macros::EnumIter;
+use ixa::GroupMappings::*;
+
+enum GroupTypes {Counties}
+
+#[derive(Debug, EnumIter)]
+enum Counties {A, B}
```
k88hudson-cfa, line 135: Instead of defining `k` as a parameter, wouldn't it make more sense to have this derived from the total group types?

examples/infection-by-group/README.md
```diff
@@ -0,0 +1,380 @@
...
+    panic!(parameters.get_parameter(k) < parameters.get_paramter(population_size));
+    context.subscribe_to_event(PersonCreationEvent, assign_and_plan_movement);
+}
+
+// note there is an alternate implementation where the per-region movement rate is calculated,
+// and then a random person is selected within the region to move.
+fn assign_and_plan_movement(context: &mut Context,
+                            person_creation_event: PersonCreationEvent) {
+    // evenly distribute ppl among the counties
+    let county_index = parameters.get_parameter(population_size) % parameters.get_parameter(k);
```
k88hudson-cfa, line 155: Is there a reason you want to do this differently from moving them (i.e. iterate through instead of randomly distributing them on the initial assignment?)

I would imagine this would end up as either an `assign_random_group` function on `PersonBuilder` (if this is how we end up interacting with person properties) or this is a `assign_random_group` is a trait extension on context that takes a person id, and you set up the listener here.

examples/infection-by-group/README.md
```diff
@@ -0,0 +1,380 @@
...
+// Evaluate whether infection can happen analytically -- because we know foi(t) exactly
+// here's how the method works: foi(t) describes the hazard rate for infection at a
+// given time, so 1 - e ^ -\int_0^t foi(u)du is the cumulative probability of infection at t.
+// Therefore, infection has happened by time t with the probability described above.
+// Drawing a random number on U(0, 1) gives a probability for the infection happening when the
+// CDF hits that value. To find the time at which this occurs, solve for the value of t
+// that gives that value of the CDF.
+// The math behind this is the same as drawing s ~ Exp(1), setting that equal to \int_0^t foi(u)du,
+// and solving for the time t at which the infection happens, which is fewer steps/easier.
+fn inverse_sampling_infection(context: &mut Context, person_id: PersonID) {
```
k88hudson-cfa, line 241: This is nice as a contrast to the other basic infection model. I wonder if this actually ends up being its own mini example?

k88hudson-cfa (commented):
I think the main thing to figure out is how this interacts with person properties if at all. The answer kind of depends on if you want similar interfaces for like mutation listeners/indexes and stuff, which you probably do.

I like the idea of internally enforcing the rules around one to one, one to many etc. depending on what relationship you define. 

Might be worth providing some nice helper utilities for randomly assigning people into groups, but maybe that's overkill since it's not that hard

examples/infection-by-group/README.md
ChiragKumar9: This is now in #54.

examples/infection-by-group/README.md
ChiragKumar9: In GCM, we could only make subscription functions that took context and the event as an argument, and nothing else! So, we had to use this kind of pattern a lot. We have changed this, so updating! Thanks!

ChiragKumar9: Oh, I think I properly understand your point here. This construction ensures that the two populations start off exactly equal, rather than on average equal.

ekr-cfa:
@ChiragKumar9  can we close this now?

PR #43, opened 2024-08-29
Create an example for random

ranya-cfa:


examples/random/main.rs
```diff
@@ -0,0 +1,35 @@
...
+static POPULATION: u64 = 10;
+
+define_rng!(MyRng);
+
+fn main() {
+    let mut context = Context::new();
+    context.init_random(SEED);
+
+    let random_person = context.sample_range(MyRng, 0..POPULATION);
+    let person_id = random_person;
```
ChiragKumar9, line 16: I believe this line is residual from when this was being "stringed"

examples/random/main.rs
```diff
@@ -0,0 +1,35 @@
...
+        move |context| {
+            println!(
+                "Person {} was infected at time {}",
+                person_id,
+                context.get_current_time()
+            );
+        }
+    });
+
+    let recovery_time: f64 = context.sample_distr(MyRng, Uniform::new(2.0, 10.0));
```
ChiragKumar9, line 28: I wonder if it wouldn't hurt to put another example here of using the `sample_bool` method -- something like if you draw some bool with probability p that is true, the recovery time is drawn from Uniform(2, 10), but if it is false, the recovery time is drawn from some custom PDF using the sample function or just Exp(1) if we want it to be more straightforward. I just think it's nice to give people examples, even if the method is straightforward to use in this early stage.

examples/random/main.rs
```diff
@@ -0,0 +1,35 @@
...
+    context.add_plan(1.0, {
```
k88hudson-cfa, line 29: How about make the plan execute at `recovery_time` and then log `context.get_current_time()`?

examples/random/main.rs
k88hudson-cfa: I agree, let's work on integrating some more example usages

k88hudson-cfa (changes requested):

k88hudson-cfa (approved):

PR #44, opened 2024-08-30
Example for periodic reports

ranya-cfa:


examples/periodic-reports/main.rs
```diff
@@ -0,0 +1,90 @@
...
+fn schedule_daily_reports(context: &mut Context, interval: f64, current_day: u32, max_days: u32) {
+    if current_day >= max_days {
+        return;
+    }
+
+    context.add_plan(f64::from(current_day) + interval, move |context| {
+        let sim_state = context.get_data_container(SimulationStatePlugin).unwrap();
+        context.send_report(Incidence {
+            day: current_day,
+            infections: sim_state.total_infected,
```
confunguido, line 85: It would be nice to have this example with a model where we actually have to count number of people currently infected/susceptible/recovered.

examples/periodic-reports/main.rs
```diff
@@ -0,0 +1,90 @@
...
+    context.execute();
+}
+
+fn schedule_infection_events(
+    context: &mut Context,
+    interval: f64,
+    current_day: u32,
+    max_days: u32,
+) {
+    if current_day >= max_days {
```
confunguido, line 61: Should we use here context.shutdown() instead?

ChiragKumar9:
This was subsumed by #82 in a more generic way, so closing.

PR #46, opened 2024-09-11
People + PersonProperties Builder implementation

k88hudson-cfa:
Adds an implementation of People / PersonProperties and a simple example.

One thing I changed from Jason's earlier sketch is that person properties shouldn't actually get added to the data container until the `.insert()` method is called (instead they internally call a builder callback). 

I don't love the stuff around returning defaults if a person property doesn't exist (I think it would be better to panic) but I figured that could be addressed in a follow-up.

### DIFF
--- a/src/people.rs
+++ b/src/people.rs
@@ -1,0 +1,278 @@
+use std::{
+    any::{Any, TypeId},
+    collections::HashMap,
+};
+
+#[derive(Clone, Copy, Hash, PartialEq, Eq)]
+pub struct PersonId {
+    id: usize,
+}
+
+struct PeopleData {
+    population: usize,
+}
+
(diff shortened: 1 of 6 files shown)


src/people.rs
```diff
@@ -0,0 +1,339 @@
...
+    }
+}
+
+pub trait ContextPeopleExt {
+    /// Creates a person by instantiating a `PersonBuilder`, which can be used to
+    /// set properties before inserting it into the data container.
+    fn create_person(&mut self) -> PersonBuilder;
+
+    /// Given a Persionid, returns the value of a defined person property
+    fn get_person_property<T: PersonProperty + 'static>(
```
jasonasher, line 202: Should we make this panic if the `PersonId` hasn't been created yet? (And same for the set method below)

jasonasher (approved):
Approved modulo some small recommendations for changes

Issue #24, opened 2024-08-20
Add parameter loading + global properties component


k88hudson-cfa:
As per the basic example, we want some way to load parameters from a config file and expose them as global properties.

Something along the lines of:
```rs
struct Parameters {
    random_seed: u64,
    population_size: usize,
    foi: f64,
    infection_period: u64
}
context.load_parameters<Parameters>("config.toml")
```

Which would then be available to any module.

Some global properties will need to be mutable.

k88hudson-cfa:
For an initial implementation at this, I'd recommend the following:

* Have people define a single `Parameters` struct for now
* Use this to load / validate everything under a`parameters` property of given yaml config path. This should be some kind of trait extension to context like `load_parameters`
* Give people immutable access to each property (don't allow changes / mutation events for now) 

We will need to adjust the implementation to allow for mutation/events etc., but I think it shares a lot of the same requirements as person properties and we can update it later when we come up with what we want to do then

PR #47, opened 2024-09-11
Definition of global properties closes issue #24

confunguido:
This PR will provide the implementation of global properties along with an example of how to load parameters and define them as global properties.


examples/parameter-loading/infection_manager.rs
```diff
@@ -0,0 +1,99 @@
...
+    define_global_property!(Parameters, ParametersValues);
+    
+    fn handle_recovery_event(context: &mut Context, event: InfectionStatusEvent) {
+        if matches!(event.updated_status, InfectionStatus::R) {
+            *context.get_data_container_mut(RecoveryPlugin) += 1;
+        }
+    }
+
+    #[test]
+    fn test_handle_infection_change() {
```
confunguido, line 69: I can't get this test to pass and not sure what I'm doing wrong. I am trying to set Parameters as a Global Property with the values specified in `parameter_values`. It works in the scope of the test, but in the function to `schedule_recovery` the global property shows as not defined. Does anyone have any idea of what I'm doing wrong? @k88hudson-cfa @jasonasher @ekr-cfa @ChiragKumar9 @bbbruce

jasonasher: You are defining the `Parameters` struct in the `test` module (i.e. inside of `mod test`), so it's in a separate namespace and thus not visible to the function `schedule_recovery` defined outside.

confunguido: I figured it was something like that. That's an outdated version though. Still, I would like to know why it works now by calling `parameters_loader::ParametersValues`. 

More importantly, I would like to understand what's the best way to define those structs in the tests.

ekr-cfa:
This is a design, not a code issue, but I think it's worth asking if we would prefer JSON to toml..

k88hudson-cfa (changes requested):

examples/parameter-loading/infection_manager.rs
```diff
@@ -0,0 +1,100 @@
...
+
+            context.add_plan(1.0, move |context| {
+                println!("time {:?} - PersonID: {:?} - status: {:?}",
+                    context.get_current_time(), person,
+                    context.get_person_property(person, InfectionStatusType));
+                context.set_person_property(
+                    person,
+                    InfectionStatusType,
+                    InfectionStatus::I);
+            });
```
confunguido, line 87: Can someone help here? I am having a hard time with this test. The purpose of the test is to make sure that people who are set to be infectious will eventually recover (`handle_infection_status_change`). However, it isn't working properly. I added this print statement here and below to check people's infection status. Weird thing is, people actually recover if I include the statement from lines 79-82, but they never do if I remove that statement. What am I doing wrong? @jasonasher @ekr-cfa @k88hudson-cfa @ChiragKumar9

ekr-cfa:
I can take a look this afternoon

examples/parameter-loading/infection_manager.rs
ekr-cfa: I bet this is a consequence of lazy person properties. More later

src/report.rs
```diff
@@ -335,10 +336,8 @@ mod test {
 
             for (j, record) in records.enumerate() {
                 let record: SampleReport = record.expect("Failed to deserialize record");
-                assert_eq!(
-                    record.id,
-                    (i * num_reports_per_thread + j).try_into().unwrap()
-                );
+                let id_expected = TryInto::<u32>::try_into(i * num_reports_per_thread + j).unwrap();
```
k88hudson-cfa, line 339: nice

Cargo.toml
```diff
@@ -14,6 +14,8 @@ rand = "0.8.5"
 csv = "1.3"
 serde = { version = "1.0", features = ["derive"] }
 serde_derive = "1.0"
+toml = "0.8.19"
```
k88hudson-cfa, line 17: This can be removed right?

src/global_properties.rs
```diff
@@ -0,0 +1,194 @@
...
+        let data_container = self.get_data_container(GlobalPropertiesPlugin).unwrap();
+        data_container.get_global_property_value::<T>()
+    }
+
+    fn load_parameters_from_json<T: 'static + Debug + DeserializeOwned>(
+        &mut self,
+        file_name: &Path,
+    ) -> Result<T, IxaError> {
+        let config_file = fs::read_to_string(file_name)?;
+        let config = serde_json::from_str(&config_file)?;
```
k88hudson-cfa, line 134: I'd probably open the file and use `from_reader` here https://docs.rs/serde_json/latest/serde_json/fn.from_reader.html rather than loading the whole file into a string first

src/people.rs
```diff
@@ -25,9 +26,9 @@ define_data_plugin!(
 
 // Represents a unique person - the id refers to that person's index in the range
 // 0 to population - 1 in the PeopleData container.
-#[derive(Clone, Copy, Hash, PartialEq, Eq)]
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Serialize, Deserialize)]
 pub struct PersonId {
-    id: usize,
+    pub id: usize,
```
k88hudson-cfa, line 31: Do you still need to make this public if you have get_person_id?

k88hudson-cfa (changes requested):
I think this is good to go with the change to the file reading

src/people.rs
confunguido: Well, I am not sure. I tried making it private, but changes in person properties emit events that contain a `person_id` (of type PersonId). This is causing issues in the incidence report as well. To send the report, I am doing the following (which doesn't look great):
```rust
`context.send_report(IncidenceReportItem` {
  time: context.get_current_time(),
  person_id: event.person_id.id
}
```
Should we implement something like `get_id()` for PersonId, so that we could run `event.person_id.get_id()`? @k88hudson-cfa

k88hudson-cfa: Oh yeah, the reporting use case seems important – I think it's ok just to leave it as `pub` then

k88hudson-cfa: Could you change/use the implementation for `Display` instead?

src/global_properties.rs
```diff
@@ -0,0 +1,202 @@
...
+
+        impl $crate::global_properties::GlobalProperty for $global_property {
+            type Value = $value;
+        }
+    };
+}
+
+/// Provides `IxaError` and maps to other errors to
+/// convert to an `IxaError`
+#[derive(Debug)]
```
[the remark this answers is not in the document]
confunguido: Oh yeah. Done.

src/people.rs
confunguido: Merged your implementation

k88hudson-cfa (approved):
🌟🌟🌟

PR #53, opened 2024-09-30
Derived properties

k88hudson-cfa:
This PR adds derived properties, for example:

```rust
define_derived_person_property!(
    MastersRunner,
    bool,
    [Age, IsRunner],
    |age, is_runner| age >= 40 && is_runner
);
```

The macro takes a list of dependent properties and a closure which computes the property from those dependencies.

Derived properties can be used as other properties (with `get_person_property`) etc.; they are recomputed whenever a dependency changes (and will also emit a change event).

As a side effect of needing to do some lazy registration of dependencies if the property is derived, I needed to introduce a wrapper around subscribing to person property changes, which I think is not ideal: 

```rust
  context.subscribe_to_person_property_changed(
        DiseaseStatusType,
        |_context, person, current, previous| {
            println!("{person:?} changed disease status from {previous:?} to {current:?}");
        },
    );
```
In particular, I think it's probably pretty easy to mess up the order of current and previous. 

Open to suggestions on how to handle this better

### DIFF
--- a/src/people.rs
+++ b/src/people.rs
@@ -2,7 +2,7 @@ use crate::{context::Context, define_data_plugin};
 use std::{
     any::{Any, TypeId},
     cell::{RefCell, RefMut},
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fmt,
 };
 
@@ -14,7 +14,8 @@ type DerivedSetter = dyn Fn(&mut Context, PersonId);
 struct PeopleData {
     current_population: usize,
     properties_map: RefCell<HashMap<TypeId, Box<dyn Any>>>,
-    property_dependencies: HashMap<TypeId, Vec<Box<DerivedSetter>>>,
+    registered_derived_properties: RefCell<HashSet<TypeId>>,
+    property_dependencies: RefCell<HashMap<TypeId, Vec<Box<DerivedSetter>>>>,
 }
 
 define_data_plugin!(
@@ -23,7 +24,8 @@ define_data_plugin!(
     PeopleData {
         current_population: 0,
         properties_map: RefCell::new(HashMap::new()),
-        property_dependencies: HashMap::new()
+        registered_derived_properties: RefCell::new(HashSet::new()),
+        property_dependencies: RefCell::new(HashMap::new())
     }
 );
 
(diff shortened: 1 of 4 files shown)


src/people.rs
```diff
@@ -60,6 +67,15 @@ macro_rules! define_person_property {
         pub struct $person_property;
         impl $crate::people::PersonProperty for $person_property {
             type Value = $value;
+            fn is_derived() -> bool {
+                false
```
ekr-cfa, line 71: You could make this a default.

src/people.rs
```diff
@@ -60,6 +67,15 @@ macro_rules! define_person_property {
...
+            }
+            fn calculate(_context: &Context, _person_id: $crate::people::PersonId) -> Self::Value {
+                panic!("Property not derived");
+            }
+            fn dependencies() -> Vec<std::any::TypeId> {
+                panic!("Property not derived");
+            }
```
ekr-cfa, line 78: These can also be defaults.

src/people.rs
```diff
@@ -75,6 +91,40 @@ macro_rules! define_person_property {
...
+#[macro_export]
+macro_rules! define_derived_person_property {
+    ($derived_property:ident, $value:ty, [$($dependency:ident),+], |$($param:ident),+| $derive_fn:expr) => {
+        #[derive(Copy, Clone)]
+        pub struct $derived_property;
+
+        impl $crate::people::PersonProperty for $derived_property {
+            type Value = $value;
+
+            fn calculate(context: &$crate::context::Context, person_id: $crate::people::PersonId) -> Self::Value {
```
ekr-cfa, line 108: Given that initialize and calculate are the same, why not just have initialize?

src/people.rs
```diff
@@ -427,15 +547,110 @@ mod test {
...
+    #[test]
+    fn derived_property_change_event() {
+        let mut context = Context::new();
+        define_derived_person_property!(MastersRunner, bool, [Age, IsRunner], |age, is_runner| age
+            >= 40
+            && is_runner);
+
+        let person = context.add_person();
+        context.set_person_property(person, Age, 50);
+        context.set_person_property(person, IsRunner, false);
```
ekr-cfa, line 608: You need to stop people from externally calling set_person_property() on derived. I think I would do this by having a set_person_property_int() that is private but called by set_person_property()

ekr-cfa (changes requested):

src/people.rs
k88hudson-cfa: Overloading the meaning of a single function seems not ideal to me, this is more explicit

PR #57, opened 2024-10-04
Add initialize_person_property. Fixes #56.

### LINKED ISSUE #56: Unexpected behavior with events not firing on setting of property with default
While implementing https://github.com/CDCgov/ixa/pull/47, we noticed some less than ideal behavior with setting properties that have default values. The issue is that because initializers are lazy, if the property is never accessed before a new value is set, an event will not be fired (since no value existed previously). 

For example, say you have a `DiseaseStatus` property that should have a value of `DiseaseStatus::S` by default.

When you call `set_person_property` in order to change the status to `DiseaseStatus::I`, you expect an event to be released; however, because initialization is lazy and the property was never initialized, the event is *not* fired.

```rust
define_person_property_with_default!(DiseaseStatusType, DiseaseStatus, DiseaseStatus::S);

for _ in 0..population_size {
    let person = context.add_person();

    // Infect the person
    context.set_person_property(person, DiseaseStatusType, DiseaseStatus::I);
    
    context.subscribe_to_event(person, |context, event:PersonPropertyChangeEvent<DiseaseStatusType>| {
        // We expect this to fire when set person_property is called
    });
}
```

If instead you call `get_person_property` before setting it, the event *will* be fired.

```rust
define_person_property_with_default!(DiseaseStatusType, DiseaseStatus, DiseaseStatus::S);

for _ in 0..population_size {
    let person = context.add_person();
    // Force initialization
    context.get_person_property(person, DiseaseStatusType);
    // Now the event is fired
    context.set_person_property(person, DiseaseStatusType, DiseaseStatus::I);
    context.subscribe_to_event(person, |context, event:PersonPropertyChangeEvent<DiseaseStatusType>| {
        // We expect this to fire when set person_property is called
    });
}
```

ekr-cfa:


src/people.rs
```diff
@@ -213,6 +224,22 @@ impl ContextPeopleExt for Context {
         initialized_value
     }
 
+    fn initialize_person_property<T: PersonProperty + 'static>(
```
k88hudson-cfa, line 227:
```suggestion
    /// Sets an initial value for a property, overriding the initialize() function if one is defined.
    /// Does *not* fire an event. Panics if a value is already set.
    fn initialize_person_property<T: PersonProperty + 'static>(
```

src/people.rs
```diff
@@ -213,6 +224,22 @@ impl ContextPeopleExt for Context {
...
+        &mut self,
```
k88hudson-cfa, line 228: Is there a reason this needs to be a mutable reference?

k88hudson-cfa (approved):
LGTM, this has much better properties. Thanks 👍

src/people.rs
ekr-cfa: Well I'm not sure it overrides it. It just bypasses it.

src/people.rs
ekr-cfa: Hmm.... I was thinking the same reason set_person_property was.

ekr-cfa: OK, so you don't need it to be `mut` but i think we should have it that way anyway, because it signals to callers that it actually changes things by intention

k88hudson-cfa: I think it would be better for it not to be, since it would prevent you from calling it where you only have an immutable reference to context (maybe you want to do it lazily somewhere). The intention seems pretty clear to me from the name of the method

ekr-cfa: You shouldn't be doing that. If you want to lazily initialize you should use an initializer.

The impact of `RefCell` is to hide the true semantics. That's OK in the lazy initialization because it has "as if" semantics but this doesnt'.

ekr-cfa:
@k88hudson-cfa I think this is ready to land

k88hudson-cfa (approved):

Issue #61, opened 2024-10-18
Is `context.get_person_id(usize_id)` necessary?


ChiragKumar9:
The challenges described in #60 arises because `context.get_person_id(usize_id)` returns `PersonId{ id: usize_id}` even if that person doesn't technically exist. Given a valid `PersonId`, we can query properties for that person in from `PeopleData`. Right now, we need this functionality for the following kinds of patterns:

1) Count the number of people who have a certain property value

```rust
// count the number of people who are infected
// could also write this with an iter and collect
// but to be explicit i will not use that syntax here
let mut counter = 0;
for usize_id in 0..context.get_current_population() {
    if matches!(context.get_person_property(context.get_person_id(usize_id), InfectionStatusType),
                       InfectionStatus::I) {
        counter += 1;
}

```

2) Identify a person who can be infected by randomly drawing Ids.

```rust
fn attempt_infection(context: &mut Context) {
    let population_size: usize = context.get_population();
    let person_to_infect: usize = context.sample_range(TransmissionRng, 0..population_size);

    let person_status: InfectionStatus = context.get_person_property(context.get_person_id(person_to_infect),
                                                                                                            InfectionStatusType);

    if matches!(person_status, InfectionStatus::S) {
        context.set_person_property(context.get_person_id(person_to_infect), InfectionStatusType,
            InfectionStatus::I);
    }
}
```

I believe both functionalities can be addressed by alternative syntax, meaning we can get rid of `context.get_person_id(...)`.

In essence, 1) is asking us to get the number of people who have a given person property value. This is just the size of a partition defined by the given person property values. If we want the entire set of people in the simulation, we could define some partition that is the entire set of alive people, and take it from there. I believe this may also negate the need for `context.get_current_population()`. Here's what some code could look like:

```rust
define_partition!(Infecteds, InfectionStatusType == InfectionStatus::I);
let counter = context.get_partition(Infecteds).len();
```

2) is asking us for something quite similar: given all people in the simulation, return a random one. First of all, #55 indicates that such a helper is in the works. But, we could use the same partition idea:

```rust
define_partition!(AllPeopleAlive, Alive);
let person_to_infect = context.sample_population(InfectionRng, context.get_partition(AllPeopleAlive));
```

This may be overkill! And, this all assumes we don't have a way to remove people (i.e., `context.remove_person`) but rather just keep removed people in the simulation but with some property flag that marks them as not in the simulation (i.e., `dead`) and from there we can create partitions that filter such people out.

[...]

Issue #60, opened 2024-10-18
Default person properties returned for people who don't exist


ChiragKumar9:
In playing around with `people` in #54, I noticed some odd behavior of people.

If at least one person has been added to the simulation, `context.get_person_property(...)` returns the default value of a person property even if the given person does not exist. Instead, it should throw an `IxaError: Person does not exist`. As I understand it, Line 128 of `people` https://github.com/CDCgov/ixa/blob/79c16105354a40ec8b67cb062849c2fa5e8427dd/src/people.rs#L128 is supposed to check for this and make the output `None`.

The attached branch provides examples of this unexpected behavior in practice and a PR is in the works for a potential solution.

```rust
define_person_property_with_default(IsRunner, bool, false);

fn bug_person_properties_for_nonexistent_ppl() {
        define_person_property!(LivesInDC, bool);
        let mut context = Context::new();
        context.add_person(); // add only one person to this sim
        // default value of IsRunner is false -- test for 2nd person (recall zero indexing)
        assert!(!context.get_person_property(context.get_person_id(1), IsRunner));
        // I can do other kinds of weird reasoning about the person
        context.set_person_property(context.get_person_id(1), IsRunner, true);
        // initialize person property for a person who doesn't yet exist?
        context.initialize_person_property(context.get_person_id(1), LivesInDC, false);
        assert!(!context.get_person_property(context.get_person_id(1), LivesInDC));
        // but current population is still just one person, not two
        assert_eq!(context.get_current_population(), 1);
    }
```

ChiragKumar9:
Update: it turns out you can do a lot of erroneous reasoning about people who don't yet exist in setting their person properties.

PR #62, opened 2024-10-18
Only return person ids for people that exist in the simulation so far

ChiragKumar9:
Bandage solution to #60 and #61


ekr-cfa (approved):
LGTM

PR #64, opened 2024-10-22
Derived properties

k88hudson-cfa:
This implementation supports multiple property dependencies and computes values lazily for derived properties (avoiding some of the mechanics around updating the cache and therefore the need to register early). Registration happens during subscription only; we will also need to do this during index creation). 

I avoid wrapping the events with a special function by implementing a `on_subscribe` hook on an `Event` trait.

### DIFF
--- a/src/people.rs
+++ b/src/people.rs
@@ -112,7 +112,7 @@ where
     ) {
         let previous = context.get_person_property(person, T::get_instance());
 
-        // Stores the current value of the person property and defer the actual event
+        // Captures the current value of the person property and defers the actual event
         // emission to when we have access to the new value.
         callback_vec.push(Box::new(move |ctx| {
             let current = ctx.get_person_property(person, T::get_instance());
--- a/src/people.rs
+++ b/src/people.rs
@@ -7,7 +7,7 @@ use std::{
     any::{Any, TypeId},
     cell::{RefCell, RefMut},
     collections::{HashMap, HashSet},
-    fmt,
+    fmt::{self},
 };
 
 // PeopleData represents each unique person in the simulation with an id ranging
@@ -63,7 +63,7 @@ pub trait PersonProperty: Copy {
         false
     }
     #[must_use]
-    fn dependencies() -> Vec<TypeId> {
+    fn dependencies() -> Vec<Box<dyn PersonPropertyHolder>> {
         panic!("Dependencies not implemented");
     }
     fn compute(context: &Context, person_id: PersonId) -> Self::Value;
@@ -83,6 +83,11 @@ pub trait PersonPropertyHolder {
         person: PersonId,
         callback_vec: &mut Vec<Box<ContextCallback>>,
     );
+    fn is_derived(&self) -> bool;
+    fn dependencies(&self) -> Vec<Box<dyn PersonPropertyHolder>>;
+    fn non_derived_dependencies(&self) -> Vec<TypeId>;
+    fn collect_non_derived_dependencies(&self, result: &mut HashSet<TypeId>);
+    fn property_type_id(&self) -> TypeId;
 }
 
 impl<T> PersonPropertyHolder for T
@@ -106,6 +111,40 @@ where
             ctx.emit_event(change_event);
         }));
     }
+
+    fn is_derived(&self) -> bool {
+        T::is_derived()
+    }
+
+    fn dependencies(&self) -> Vec<Box<dyn PersonPropertyHolder>> {
+        T::dependencies()
+    }
+
+    fn property_type_id(&self) -> TypeId {
+        TypeId::of::<T>()
+    }
+
+    /// Returns of dependencies, where any derived dependencies
+    /// are recursively expanded to their non-derived dependencies.
+    /// If the property is not derived, the Vec will be empty.
+    fn non_derived_dependencies(&self) -> Vec<TypeId> {
+        let mut result = HashSet::new();
+        self.collect_non_derived_dependencies(&mut result);
+        result.into_iter().collect()
+    }
+
+    fn collect_non_derived_dependencies(&self, result: &mut HashSet<TypeId>) {
+        if !self.is_derived() {
+            return;
+        }
+        for dependency in self.dependencies() {
+            if dependency.is_derived() {
+                dependency.collect_non_derived_dependencies(result);
+            } else {
+                result.insert(dependency.property_type_id());
+            }
+        }
+    }
 }
 
 /// Defines a person property with the following parameters:
@@ -117,7 +156,7 @@ where
 #[macro_export]
 macro_rules! define_person_property {
     ($person_property:ident, $value:ty, $initialize:expr) => {
-        #[derive(Copy, Clone)]
+        #[derive(Debug, Copy, Clone)]
         pub struct $person_property;
         impl $crate::people::PersonProperty for $person_property {
             type Value = $value;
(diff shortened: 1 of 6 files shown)


k88hudson-cfa:
There is an existing issue with dependencies that are derived properties (you can see a failing example in the tests, but here's a bit of an explanation of what I think is going on).

Say you have a property `UltraRunner` that depends on another derived property,
`TrailRunner`, which itself depends on `IsRunner`.

```rust
define_derived_property!(UltraRunner, bool, [TrailRunner], |age, trail_runner| {
    age > 40 && trail_runner
});
define_derived_property!(TrailRunner, bool, [Age, IsRunner], |age, is_runner| {
    age > 29 && is_runner
});
```

In this scenario, when you register for a PersonPropertyChangedEvent of `UltraRunner`,
the following dependency chain gets added (note, I've simplified the types here for readability):

```rust
{
    Age: [UltraRunner],
    TrailRunner: [UltraRunner]
}
```

Because the relationship between `IsRunner` is only inside `TrailRunner`'s dependency list, that isn't 
properly reflected in the dependency registration here and no event for `UltraRunner` will be fired.

Here is the implementation:

```rust
 fn register_property<T: PersonProperty + 'static>(&mut self) {
    // ...
    let dependencies = T::dependencies();
    for dependency in dependencies {
        let mut property_dependencies = data_container.dependency_map.borrow_mut();
        let deps = property_dependencies.entry(dependency).or_default();
        deps.push(Box::new(T::get_instance()));
    }
    data_container
        .registered_derived_properties
        .borrow_mut()
        .insert(TypeId::of::<T>());
}
```
What this code *should* do is check each dependency type
to see if it's derived or not, and follow it recursively (so in the case of
`TrailRunner`, it should not add an entry for `TrailRunner` directly, but rather
call itself and add `UltraRunner` to the list of the dependencies for `TrailRunner`),
but unfortunately `T::dependencies()` currently only returns a `TypeId`, not 
an actual type like I would need to do that.

k88hudson-cfa:
Alright, I managed to resolve the issue I described above by adding a method to recursively get all the non derived properties and set up the dependency chain using that instead.

src/people.rs
```diff
@@ -257,26 +421,56 @@ impl ContextPeopleExt for Context {
...
+            let mut dependencies = data_container.dependency_map.borrow_mut();
+            dependencies.get_mut(&TypeId::of::<T>()).map(std::mem::take)
+        };
+
+        let mut dependency_event_callbacks = Vec::new();
+        if let Some(mut deps) = deps_temp {
+            // If there are dependencies, set up a bunch of callbacks with the
+            // current value
+            for dep in &mut deps {
+                dep.add_event_callback(self, person_id, &mut dependency_event_callbacks);
```
ekr-cfa, line 454: I would call this `dependency_changed()` so that (for instance) you can handle indexing in the same cb if you want to.

src/people.rs
```diff
@@ -257,26 +421,56 @@ impl ContextPeopleExt for Context {
...
+        data_container.set_person_property(person_id, property, value);
         let change_event: PersonPropertyChangeEvent<T> = PersonPropertyChangeEvent {
             person_id,
             current: value,
             previous: previous_value,
         };
-        data_container.set_person_property(person_id, property, value);
         self.emit_event(change_event);
+
+        // If there are dependency callbacks, call them with the updated value
```
ekr-cfa, line 473: You're not calling them with the updated value, though, you're just calling them and they get the updated value on their own.

src/people.rs
```diff
@@ -51,7 +58,93 @@ impl fmt::Debug for PersonId {
...
+        callback_vec: &mut Vec<Box<ContextCallback>>,
+    );
+    fn is_derived(&self) -> bool;
+    fn dependencies(&self) -> Vec<Box<dyn PersonPropertyHolder>>;
+    fn non_derived_dependencies(&self) -> Vec<TypeId>;
+    fn collect_non_derived_dependencies(&self, result: &mut HashSet<TypeId>);
+    fn property_type_id(&self) -> TypeId;
+}
+
+impl<T> PersonPropertyHolder for T
```
ekr-cfa, line 93: As an aside I spent a while trying to figure out how to make this a supertrait to avoid the forwarders, but regrettably Copy appears to require Sized and so you can't make it object safe.

ekr-cfa (commented):
As we discussed, some small comments. If we decided to land this, we'll want to add a comment explaining the design.

src/people.rs
```diff
@@ -310,6 +523,15 @@ mod test {
             0
         }
     });
+    define_derived_property!(TrailRunner, bool, [IsRunner, Age], |is_runner, age| {
+        is_runner && age > 29
+    });
+    define_derived_property!(
+        UltraRunner,
+        bool,
+        [TrailRunner, Age],
```
ekr-cfa, line 532: I know I'm nitpicking, but I think it would be better to have properties which were self-evident from the name,  and it's not obvious that a trailrunner is a runner > 29.

src/people.rs
```diff
@@ -291,10 +488,26 @@ impl ContextPeopleExt for Context {
 #[cfg(test)]
 mod test {
     use super::{ContextPeopleExt, PersonCreatedEvent, PersonId, PersonPropertyChangeEvent};
-    use crate::{context::Context, people::PeoplePlugin};
-    use std::{cell::RefCell, rc::Rc};
+    use crate::{
+        context::Context,
+        people::{PeoplePlugin, PersonPropertyHolder},
```
ekr-cfa, line 493: Why do you need PersonPropertyHolder.

src/people.rs
```diff
@@ -553,4 +775,129 @@ mod test {
...
+        );
+        context.subscribe_to_event(
+            move |_context, _event: PersonPropertyChangeEvent<TrailRunner>| {
+                // Make sure that we don't register multiple times
+            },
+        );
+        context.set_person_property(person, Age, 30);
+        context.execute();
+        assert_eq!(*flag.borrow(), 1);
+    }
```
ekr-cfa, line 870: Can you add a test with a diamond dependency that verifies that something only fires once: i.e.

A depends on B and C and B and C both depend on D?

src/people.rs
```diff
@@ -553,4 +775,129 @@ mod test {
...
+
+        let flag = Rc::new(RefCell::new(false));
+        let flag_clone = flag.clone();
+        assert!(!context.get_person_property(person, UltraRunner));
+        context.subscribe_to_event(
+            move |_context, event: PersonPropertyChangeEvent<UltraRunner>| {
+                assert_eq!(event.person_id.id, 0);
+                assert!(!event.previous);
+                assert!(event.current);
+                *flag_clone.borrow_mut() = true;
```
ekr-cfa, line 896: Can you make this a counter, so you verify it fires once.

ekr-cfa (approved):
LGTM with the indicated changes.

src/people.rs
k88hudson-cfa: Because I am calling `non_derived_dependencies()` in the test

src/people.rs
```diff
@@ -51,7 +58,93 @@ impl fmt::Debug for PersonId {
...
+    }
+    fn compute(context: &Context, person_id: PersonId) -> Self::Value;
+    fn get_instance() -> Self;
+}
+
+type ContextCallback = dyn FnOnce(&mut Context);
+
+// The purpose of this trait is to allow us to store a vector of different PersonProperties
+// in an object safe way, and use them to emit change events.
+pub trait PersonPropertyHolder {
```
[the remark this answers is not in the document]
k88hudson-cfa: I get a warning if I remove it about it being more private than PersonProperty::dependencies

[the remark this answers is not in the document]
ekr-cfa: Comment here?

src/people.rs
```diff
@@ -94,13 +104,16 @@ impl<T> PersonPropertyHolder for T
...
-    fn add_event_callback(
+    fn dependency_changed(
         &self,
         context: &mut Context,
         person: PersonId,
         callback_vec: &mut Vec<Box<ContextCallback>>,
     ) {
         let previous = context.get_person_property(person, T::get_instance());
+
+        // Stores the current value of the person property and defer the actual event
```
ekr-cfa, line 115:
```suggestion
        // Captures the current value of the person property and defers the actual event
```

ekr-cfa (approved):
LGTM modulo one comment about pub and a typo.

Issue #51, opened 2024-09-16
Add births/deaths example


k88hudson-cfa:
Also a good opportunity to exercise person properties / person creation

ChiragKumar9:
Closed by #65

PR #65, opened 2024-10-24
Births and deaths closes issue #51

confunguido:
This PR provides an example of births, deaths, and aging. This example highlights the need for a tool to look up people by person properties (1+) as well as how to sample from those. 

**Being alive** 
To handle births and deaths, I created a person property named `Alive`. This is set to true when a person is added to the simulation and to false when they die. 

**Aging** 
I defined a person property as Birth, instead of age to determine someone's current risk category, which is based on age. For instance, in 1+ years time, someone may switch from one group to the other, and that should be considered. Age and Birth here are defined in years but as f64, so that would account for 1.5 years).
```rust
fn get_person_age(&mut context, person_id: PersonId) -> f64 {
  let birth_time = self.get_person_property(person_id, Birth);
 (context.get_current_time() - birth_time) / 365.0
}
```
**How to get current alive population**
For every operation that requires sampling an individual, estimating current population, etc., we need to filter by people who are alive. I created a function called `get_population_by_property` which finds the total number of people by a specified property. 

```rust
fn get_population_by_property<T: PersonProperty + 'static> (&mut self, property: T, value: T::Value) -> usize {
  let mut population_counter = 0;
  for i in 0..self.get_current_population() {
     let person = self.get_person_id(i);
     if self.get_person_property(person_id, property) == value { 
        population_counter += 1;
     } 
  }
  population_counter
}
```
**Should alive be treated differently when filtering people? **
I wonder whether we should make being alive a special case for some of these filters. For instance, if I want to sample someone to infect, it is obvious I want someone alive, and if the FOI depends on the current population, it is also obvious that it requires the current alive population. Another example of this is periodic reports, where we want to get the number of people per person property, but we want to get those alive. Of course, not all models include aging, so maybe this isn't an issue.


examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,353 @@
...
+        context.attempt_death(person);
+
+        let next_death_event = context.get_current_time()
+            + context.sample_distr(PeopleRng, Exp::new(parameters.death_rate).unwrap());
+
+        context.add_plan(next_death_event, move |context| {
+            schedule_death(context);
+        });
+    }
+    // Cancel all plans
```
k88hudson-cfa, line 69: Did you mean to leave this as TODO?

examples/births-deaths/infection_manager.rs
```diff
@@ -0,0 +1,178 @@
...
+        let population_size: usize = 10;
+        for _ in 0..population_size {
+            let person = context.create_new_person(0.0);
+
+            context.add_plan(1.0, move |context| {
+                context.set_person_property(person, InfectionStatusType, InfectionStatus::I);
+            });
+        }
+
+        context.add_plan(1.1, move |context| {
```
k88hudson-cfa, line 149: why 1.1?

examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn sample_person_by_property<T: PersonProperty + 'static>(
+        &mut self,
+        property: T,
+        value: T::Value,
+    ) -> Option<PersonId>
+    where
+        <T as PersonProperty>::Value: PartialEq,
+    {
+        let mut people_vec = Vec::<PersonId>::new();
+        for i in 0..self.get_current_population() {
```
k88hudson-cfa, line 209: I would probably use `filter_map` here but this is perfectly fine

k88hudson-cfa (changes requested):
A few questions, mostly this looks pretty reasonable to me – the age / age group varying with time will make it tricky for indexing to be useful for a lot of stuff here

For a more general API plan cancelling – we could consider something like `cancel_plans_for_person` (which would require you to also register it, I think), but what you're doing here doesn't seem too bad. What are your thoughts after implementing?

examples/births-deaths/plot_output.R
```diff
@@ -0,0 +1,31 @@
...
+params <- read_json(file.path(dir, "input.json"))
+population <- params$population
+age_groups <- params$foi_groups
+
+
+
+output_df <- read_csv(file.path(dir, "incidence.csv")) |>
+  dplyr::filter(infection_status == "I") |>
+  group_by(time, age_group) |>
+  mutate(inf = n()) |>
```
ChiragKumar9, line 17: this should be a summarize. Imagine you have two infections that occur at the same time -- you want to count those each once, not twice.

examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,353 @@
...
+            schedule_death(context);
+        });
+    }
+    // Cancel all plans
+}
+
+pub fn init(context: &mut Context) {
+    let parameters = context.get_global_property_value(Parameters).clone();
+
+    let foi_map = parameters
```
ChiragKumar9, line 75: Why not do this in the `parameters_loader`?

ChiragKumar9 (changes requested):
The only change I am tied to is the comment about the mathematical inexactness of this simulation compared to what is described in the readme.

examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,353 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{self:?}")
+    }
+}
+
+define_global_property!(Foi, HashMap<AgeGroupRisk, f64>);
+define_person_property_with_default!(InfectionStatusType, InfectionStatus, InfectionStatus::S);
+
+define_person_property!(Birth, f64);
+define_person_property!(Alive, bool);
```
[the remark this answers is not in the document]
confunguido: Yeah, good point. I guess people by default are alive XD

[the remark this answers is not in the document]
confunguido: @k88hudson-cfa I have changed the way age is implemented. Now is back to being u8 and age scheduled to update every year. For canceling plans, I think creating something like `cancel_plans_for_person` would be fine. I think it wasn't too hard to keep track of the plans but made me think of how to keep track of which plans specifically need to be canceled when a person is removed from the simulation.

examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,353 @@
...
+    if parameters.death_rate > 0.0 {
+        context.add_plan(0.0, |context| {
+            schedule_death(context);
+        });
+    }
+}
+
+pub trait ContextPopulationExt {
+    fn create_new_person(&mut self, birth_time: f64) -> PersonId;
+    fn attempt_death(&mut self, person_id: PersonId);
```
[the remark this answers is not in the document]
confunguido: Honestly, this is nothing else that me being unable to write something like `kill_person`. I will change that.

examples/births-deaths/population_manager.rs
confunguido: Not really. I think I realized later that I don't actually know who should take care of canceling all plans. I ended up implementing it in `infection_manager` and just registering for death events. I think that approach isn't too bad, so I just forgot to remove from population manager.

examples/births-deaths/infection_manager.rs
confunguido: I wanted to kill right after infection so that recovery would have to be canceled. It could also be at 1.0, I guess.

examples/births-deaths/population_manager.rs
confunguido: Noted. I'm not sure I should work too much on this sample since it will be replaced once we land the PR to query people, right?

examples/births-deaths/plot_output.R
confunguido: Yeah that makes sense. Though, unlikely with this approach.

examples/births-deaths/population_manager.rs
k88hudson-cfa: Ok thanks. Yeah, just remove the comment then

examples/births-deaths/population_manager.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+});
+
+fn schedule_aging(context: &mut Context, person_id: PersonId) {
+    if context.get_person_property(person_id, Alive) {
+        let prev_age = context.get_person_property(person_id, Age);
+        context.set_person_property(person_id, Age, prev_age + 1);
+        let next_age_event = context.get_current_time() + 365.0;
+        context.add_plan(next_age_event, move |context| {
+            schedule_aging(context, person_id);
```
k88hudson-cfa, line 58: Isn't it kind of weird that everyone in the simulation gets a birthday of t = 0 (well, -365*Age) except people who are born after? I defer to you on what is more natural on the simulation side of things but I'd probably do this a bit differently....

Rather than scheduling aging events for each person, I'd probably index by birthday and then check who's age needs to be incremented. Alternatively, you could assume everyone's birthday is t=0 and just increment everyone's age every 365 days. What do you think?

k88hudson-cfa (changes requested):
This mostly looks good and I like the age incrementing approach – just have a question / alternative idea for scheduling by birthday (rather than by individual) I think we should discuss.

Looks like @ChiragKumar9 has some comments too

examples/births-deaths/population_manager.rs
confunguido: I modified this by adding a random birthday from 0 to 365 when creating a new person at the beginning of the simulation. Let me know if this looks good for you. I addressed @ChiragKumar9 comments as well but let me know if I missed anything.

k88hudson-cfa (approved):

k88hudson-cfa:
LGTM! @ChiragKumar9 needs to approve to merge

ChiragKumar9 (approved):

PR #66, opened 2024-10-27
Person Properties Periodic Report

ChiragKumar9:
I started hacking around and got pretty close to building in easy user functionality for a periodic person properties report like GCM had. In this PR, I'm trying to put on more of my developer hat, so any feedback on how to improve the Rust code and structure is welcome. One of the features that I always wanted in GCM was a periodic report that listed people by cross combination of two person properties (number of people who are sick and under five, sick and over 65, etc.), so I implemented this as a necessary first step to adding that functionality.

However, this process raised a bunch of key questions for me about our `ixa` design so far, mainly on lazy property initialization.

### Design Sketch

1. When defining a person property, the user specifies whether they would like the property included in the report or not.
```rust
define_person_property!(IsRunner, bool, true); // true means to include in report
define_person_property_with_default!(RunningShoes, u8, false, 0); // third arg is whether in report
```
2. All person properties implement a trait `PersonPropertiesPeriodicReport` that provides a method that tabulates the number of occurrences of each of the observed property values.
3. The reports module has a method that allows people to add a person properties periodic report that reports out the properties tabulation every `report_period` amount of time. Properties that are included in the report are passed onto the reports module automatically, and the reports module iterates over them, calling their tabulation method and sending a report for each property value to the `person_properties_report.csv`.

### Implementation challenges
Challenges I describe here are also listed in the comments in the code, if it's easier to read inline.
1. At the current time, we have no `context.remove_person(person_id)` functionality, and it does not seem like we will add this functionality. This means that people we have set to be dead are still reported out on in the report. We could create add a new person property value to every property specific to dead people, or make every property an Option of what it already is and set dead people to be None, but neither of these feel like (a) good solutions (the first case sounds dangerous for f64 types and in the second case what if the type is already an Option?) or (b) user-friendly solutions. **Some sort of `context.person_teardown(person_id)` functionality would be really helpful.**
2. Each row in the report is associated with a time, but the interpretation of that time is not consistent. The report plan is added to the plan queue like any other plan, and there could be plans added after it at the same time as it. It is logical to make the row mean "the state of the world at _the end_ of the listed time", but to implement that **we need to have plans with priority.**
3. Lazy initialization for person properties means that the initialize method for a property isn't called until that property is first accessed. The initialize method also adds the property to the list of properties, so a person property isn't included in the report until it is first accessed. One way around this is by producing an "initialization" report as well, and then our accompanying Python analysis package for `ixa` can stitch together the two to make a true person properties periodic report.
    - Moreover, lazy initialization also includes automatic set expansion. So, if the property is only accessed for the first two of five people (i.e., people with Ids 0 and 1), only property values for the first two people will be listed, not even listing "None" for the remaining three. if the property is accessed for people 0 and 3, two values will be listed and two "None"s will be listed. I have a test that shows this counterintuitive behavior in `people.rs`.

I think I caught a bug on the data plugins macro. I believe the data plugins struct must be public so that data plugins are accessible beyond just the mod in which they are created. This is necessary for being able to share the person properties data plugin with the reports module and generally for user-defined data plugins. I can pull out those commits into a separate PR if this is a bug fix. However, to not give the user the entirety of the data container plugin in a way that they could mess with it, I've made the `properties_map` public only to the crate. I couldn't do the same for the `include_in_periodic_report` attribute, but I do not understand why.

### Next steps
1. GCM printed out person properties tabulations by region. With the argument that we just want to support a broad "partition" plugin for users, of which regions is one subclass, it is not clear to me where listing out number of people with each person property by some group fits in here -- perhaps the user can pick one of their one-to-one groups as a stratifier?
2. The code for counting number of people with each person property may be refactored as new `ixa` functionality comes online.
3. I realized that the implementation for the cross combination properties report I described is probably very different, instead with the user just providing the two particular properties they want cross tabulated as function inputs and does not require the `PersonPropertiesPeriodicReport` trait.

### Example of what the report looks like
image

### DIFF
--- a/src/global_properties.rs
+++ b/src/global_properties.rs
@@ -31,7 +31,8 @@ pub trait GlobalProperty: Any {
 
 pub use define_global_property;
 
-struct GlobalPropertiesDataContainer {
+#[allow(clippy::module_name_repetitions)]
+pub struct GlobalPropertiesDataContainer {
     global_property_container: HashMap<TypeId, Box<dyn Any>>,
 }
 
--- a/src/people.rs
+++ b/src/people.rs
@@ -13,7 +13,8 @@ use std::{
 #[allow(clippy::module_name_repetitions)]
 pub struct PeopleData {
     current_population: usize,
-    properties_map: RefCell<HashMap<TypeId, Box<dyn Any>>>,
+    pub(crate) properties_map: RefCell<HashMap<TypeId, Box<dyn Any>>>,
+    pub include_in_periodic_report: HashMap<TypeId, String>,
 }
 
 define_data_plugin!(
@@ -21,7 +22,8 @@ define_data_plugin!(
     PeopleData,
     PeopleData {
         current_population: 0,
-        properties_map: RefCell::new(HashMap::new())
+        properties_map: RefCell::new(HashMap::new()),
+        include_in_periodic_report: HashMap::new()
     }
 );
 
@@ -53,17 +55,19 @@ impl fmt::Debug for PersonId {
 pub trait PersonProperty: Copy {
     type Value: Copy;
     fn initialize(context: &Context, person_id: PersonId) -> Self::Value;
+    fn include_in_periodic_report(&self) -> bool;
 }
 
 /// Defines a person property with the following parameters:
 /// * `$person_property`: A name for the identifier type of the property
 /// * `$value`: The type of the property's value
+/// * `$include`: A boolean indicating whether the property should be included in the periodic report
 /// * `$initialize`: (Optional) A function that takes a `Context` and `PersonId` and
 ///   returns the initial value. If it is not defined, calling `get_person_property`
 ///   on the property without explicitly setting a value first will panic.
 #[macro_export]
 macro_rules! define_person_property {
-    ($person_property:ident, $value:ty, $initialize:expr) => {
+    ($person_property:ident, $value:ty, $include:expr, $initialize:expr) => {
         #[derive(Copy, Clone)]
         pub struct $person_property;
         impl $crate::people::PersonProperty for $person_property {
@@ -74,25 +78,37 @@ macro_rules! define_person_property {
             ) -> Self::Value {
                 $initialize(_context, _person)
             }
+            fn include_in_periodic_report(&self) -> bool {
+                $include
+            }
         }
     };
-    ($person_property:ident, $value:ty) => {
-        define_person_property!($person_property, $value, |_context, _person_id| {
-            panic!("Property not initialized");
-        });
+    ($person_property:ident, $value:ty, $include:expr) => {
+        define_person_property!(
+            $person_property,
+            $value,
+            $include,
+            |_context, _person_id| {
+                panic!("Property not initialized");
+            }
+        );
     };
 }
 
 /// Defines a person property with the following parameters:
 /// * `$person_property`: A name for the identifier type of the property
 /// * `$value`: The type of the property's value
+/// * `$include`: A boolean indicating whether the property should be included in the periodic report
 /// * `$default`: An initial value
 #[macro_export]
 macro_rules! define_person_property_with_default {
-    ($person_property:ident, $value:ty, $default:expr) => {
-        define_person_property!($person_property, $value, |_context, _person_id| {
-            $default
-        });
+    ($person_property:ident, $value:ty, $include:expr, $default:expr) => {
+        define_person_property!(
+            $person_property,
+            $value,
+            $include,
+            |_context, _person_id| { $default }
+        );
     };
 }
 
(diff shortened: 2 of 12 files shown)


src/global_properties.rs
```diff
@@ -31,7 +31,8 @@ pub trait GlobalProperty: Any {
 
 pub use define_global_property;
 
-struct GlobalPropertiesDataContainer {
+#[allow(clippy::module_name_repetitions)]
```
k88hudson-cfa, line 34: Thanks for the lint fixes

src/people.rs
```diff
@@ -174,7 +267,7 @@ pub trait ContextPeopleExt {
     /// initializing it if it hasn't been set yet. If no initializer is
     /// provided, and the property is not set this will panic
     fn get_person_property<T: PersonProperty + 'static>(
-        &self,
+        &mut self,
```
k88hudson-cfa, line 270: Having `get_person_property` require a mutable reference is probably not what we want – that would be a pretty big changed (since we want to provide read access to properties in scenarios where you only have an immutable reference to context)

src/people.rs
```diff
@@ -230,13 +323,18 @@ impl ContextPeopleExt for Context {
         }
 
         // Initialize the property. This does not fire a change event
-        let initialized_value = T::initialize(self, person_id);
+        // ok to call unwrap here because the property that returns None from get_person_property_ref
+        // in this case (i.e., hasn't been initialized yet) has a user-specified default/initializer
+        // and will return a some variant accordingly
+        let initialized_value = T::initialize(self, person_id).expect("Property not initialized");
```
k88hudson-cfa, line 329: So while I think overall I'm going to recommend a different approach here, I wanted to give you some feedback on what would be a better way (IMO) to handle the mutability issues you're running into here –

What I would do here is have `include_in_periodic_report` be wrapped in a `RefCell` so that you could borrow it mutably without needing a mutable reference to context. That way get_person_property could still take an immutable reference.

k88hudson-cfa (changes requested):
I left a few comments in line and I think this is a good start but I think it might be worth taking a slightly different approach. Here's what I'd recommend:

- I would define the computer function as part of a `register_periodic_report` API method (i.e., an extension to context) rather than tied tightly to each `PersonProperty` 
- Internally, this would depend on some implementation of "phases" as in here: https://github.com/CDCgov/ixa/pull/39/files – I'd recommend doing that part first in a separate PR (the requirements being: a first/normal/last phase, plans get added to Normal by default but you can use add_plan_to_phase with a phase name instead)

I think moving it to an API method would also simplify registration quite a bit (since you would just expect people to initialize it manually) and we wouldn't really need automatic registration, although I'm open to other arguments.

ChiragKumar9:
> I left a few comments in line and I think this is a good start but I think it might be worth taking a slightly different approach. Here's what I'd recommend:

Sounds good! I appreciate the comments here -- I was more just hacking around and pleasantly surprised at how far I got in recreating the person properties periodic report from GCM. Was pretty sure this was not the right approach, and happy to switch to an alternative approach.

> * I would define the computer function as part of a `register_periodic_report` API method (i.e., an extension to context) rather than tied tightly to each `PersonProperty`

I thought through this design a bit, and the problem I had with it is that you would need to specify all person properties explicitly, right? That doesn't seem ideal to me when it could be done automatically and more importantly at the site of property initialization. However, I may not be understanding your idea properly.

> * Internally, this would depend on some implementation of "phases" as in here: https://github.com/CDCgov/ixa/pull/39/files – I'd recommend doing that part first in a separate PR (the requirements being: a first/normal/last phase, plans get added to Normal by default but you can use add_plan_to_phase with a phase name instead)

Yes! Wholeheartedly agreed. I added a test to #39 that I think users may find useful (since some of our users will honestly just read the tests to figure out how things work) and that's awaiting your review. Happy to make further changes there too.

I also think that even after that, pulling out some changes from here like adding some of the helper methods to context that see if there are more plans should be landed as PRs before this one.

> I think moving it to an API method would also simplify registration quite a bit (since you would just expect people to initialize it manually) and we wouldn't really need automatic registration, although I'm open to other arguments.

k88hudson-cfa:
> I thought through this design a bit, and the problem I had with it is that you would need to specify all person properties explicitly, right? That doesn't seem ideal to me when it could be done automatically and more importantly at the site of property initialization. However, I may not be understanding your idea properly.

I can see how doing it at the site of property initialization would be nice in some cases, but I think there's plenty of scenarios where it wouldn't be – for example, if you wanted to reuse a person property across models but have different reporting, or maybe even if you have different kinds of reports that have different computations. I guess it's a judgment call but I don't think the alternative (i.e., just specifying them all explicitly) is that bad. 

>Yes! Wholeheartedly agreed. I added a test to https://github.com/CDCgov/ixa/pull/39 that I think users may find useful (since some of our users will honestly just read the tests to figure out how things work) and that's awaiting your review. Happy to make further changes there too.

Oh sweet didn't realize. Will take a look :)

ekr-cfa:
> I can see how doing it at the site of property initialization would be nice in some cases, but I think there's plenty of scenarios where it wouldn't be – for example, if you wanted to reuse a person property across models but have different reporting, or maybe even if you have different kinds of reports that have different computations. I guess it's a judgment call but I don't think the alternative (i.e., just specifying them all explicitly) is that bad.

I agree with Kate here. I don't think that the authors of properties are in a position to know which reports people want and I don't want random property authors littering my reports directory with stuff I don't want.

ChiragKumar9:
Ok! I am convinced on this front! This version would make registration much easier. Looking forward to diving into it!

PR #67, opened 2024-10-31
generate parent folders for report writing file if missing

ChiragKumar9:


src/report.rs
```diff
@@ -32,7 +32,11 @@ impl ConfigReportOptions {
         self
     }
     /// Sets the directory where reports will be output
+    /// # Panics
+    /// if the directory in which the report is to be stored does not exist and cannot be created
     pub fn directory(&mut self, directory: PathBuf) -> &mut ConfigReportOptions {
+        // if the directory does not exist, create it
+        create_dir_all(directory.clone()).expect("Failed to create directory");
```
ekr-cfa, line 39:
```suggestion
        create_dir_all(&directory).expect("Failed to create directory");
```

ekr-cfa (changes requested):
This shouldn't panic. Instead, we need to bubble up `Result<IxaError>`

ekr-cfa:
Thanks for making changes.

Why are you exposing this function and making all callers use it instead of doing it internally in `add_report()`?

ChiragKumar9:
> Why are you exposing this function and making all callers use it instead of doing it internally in `add_report()`?

This sounds to me like a restructure of the current design? So `add_report()` would take care of all details of report output path, rendering the `ConfigReportOptions` structure no longer necessary? I am happy to do this, and in a way I think this is much more parsimonious, but I did not want to immediately overwrite the existing structure. Thanks!

ekr-cfa:
I don't think I'm advocating for that. Currently, `add_report()` invokes `generate_filename()` which looks at `ConfigReportOptions`. It then tries to open the file, and if the directory doesn't exist, it panics (as an aside this should also return `Result<IxaError>`. What I'm saying is just that if we want auto-creation, it should create the directory in `add_report()`

ChiragKumar9:
Thanks for the explanation -- this really helps. I follow and see exactly what you're saying. I played around a bit with this, and here's the structure I ended up using in `main.rs` that I think feels quite nice:

```rust
fn init() -> Result<Context, IxaError> {
   parameters_loader::init_parameters(...)?;
   // call all the init functions
   context.add_report::<ReportType)(...)?;
}

fn main() {
   let context_result = init();
   match context_result {
      Ok(mut context) => {
         context.execute();
      }
      Err(ixa_error) => println!("Initializaiton failed: {ixa_error}");
   }
}
```

I am open to discussion on this, but I think having auto-creation is just a really helpful feature to have embedded. When I'm experimenting with parameters and running experiments, it's helpful to have folder generation taken care of by running the model and basically specified in the output path provided and not need to write some Python or bash code to do it externally.

I wasn't here when report options was developed, but I actually want to push a bit on the idea from above: why not have the explicit filepath as an input to `add_report()` and cut the need to specify the directory separately/the optional prefix? You can get out all the same functionality of specifying the directory and applying some sort of prefix in giving the filepath as an argument to `add_report()`, so no functionality would be cut. From a model running organization perspective, we plan on having a Python package that generates the full Ixa json file from input tomls, so really it's the Python package that needs to have an easy-to-understand interface for setting up output paths. Plus, I think having a complete filepath in one place is more explicit to the user and overall cleaner.

Appreciate the discussion around this, and would be happy to discuss on Monday in more depth.

ekr-cfa:
I'm not sure I follow the API you have in mind. Would you mind sketching it out?

ChiragKumar9:
Of course. What do you think about the pattern in c4ac918 (head on this branch)?

ekr-cfa:
So you mean the caller would do the following?

```
    context.add_report::<IncidenceReportItem>("./examples/basic-infection/incidence.csv")?;
```

Doesn't this just end up with the caller doing ...


```
let base_directory= String::from("./examples/basic-infection/");
context.add_report::<IncidenceReportItem>("./examples/basic-infection/incidence.csv")?;
context.add_report::<DeathReportItem>("./examples/basic-infection/incidence.csv")?;
context.add_report::<RecoveryReportItem>("./examples/basic-infection/incidence.csv")?;
```

ChiragKumar9:
Yes, and I see what you're saying about redundancy in needing to reference the directory multiple times (assuming all reports go to the same directory). But, if the directory/file creation is being done in `add_report()` and that can return an error, shouldn't the argument to `add_report()` be explicit to include the directory rather than having that set up above somewhere?

I recognize that I am perhaps thinking too much about a small point, so feel free to overrule here.

ekr-cfa:
> Yes, and I see what you're saying about redundancy in needing to reference the directory multiple times (assuming all reports go to the same directory). But, if the directory/file creation is being done in `add_report()` and that can return an error, shouldn't the argument to `add_report()` be explicit to include the directory rather than having that set up above somewhere?
> 
> I recognize that I am perhaps thinking too much about a small point, so feel free to overrule here.

Can you provide an example of the API you have in mind?

ChiragKumar9:
As I understand, there are two options we are discussing:

```rust
fn init() -> Result<(), IxaError> {
    // sets the directory, does not create the directory,
    // so does not error (just changes the value of config_report_options.directory)
    context.report_options().directory("./read-only-directory");
    // ...
    // add the actual report
    // could error out because of a directory creation problem
    // from the directory set in the config_report_options above
    context.add_report::<Incidence>("incidence")?;
    context.report_options().directory("./results");
    context.add_report::<RaceTimes>("times")?;
}
```

If `add_report()` is going to handle directory creation and that can error, I feel that it is more logical to specify the directory directly to it:

```rust
fn init() -> Result<(), IxaError> {
    // specify the directory to the method which creates it and can
    // error out with a directory creation error
    context.add_report::<Incidence>("./read-only-directory/incidence.csv")?;
    // this would mean that the user would have to include the directory
    // again in each call to add_report but it is more explicit to the arguments
    // that may cause an error being directly provided to add_report()
    context.add_report::<RaceTime>("./results/times.csv")?; 
}
```

ChiragKumar9:
After a discussing with @ekr-cfa, we are closing this PR for now.

PR #73, opened 2024-11-04
Change |get_global_property()| to return Option<&T::Value>. Fixes #72

### LINKED ISSUE #72: Missing global properties should produce an IxaError
Right now, if you ask for a global property that doesn't exist, you get a panic. This should instead return an IxaError/

ekr-cfa:
Returns None when either:

- No global properties have been set
- Global properties exist but the requested property has not been set.


confunguido (approved):
This makes sense to me.

k88hudson-cfa (approved):

PR #79, opened 2024-11-10
Initialize properties in add_person

ekr-cfa:
This PR adds the ability to initialize properties in `add_person()`, as in:

```
       let person = context.add_person(((Age, 17), (IsSwimmer, true))).unwrap()
```

Any non-derived property that doesn't have automatic initialization must be initialized in this fashion during creation, which means we have removed `initiialize_person_property()`. This is technically enforced by checking that any known such property is present in the initialization list. This enforcement is imperfect because `Context` doesn't know about any property which has never been used. This means that if you do `add_person()` followed by `get_person_property()` for such a property, it will panic. However, as long as there is any use of a property this invariant will instead be enforced in `add_person()`

### DIFF
--- a/src/people.rs
+++ b/src/people.rs
@@ -774,19 +774,19 @@ impl ContextPeopleExt for Context {
         let (previous_value, deps_temp) = if initializing {
             (None, None)
         } else {
-
             let previous_value = self.get_person_property(person_id, property);
             if previous_value != value {
-                    self.remove_from_index_maybe(person_id, property);
+                self.remove_from_index_maybe(person_id, property);
             }
-            
-            ( Some(previous_value),
-              self.get_data_container(PeoplePlugin)
-              .unwrap()
-              .dependency_map
-              .borrow_mut()
-              .get_mut(&TypeId::of::<T>())
-              .map(std::mem::take),
+
+            (
+                Some(previous_value),
+                self.get_data_container(PeoplePlugin)
+                    .unwrap()
+                    .dependency_map
+                    .borrow_mut()
+                    .get_mut(&TypeId::of::<T>())
+                    .map(std::mem::take),
             )
         };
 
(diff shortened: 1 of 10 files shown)


src/people.rs
```diff
@@ -285,6 +380,20 @@ impl PeopleData {
         let mut property_ref = self.get_person_property_ref(person_id, property);
         *property_ref = Some(value);
     }
+
+    fn check_initialization_list<T: InitializationList>(
+        &self,
+        initialization: &T,
+    ) -> Result<(), IxaError> {
+        let properties_map = self.properties_map.borrow();
+        for (t, property) in properties_map.iter() {
```
k88hudson-cfa, line 389: Probably worth a comment that this only checks properties that have been set at least once

src/people.rs
```diff
@@ -361,10 +462,25 @@ impl ContextPeopleExt for Context {
...
+        // actually was.
+        data_container.check_initialization_list(&props)?;
+
+        // Actually add the person. Nothing can fail after this point because
+        // it would leave the person in an inconsistent state.
+        let person_id = data_container.add_person();
+
+        // Initialize the properties. We set |is_initializing| to prevent
+        // set_person_property() from generating an event.
+        data_container.is_initializing = true;
```
k88hudson-cfa, line 477: Wdyt about using a with_initializing method that takes a closure or something like that?

src/people.rs
```diff
@@ -64,13 +82,75 @@ pub trait PersonProperty: Copy {
         false
     }
     #[must_use]
+    fn must_be_initialized() -> bool {
```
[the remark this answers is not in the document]
ekr-cfa: I don't think that this is correct, because the semantics of "must_be_initialized" is that you need to supply an initializer which is of course forbidden for derived properties.

src/people.rs
ekr-cfa: I don't think I understand what you are proposing here.

ekr-cfa:
@k88hudson-cfa I added a comment and responded to your other questions. LMK if you want to further discuss those.

k88hudson-cfa (approved):

src/people.rs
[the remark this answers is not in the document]
ekr-cfa: How about we rename it.

ekr-cfa:
@k88hudson-cfa rebased, errors fixed. 

I renamed `must_be_initialized` to `is_required`.

PR #81, opened 2024-11-13
Use refs instead of copy for query_people_internal

k88hudson-cfa:


src/people.rs
```diff
@@ -471,12 +470,21 @@ impl PeopleData {
...
+    fn get_index_ref(&self, t: TypeId) -> Option<Ref<Index>> {
+        let index_map = self.property_indexes.borrow();
+        if index_map.contains_key(&t) {
+            Some(Ref::map(index_map, |map| map.get(&t).unwrap()))
+        } else {
+            None
+        }
+    }
+
     fn get_index_ref_by_prop<T: PersonProperty + 'static>(
```
jasonasher, line 482: Consider calling this ref_mut too as it gives a mutable reference

src/people.rs
```diff
@@ -807,21 +815,24 @@ impl ContextPeopleExtInternal for Context {
...
+        // 1. Walk through each property and update the indexes.
+        for (t, _) in &property_hashes {
+            let mut index = data_container.get_index_ref_mut(*t).unwrap();
             index.index_unindexed_people(self);
+        }
 
+        // 2. Collect the index entry corresponding to the value.
+        for (t, hash) in property_hashes {
+            let index = data_container.get_index_ref(t).unwrap();
             // Update the index.
```
jasonasher, line 832: I think this comment isn't needed anymore as the updating happens earlier.

src/people.rs
```diff
@@ -1368,7 +1379,7 @@ mod test {
         context
             .get_data_container(PeoplePlugin)
             .unwrap()
-            .get_index_ref(TypeId::of::<T>())
+            .get_index_ref_mut(TypeId::of::<T>())
```
jasonasher, line 1382: I think you don't need the ref_mut here and the original code will work as this function doesn't need the mutability.

jasonasher (approved):
LGTM - minor comments that can easily be addressed.

PR #82, opened 2024-11-13
Add periodic report

ChiragKumar9:
Self scheduling periodic report that implements a given closure with a given periodicity.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -28,7 +28,7 @@ pub trait IxaEvent {
 /// handled after all `Normal` plans. In all cases ties between plans at the
 /// same time and with the same phase are handled in the order of scheduling.
 ///
-#[derive(PartialEq, Eq, Ord, PartialOrd)]
+#[derive(PartialEq, Eq, Ord, Clone, Copy, PartialOrd)]
 pub enum ExecutionPhase {
     First,
     Normal,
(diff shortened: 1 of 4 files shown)


src/report.rs
```diff
@@ -138,6 +166,26 @@ impl ContextReportExt for Context {
         let data_container = self.get_data_container_mut(ReportPlugin);
         &mut data_container.config
     }
+
+    /// Adds a periodic report to the context. The report will be output every `report_period` time
+    /// and will print out the output from the `report_output_fcn` function.
+    fn add_periodic_report<T: Report + 'static, C: Fn(&Context) -> T + 'static>(
```
k88hudson-cfa, line 172: I think this would actually be better as a more generic `add_periodic_plan_with_priority`: would you mind moving it there instead? you could still have a test/example for adding the report and then scheduling it. 

Also, I think then C can just be generic then (like other plans).

LGTM with that change

k88hudson-cfa (changes requested):

src/context.rs
```diff
@@ -158,6 +158,50 @@ impl Context {
...
+    /// if needed.
+    /// # Panics
+    ///
+    /// Panics if time is negative, infinite, or NaN.
+    pub fn add_periodic_plan_with_phase(
+        &mut self,
+        period: f64,
+        callback: impl Fn(&mut Context) + 'static,
+        phase: ExecutionPhase,
+    ) {
```
ChiragKumar9, line 192: My claim is that periodic plans should not return `Id` because that `Id` has no meaning after the first period's plan has run.

src/report.rs
```diff
@@ -213,12 +213,8 @@ mod test {
     #[test]
     fn add_report_no_dir() {
```
ChiragKumar9, line 214: I rewrote this test to be true to its name (previously it was just a repeat of the test above). I realize I should mayb pull this out and put in its own PR?

PR #84, opened 2024-11-18
Initial implementation of a network facility.

ekr-cfa:
A network is modeled as a directed graph.  Edges are typed in the usual fashion, i.e., keyed by a Rust type, and each person can have an arbitrary number of outgoing edges of a given type, with each edge having a weight, as in:

        context
            .add_edge::<EdgeType1>(person1, person2, 0.01);
        context
            .add_edge::<EdgeType1>(person1, person3, 0.02);

Which creates a graph where `person1` has two outgoing edges of type `EdgeType`, to `person2` and `person3`.

Edges are directional but `add_edge_bidi()` will create one edge in each direction between two people with the same weight for both edges. If you want to create edges with different weights, you have to do it separately.

Creating an edge to yourself is forbidden (returns IxaError). It is not technically possible to create an edge to a person who doesn't exist because `PersonId` is minted by `add_person()`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -29,6 +29,7 @@
 pub mod context;
 pub mod error;
 pub mod global_properties;
+pub mod network;
 pub mod people;
 pub mod plan;
 pub mod random;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,7 +5,7 @@
 //! agent-based models for disease transmission, but the approach is applicable
 //! in a wide array of circumstances.
 //!
-//! The central object of an Eosim simulation is the `Context` that is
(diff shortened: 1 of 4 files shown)


ekr-cfa:
Note that I did not follow the convention of using an indicator struct to determine the edge type, as in:

```
context.add_edge(EdgeType1, person1, person2, 0.01);
```

That's technically straightforward, but it seemed like the turbofish did a slightly better job of showing what edge type was in use, and that was better than perfect consistency. If others disagree, I can easily change it.

k88hudson-cfa (approved):

PR #85, opened 2024-11-20
Load global properties from a configuration file

ekr-cfa:
This PR adds a facility to load arbitrary global properties from a configuration file.

The way this works is that Ixa automatically knows about all global properties defined with `define_global_property!()`, no matter which crate they are defined in. The config file is formatted as a JSON dictionary with the top-level keys being the fully qualified names of global properties, like so:

```
{
  "ixa::global_properties::test::Property1": {
    "field_int": 1,
    "field_str": "test"
  },

  "ixa::global_properties::test::Property2": {
    "field_int": 2
  }
}
```
`context.load_global_properties()` reads the file and then uses the name to find the associated global property and deserialize the associated object into the global property using serde, where it can then be accessed in the usual fashion with `get_global_property()`.

You can call `load_global_properties()` an arbitrary number of times with different files, but you can't load the same property twice, so the files can't overlap.

### DIFF
--- a/src/global_properties.rs
+++ b/src/global_properties.rs
@@ -7,6 +7,36 @@ use std::fmt::Debug;
 use std::fs;
 use std::io::BufReader;
 use std::path::Path;
+use std::cell::RefCell;
+use std::sync::Mutex;
+use std::sync::Arc;
+use std::sync::LazyLock;
+
+type PropertySetterFn = dyn Fn(&mut Context, serde_json::Value)-> Result<(), IxaError> + Send + Sync;
+
+pub static GLOBAL_PROPERTIES: LazyLock<Mutex<RefCell<HashMap<String, Arc<PropertySetterFn>>>>> = LazyLock::new(|| {
+    Mutex::new(RefCell::new(HashMap::new())) });
+
+
+#[allow(clippy::missing_panics_doc)]
+pub fn add_global_property<T: GlobalProperty> (name: &String) where for<'de> <T as GlobalProperty>::Value: serde::Deserialize<'de>{
+    let properties = GLOBAL_PROPERTIES.lock().unwrap();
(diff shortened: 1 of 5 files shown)


src/global_properties.rs
```diff
@@ -2,11 +2,53 @@ use crate::context::Context;
...
 use std::path::Path;
+use std::sync::Arc;
+use std::sync::LazyLock;
+use std::sync::Mutex;
+
+type PropertySetterFn =
+    dyn Fn(&mut Context, &str, serde_json::Value) -> Result<(), IxaError> + Send + Sync;
+
+#[allow(clippy::type_complexity)]
+pub static GLOBAL_PROPERTIES: LazyLock<Mutex<RefCell<HashMap<String, Arc<PropertySetterFn>>>>> =
```
k88hudson-cfa, line 19: This could probably use some explanation in a comment

k88hudson-cfa (changes requested):
As discussed if you an get this to work with just a namespace rather than the fully qualified path I think that would definitely be preferable. I think the API is pretty reasonable here

src/global_properties.rs
ekr-cfa: CoPilot says :"The GLOBAL_PROPERTIES variable is a static global variable that holds a thread-safe, lazily-initialized, and mutable reference to a HashMap. The HashMap maps string keys to Arc-wrapped function pointers (PropertySetterFn). This variable is defined using the LazyLock type to ensure it is only initialized once and is safe to access from multiple threads. The Mutex ensures mutual exclusion for modifying the HashMap, while the RefCell allows for interior mutability, enabling modifications to the HashMap even when it is accessed through a shared reference."

ekr-cfa: I'll obviously clean that up.

ekr-cfa:
@k88hudson-cfa ready for re-review

PR #96, opened 2024-11-25
More error types

ChiragKumar9:
I just did this the same way I did it for `IOError` before. But, let me know if there's something else wanted here. Thanks!


k88hudson-cfa (approved):
Thanks

PR #99, opened 2024-11-26
Ekr validate global properties

ekr-cfa:


tests/data/global_properties_missing.json
confunguido, line 1: Where is this file being used?

confunguido: nvm, I guess that's not part of this PR.

ekr-cfa: Correct. I just had to do a rename.

k88hudson-cfa (approved):

PR #102, opened 2024-11-27
Added ParseIntError to IxaError

KOVALW:
Added the ParseIntError to the IxaError `enum` to propagate errors expected in ixa-epi-isolation when selecting digits from integers as strings.


ChiragKumar9 (approved):
Thanks

PR #108, opened 2024-11-29
Periodic reports

k88hudson-cfa:


src/people.rs
```diff
@@ -167,6 +167,68 @@ seq!(Z in 1..20 {
...
+impl<T: PersonProperty + 'static> Tabulator for (T,) {
+    fn setup(&self, context: &mut Context) {
+        context.index_property_noargs::<T>();
+    }
+    fn get_typelist(&self) -> Vec<TypeId> {
+        vec![std::any::TypeId::of::<T>()]
+    }
+    fn get_columns(&self) -> Vec<String> {
+        vec![String::from(
+            std::any::type_name::<T>().split("::").last().unwrap(),
```
ekr-cfa, line 185: You actually can't guarantee that types will be named like this, as it's a compiler internal. `type_name::<T>()` could be legally implemented as `format!("Banana|{:?}", TypeId::of::<T>())`. I think instead we need to extend `PersonProperty` to have a name field.

src/people.rs
```diff
@@ -167,6 +167,68 @@ seq!(Z in 1..20 {
...
+                    #(
+                        std::any::TypeId::of::<T~N>(),
+                    )*
+                    ]
+                }
+
+                fn get_columns(&self) -> Vec<String> {
+                    vec![
+                    #(
+                        String::from(std::any::type_name::<T~N>().split("::").last().unwrap()),
```
ekr-cfa, line 219: Same comment here as above.

src/people.rs
```diff
@@ -214,28 +276,31 @@ impl Index {
             lookup: None,
             indexer: Box::new(move |context: &Context, person_id: PersonId| {
                 let value = context.get_person_property(person_id, property);
-                IndexValue::compute(&value)
+                (IndexValue::compute(&value), format!("{value:?}"))
```
ekr-cfa, line 279: Architecture: I'm not sure we want to compute the string every time we do an index. This is in an inner loop. I would prefer instead to have a separate formatter and then only compute the string the first time.

Are you sure we want to use Debug and not Display here?

src/people.rs
```diff
@@ -199,7 +261,7 @@ struct Index {
     name: &'static str,
     // The hash of the property value maps to a list of PersonIds
     // or None if we're not indexing
-    lookup: Option<HashMap<IndexValue, HashSet<PersonId>>>,
+    lookup: Option<HashMap<IndexValue, (HashSet<PersonId>, String)>>,
```
ekr-cfa, line 264: Nit: I would put the String first to make it a little more readable.

src/people.rs
```diff
@@ -262,7 +327,7 @@ impl StoredPeopleProperties {
     fn new<T: PersonProperty + 'static>() -> Self {
         StoredPeopleProperties {
             is_required: T::is_required(),
-            values: Box::new(Vec::<Option<T::Value>>::new()),
+            values: Box::<Vec<Option<T::Value>>>::default(),
```
ekr-cfa, line 330: Is there a difference here?

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
     /// `context.query_people(((Age, 30), (Gender, Female)))`.
     fn query_people<T: Query>(&self, q: T) -> Vec<PersonId>;
     fn match_person<T: Query>(&self, person_id: PersonId, q: T) -> bool;
+    fn get_counts<T: Tabulator, F>(&self, tabulator: &T, print_fn: F)
```
ekr-cfa, line 864: Architecture:
I wouldn't call this `get_counts()` but rather `tabulate_person_properties()` or the like.

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+    where
+        F: Fn(&Context, &[String], usize);
+}
+
+fn process_indices(
+    context: &Context,
+    remaining_indices: &[&Index],
+    accumulated_values: &mut [IndexValue],
+    display_values: &mut Vec<String>,
```
ekr-cfa, line 873: I would call these `property_names`

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+    current_matches: &mut HashSet<PersonId>,
+    intersect: bool,
```
ekr-cfa, line 875: Do you need this? Can't you instead just check for `display_values.is_empty()`

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+    print_fn: &dyn Fn(&Context, &[String], usize),
+) {
+    if remaining_indices.is_empty() {
+        print_fn(context, display_values, current_matches.len());
+        return;
+    }
+
+    if let Some((next_index, rest_indices)) = remaining_indices.split_first() {
```
ekr-cfa, line 883: `(next_index, rest_indices)) = remaining_indices.split_first().unwrap()`. This can't fail, so why have a conditional

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+        let lookup = next_index.lookup.as_ref().unwrap();
+
+        // If there is nothing in the index, we don't need to process it
+        if lookup.is_empty() {
+            return;
+        }
```
ekr-cfa, line 889: The implication of this is that you don't dump the cartesian product of properties, but just the ones that have nonzero counts. For instance, say we have:

```
Gender,Over50
M, True
F, True
M, False
```

You would emit:

```
Gender,Over50,Count
M, True, 1
F, True, 1
M, False, 1
```
But not

```
M, False, 0
```
This seems like an understandable decision but potentially inconvenient for analysis. Do you think we should have a flag?

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+
+        for (value, (people, display)) in lookup {
+            let mut updated_values = accumulated_values.to_owned();
+            updated_values.push(value.clone());
```
ekr-cfa, line 893: It looks to me like you don't use `accumulated_values` except to maintain the stack here, so can you just remove it?

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+
+    if let Some((next_index, rest_indices)) = remaining_indices.split_first() {
+        let lookup = next_index.lookup.as_ref().unwrap();
+
+        // If there is nothing in the index, we don't need to process it
+        if lookup.is_empty() {
+            return;
+        }
+
+        for (value, (people, display)) in lookup {
```
ekr-cfa, line 891: `value` appears to be unused.

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+    where
+        F: Fn(&Context, &[String], usize);
+}
+
+fn process_indices(
+    context: &Context,
+    remaining_indices: &[&Index],
+    accumulated_values: &mut [IndexValue],
+    display_values: &mut Vec<String>,
+    current_matches: &mut HashSet<PersonId>,
```
ekr-cfa, line 874: This doesn't need to be `mut`, because each intersect creates a copy anyway.

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+            let mut updated_values = accumulated_values.to_owned();
+            updated_values.push(value.clone());
+            display_values.push(display.clone());
+
+            let mut matches = if intersect {
+                current_matches
+                    .iter()
+                    .filter(|person| people.contains(person))
+                    .copied()
+                    .collect()
```
ekr-cfa, line 901: Replace this with `current_matches.intersection(&people).collect()`

src/people.rs
```diff
@@ -789,6 +861,61 @@ pub trait ContextPeopleExt {
...
+            } else {
+                people.clone()
```
ekr-cfa, line 903: Why do you have to clone this? You can just pass the ref below, I think.

src/people.rs
```diff
@@ -167,6 +167,68 @@ seq!(Z in 1..20 {
...
 
+pub trait Tabulator {
+    fn setup(&self, context: &mut Context);
+    fn get_typelist(&self) -> Vec<TypeId>;
+    fn get_columns(&self) -> Vec<String>;
+}
+
+impl<T: PersonProperty + 'static> Tabulator for (T,) {
+    fn setup(&self, context: &mut Context) {
+        context.index_property_noargs::<T>();
```
ekr-cfa, line 178: Is there a reason not to do `T::instance()`

src/people.rs
```diff
@@ -997,6 +1130,49 @@ impl ContextPeopleExt for Context {
...
         }
     }
+
+    fn get_counts<T: Tabulator, F>(&self, tabulator: &T, print_fn: F)
+    where
+        F: Fn(&Context, &[String], usize),
+    {
+        let type_ids = tabulator.get_typelist();
+
+        // First, update indexes
```
ekr-cfa, line 1140: Since we have to do this here and in query_people, can we pull it out into a function?

src/report.rs
```diff
@@ -451,10 +525,51 @@ mod test {
...
+        });
+
+        context.execute();
+
+        let file_path = path.join("test_periodic.csv");
+        assert!(file_path.exists(), "CSV file should exist");
+
+        let mut reader = csv::Reader::from_path(file_path).unwrap();
+
+        assert_eq!(reader.headers().unwrap(), vec!["t", "IsRunner", "count"]);
```
ekr-cfa, line 558: Should floats be rendering without a decimal?

ekr-cfa (changes requested):

src/people.rs
```diff
@@ -252,7 +250,7 @@ impl Hasher for IndexValueHasher {
     }
 }
 
-type Indexer = dyn Fn(&Context, PersonId) -> (IndexValue, String);
+type PersonCallback<T> = dyn Fn(&Context, PersonId) -> T;
```
ekr-cfa, line 253: I guess it's a taste issue, but I probably would just define two functions here.

src/people.rs
```diff
@@ -261,9 +259,11 @@ struct Index {
     name: &'static str,
     // The hash of the property value maps to a list of PersonIds
     // or None if we're not indexing
-    lookup: Option<HashMap<IndexValue, (HashSet<PersonId>, String)>>,
+    lookup: Option<HashMap<IndexValue, (String, HashSet<PersonId>)>>,
     // A callback that calculates the IndexValue of a person's current property value
-    indexer: Box<Indexer>,
+    indexer: Box<PersonCallback<IndexValue>>,
+    // A callback that calculates the display value of a person's current property value
+    get_display: Box<PersonCallback<String>>,
```
ekr-cfa, line 266: Maybe `get_display_string()` or `render_as_display()`

src/people.rs
```diff
@@ -710,18 +724,12 @@ impl PeopleData {
...
-        self.get_index_ref_mut(type_id)
-    }
-
     fn get_index_ref_mut_by_prop<T: PersonProperty + 'static>(
         &self,
         _property: T,
     ) -> Option<RefMut<Index>> {
-        self.get_index_ref_mut_by_prop_noargs::<T>()
+        let type_id = TypeId::of::<T>();
+        self.get_index_ref_mut(type_id)
```
ekr-cfa, line 732: I would probably merge these lines

src/report.rs
```diff
@@ -367,25 +363,28 @@ mod test {
...
-            value: "Value,1".to_string(),
-        };
-        let report2 = SampleReport {
-            id: 2,
-            value: "Value\n2".to_string(),
-        };
-
-        context.send_report(report1);
-        context.send_report(report2);
+        // We need the writer to go out of scope so the file is flushed
```
ekr-cfa, line 368: Maybe do `drop context` instead, here and below.

ekr-cfa (approved):
LGTM

PR #110, opened 2024-11-29
update docs

ChiragKumar9:
Now that we have a docs webpage, I noticed some places where the docs I had written were.... not detailed enough. I have attempted to correct that here. Thanks.


k88hudson-cfa (approved):
👍

PR #114, opened 2024-12-02
Add runner with arg parsing

k88hudson-cfa:
I still need to add custom arg parsing, but here is a start to a runner with some base arguments and generic setup functionality.

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,0 +1,224 @@
+use std::path::{Path, PathBuf};
+
+use crate::context::Context;
+use crate::error::IxaError;
+use crate::global_properties::ContextGlobalPropertiesExt;
+use crate::random::ContextRandomExt;
+use crate::report::ContextReportExt;
+use clap::{Args, Command, FromArgMatches as _};
+
+/// Default cli arguments for ixa runner
+#[derive(Args, Debug)]
+pub struct BaseArgs {
+    /// Random seed
+    #[arg(short, long, default_value = "0")]
+    pub random_seed: u64,
+
+    /// Optional path for a global properties config file
+    #[arg(short, long, default_value = "")]
+    pub config: String,
+
+    /// Optional path for report output
+    #[arg(short, long, default_value = "")]
+    pub output_dir: String,
+}
+
+#[derive(Args)]
+pub struct PlaceholderCustom {}
+
+fn create_ixa_cli() -> Command {
+    let cli = Command::new("ixa");
+    BaseArgs::augment_args(cli)
+}
+
+/// Runs a simulation with custom cli arguments.
+///
+/// This function allows you to define custom arguments and a setup function
+///
+/// # Parameters
+/// - `setup_fn`: A function that takes a mutable reference to a `Context`, a `BaseArgs` struct,
+///    a Option<A> where A is the custom cli arguments struct
+///
+/// # Errors
+/// Returns an error if argument parsing or the setup function fails
+#[allow(clippy::missing_errors_doc)]
+pub fn run_with_custom_args<A, F>(setup_fn: F) -> Result<(), Box<dyn std::error::Error>>
+where
+    A: Args,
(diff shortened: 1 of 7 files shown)


src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+use crate::global_properties::ContextGlobalPropertiesExt;
+use crate::random::ContextRandomExt;
+use crate::report::ContextReportExt;
+use clap::{Args, Command, FromArgMatches as _};
+
+#[derive(Args, Debug)]
+pub struct BaseArgs {
+    /// Random seed
+    #[arg(short, long, default_value = "0")]
+    pub seed: u64,
```
ekr-cfa, line 14: I would probably call this `random_seed`

src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Optional path for report output
+    #[arg(short, long, default_value = "")]
+    pub output_dir: String,
+}
+
+#[derive(Args)]
+pub struct PlaceholderCustom {}
+
+#[allow(clippy::missing_errors_doc)]
+pub fn run_with_custom_args<A, F>(load: F) -> Result<(), Box<dyn std::error::Error>>
```
ekr-cfa, line 34: I don't love "load" as a name. Maybe "setup_fn"?

src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+where
+    A: Args,
+    F: Fn(&mut Context, BaseArgs, Option<A>) -> Result<(), IxaError>,
+{
+    let cli = Command::new("Ixa");
```
ekr-cfa: Shouldn't this instead be "ixa"

src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+    let base_args_matches = BaseArgs::from_arg_matches(&matches)?;
+    let custom_matches = A::from_arg_matches(&matches)?;
+
+    run_with_args_internal(base_args_matches, Some(custom_matches), load)
+}
+
+#[allow(clippy::missing_errors_doc)]
+pub fn run_with_args<F>(load: F) -> Result<(), Box<dyn std::error::Error>>
+where
+    F: Fn(&mut Context, BaseArgs, Option<PlaceholderCustom>) -> Result<(), IxaError>,
```
ekr-cfa, line 51: Can you just do Option<()>. It doesnt' seem like there are any trait bounds.

src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+    args: BaseArgs,
+    custom_args: Option<A>,
+    load: F,
+) -> Result<(), Box<dyn std::error::Error>>
+where
+    A: Args,
+    F: Fn(&mut Context, BaseArgs, Option<A>) -> Result<(), IxaError>,
+{
+    // Create a context
+    let mut context = setup_context(&args)?;
```
ekr-cfa, line 92: Is there a reason not to inline setup_context, which seems to only be called from ehre.

src/runner.rs
```diff
@@ -0,0 +1,174 @@
...
+    struct CustomArgs {
+        #[arg(short, long, default_value = "0")]
+        field: u32,
+    }
+
+    #[test]
+    fn test_run_with_custom_args() {
+        let result = run_with_custom_args(|_, _, _: Option<CustomArgs>| Ok(()));
+        assert!(result.is_ok());
+    }
```
ekr-cfa, line 118: These tests don't seem to verify that any of the arguments exist or are correct. It seems like that's really challenging, though.

ekr-cfa (changes requested):

src/runner.rs
k88hudson-cfa: I have `A` implementing `Args` but i guess for this function it's not actually necessary

src/runner.rs
k88hudson-cfa: it would probably be good to have some integration tests yes, would you be ok if I file that

ekr-cfa (approved):
LGTM

PR #123, opened 2024-12-13
MVP debugger

k88hudson-cfa:
This PR adds a basic cli debugger

### DIFF
--- a/src/debugger.rs
+++ b/src/debugger.rs
@@ -12,9 +12,8 @@ trait DebuggerCommand {
     fn handle(
         &self,
         context: &mut Context,
-        cli: &DebuggerRepl,
         matches: &ArgMatches,
-    ) -> Result<bool, String>;
+    ) -> Result<(bool, Option<String>), String>;
     fn about(&self) -> &'static str;
     fn extend(&self, subcommand: Command) -> Command {
         subcommand
@@ -69,7 +68,7 @@ impl DebuggerRepl {
     }
 
     fn process_line(&self, l: &str, context: &mut Context) -> Result<bool, String> {
-        let args = shlex::split(l).ok_or("error: Invalid quoting")?;
+        let args = shlex::split(l).ok_or("Error splitting lines")?;
         let matches = self
             .build_cli()
             .try_get_matches_from(args)
@@ -78,7 +77,11 @@ impl DebuggerRepl {
         if let Some((command, sub_matches)) = matches.subcommand() {
             // If the provided command is known, run its handler
             if let Some(handler) = self.get_command(command) {
-                return handler.handle(context, self, sub_matches);
+                let (quit, output) = handler.handle(context, sub_matches)?;
+                if let Some(output) = output {
+                    self.writeln(&output);
+                }
+                return Ok(quit);
             }
             // Unexpected command: print an error
             return Err(format!("Unknown command: {command}"));
@@ -97,11 +100,10 @@ impl DebuggerCommand for PopulationCommand {
     fn handle(
         &self,
         context: &mut Context,
-        cli: &DebuggerRepl,
         _matches: &ArgMatches,
-    ) -> Result<bool, String> {
-        cli.writeln(&format!("{}", context.get_current_population()));
-        Ok(false)
+    ) -> Result<(bool, Option<String>), String> {
+        let output = format!("{}", context.get_current_population());
+        Ok((false, Some(output)))
     }
 }
 
@@ -122,12 +124,11 @@ impl DebuggerCommand for NextCommand {
     fn handle(
         &self,
         context: &mut Context,
-        _cli: &DebuggerRepl,
         matches: &ArgMatches,
-    ) -> Result<bool, String> {
+    ) -> Result<(bool, Option<String>), String> {
         let t = *matches.get_one::<f64>("t").unwrap();
         context.schedule_debugger(t);
-        Ok(true)
+        Ok((true, None))
     }
 }
 
@@ -140,10 +141,9 @@ impl DebuggerCommand for ContinueCommand {
     fn handle(
         &self,
         _context: &mut Context,
-        _cli: &DebuggerRepl,
         _matches: &ArgMatches,
-    ) -> Result<bool, String> {
-        Ok(true)
+    ) -> Result<(bool, Option<String>), String> {
+        Ok((true, None))
     }
 }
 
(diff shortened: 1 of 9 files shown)


src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    fn register_command(&mut self, name: &'static str, handler: Box<dyn DebuggerCommand>) {
+        self.commands.insert(name, handler);
+    }
+
+    fn get_command(&self, name: &str) -> Option<&dyn DebuggerCommand> {
+        self.commands.get(name).map(|command| &**command)
+    }
+
+    fn writeln(&self, formatted_string: &str) -> Result<(), String> {
```
ekr-cfa, line 45: I don't think this function should fail even in principle.. There's nothing reasonable for the command to do if it does.

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+    fn about(&self) -> &'static str {
+        "Continue the simulation and exit the debugger"
+    }
+    fn handle(
+        &self,
+        context: &mut Context,
+        cli: &DebuggerRepl,
+        _matches: &ArgMatches,
+    ) -> Result<bool, String> {
+        cli.writeln(&format!(
```
ekr-cfa, line 99: I probably wouldn't write anything here.

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+            "Continuing the simulation from t = {}",
+            context.get_current_time()
+        ))?;
+        Ok(true)
+    }
+}
+
+fn readline() -> Result<String, String> {
```
ekr-cfa, line 107: I think this is fine for now, but in the future I would use `rustylines`

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+    if let Some((command, sub_matches)) = matches.subcommand() {
+        // If the provided command is known, run its handler
+        if let Some(handler) = repl.get_command(command) {
+            return handler.handle(context, &repl, sub_matches);
+        }
+        // Unexpected command: print an error
+        return Err(format!("Unknown command: {command}"));
+    }
+
+    unreachable!("subcommand required");
```
ekr-cfa, line 137: This can never happen, right? Because you register commands above.

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+        cli.writeln(&format!(
+            "Continuing the simulation from t = {}",
+            context.get_current_time()
+        ))?;
+        Ok(true)
+    }
+}
+
+fn readline() -> Result<String, String> {
+    write!(std::io::stdout(), "$ ").map_err(|e| e.to_string())?;
```
ekr-cfa, line 108: I would put `t` in the prompt.

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+
+/// Starts the debugger and pauses execution
+fn start_debugger(context: &mut Context) -> Result<(), IxaError> {
+    println!("Debugging simulation at t = {}", context.get_current_time());
+    loop {
+        let line = readline()?;
+        let line = line.trim();
+        if line.is_empty() {
+            continue;
+        }
```
ekr-cfa, line 148: If `readline()` fails, is what you want to really go back to the simulation?

src/debugger.rs
```diff
@@ -0,0 +1,218 @@
...
+            .unwrap()
+            .args(["--debugger", "1.0"])
+            .write_stdin("population\n")
+            .write_stdin("continue\n")
+            .assert();
+
+        assert
+            .success()
+            // This doesn't seem to work for some reason
+            .stdout(contains("The number of people is 3"));
```
ekr-cfa, line 216: It should just say `3` right?

ekr-cfa (changes requested):
This seems mostly right, but I think could use one more spin.

src/debugger.rs
k88hudson-cfa: Well this is like when you press enter and you didn't type anything right?

ekr-cfa: Sorry, look at line 144, which actually returns with an error

src/debugger.rs
k88hudson-cfa: correct this is just to handle the unhandled else case which should in practice never be hit

src/debugger.rs
k88hudson-cfa: https://github.com/CDCgov/ixa/issues/126

src/plan.rs
```diff
@@ -97,6 +97,12 @@ impl<T, P: Eq + PartialEq + Ord> Queue<T, P> {
             }
         }
     }
+
+    /// Returns the number of remaining plans
+    #[must_use]
+    pub fn remaining_plan_count(&self) -> usize {
```
ekr-cfa, line 103: Does this have to be pub? You just use it in a test, right?

src/debugger.rs
```diff
@@ -183,36 +218,91 @@ impl ContextDebugExt for Context {
...
-    use assert_cmd::Command;
+    use crate::{Context, ContextPeopleExt};
     use predicates::str::contains;
+    use std::{cell::RefCell, io::Write, rc::Rc};
+
+    use super::build_repl;
+
+    #[derive(Clone)]
+    struct StdoutMock {
+        storage: Rc<RefCell<Vec<u8>>>,
```
ekr-cfa, line 229: Why does this have to be an Rc?

src/debugger.rs
```diff
@@ -0,0 +1,308 @@
...
+
+        cli
+    }
+
+    fn process_line(&self, l: &str, context: &mut Context) -> Result<bool, String> {
+        let args = shlex::split(l).ok_or("error: Invalid quoting")?;
+        let matches = self
+            .build_cli()
+            .try_get_matches_from(args)
+            .map_err(|e| e.to_string())?;
```
ekr-cfa, line 76: Do we really want to rebuild the CLI each tme

ekr-cfa (commented):
This looks pretty much right  but the tests don't pass.

src/debugger.rs
k88hudson-cfa: I did this to get around having to pass in a reference that can be stored in the cli struct and modify the vec, but also be able to check it later for tests once you release the cli's reference

src/debugger.rs
```diff
@@ -79,11 +100,37 @@ impl DebuggerCommand for PopulationCommand {
         cli: &DebuggerRepl,
         _matches: &ArgMatches,
     ) -> Result<bool, String> {
-        cli.writeln(&format!("{}", context.get_current_population()))?;
+        cli.writeln(&format!("{}", context.get_current_population()));
```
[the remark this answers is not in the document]
k88hudson-cfa: I changed `handle` to return a `(bool, Option<String>)` instead

ekr-cfa (approved):
LGTM

PR #124, opened 2024-12-13
Changed `sample_person()` to take a query

ekr-cfa:
The original `sample_person()` would query over the entire population. With this change you can provide a `query_people()` style query and it will sample over the matching subset.

For whole-population queries, we just draw a random value from the population size.

For subset queries, to avoid allocating a vector with all the matching people, this uses the efficient Reservoir Sampling algorithm published by Kim-Hung Li.

### DIFF
--- a/src/people.rs
+++ b/src/people.rs
@@ -2189,7 +2189,7 @@ mod test {
        
         let mut count_p1: usize = 0;
         let mut count_p2: usize = 0;
-        for _ in 0..100 {
+        for _ in 0..10000 {
             let p = context.sample_person(SampleRng2, (Age, 10)).unwrap();
             if p == person1 {
                 count_p1 += 1;
@@ -2199,7 +2199,9 @@ mod test {
                 panic!("Unexpected person");
             }
         }
-        assert!(count_p1 >= 30);
-        assert!(count_p2 >= 30);
+
+        // The chance of being more unbalanced than this is ~10^{-4}
+        assert!(count_p1 >= 4800);
+        assert!(count_p2 >= 4800);
     } 
 }
--- a/src/people.rs
+++ b/src/people.rs
@@ -897,10 +897,16 @@ pub trait ContextPeopleExt {
     /// Randomly sample a person from the population of people who match the query.
     ///
     /// The syntax here is the same as with [`Context::query_people()`].
-    /// 
+    ///
     /// # Errors
     /// Returns `IxaError` if population is 0.
-    fn sample_person<R: RngId + 'static, T: Query>(&self, rng_id: R, q: T) -> Result<PersonId, IxaError>  where R::RngType: Rng;
+    fn sample_person<R: RngId + 'static, T: Query>(
+        &self,
+        rng_id: R,
+        q: T,
+    ) -> Result<PersonId, IxaError>
+    where
+        R::RngType: Rng;
 }
 
 fn process_indices(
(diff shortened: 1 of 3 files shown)


src/people.rs
```diff
@@ -1204,15 +1207,53 @@ impl ContextPeopleExt for Context {
...
+
+        // Special case the empty query because we can do it in O(1).
+        if q.get_query().is_empty() {
+            let result = self.sample_range(rng_id, 0..self.get_current_population());
+            return Ok(PersonId { id: result });
+        }
+
+        T::setup(self);
+
+        // This function implements "Algorithm R" from KIM-HUNG LI
```
k88hudson-cfa, line 1231: @jasonasher would you mind taking a look at this

src/people.rs
```diff
@@ -894,14 +894,17 @@ pub trait ContextPeopleExt {
...
-    /// but in the future we might have holes where people were removed.
+    /// The syntax here is the same as with [`Context::query_people()`].
     ///
     /// # Errors
     /// Returns `IxaError` if population is 0.
-    fn sample_person<R: RngId + 'static>(&self, rng_id: R) -> Result<PersonId, IxaError>
+    fn sample_person<R: RngId + 'static, T: Query>(
+        &self,
+        rng_id: R,
+        q: T,
```
k88hudson-cfa, line 906: can we name this `query`

k88hudson-cfa (commented):
I might consider  having two functions (like `sample_person` and `sample_query`or something), but if in practice we're almost always sampling from a query then it's probably not worth it.

Looks good on my end but I'll let @jasonasher do the R on your algorithm

src/people.rs
jasonasher: I corrected the algorithm and expanded the test a little bit in PR #127

jasonasher (changes requested):
See PR #127

src/people.rs
ekr-cfa: Ugh. Missed the W* in L.4.

src/people.rs
```diff
@@ -1204,15 +1207,52 @@ impl ContextPeopleExt for Context {
...
+                        as usize
+                        + 1;
+                    w *= self.sample_range(rng_id, 0.0..1.0);
+                }
+            },
+            q.get_query(),
+        );
+
+        // This should always be set by the above algorithm.
+        Ok(selected.unwrap())
```
ekr-cfa, line 1255: This actually needs to check for None and then generate an error rather than just crashing.

src/people.rs
ekr-cfa: Sure.

I did think about having two functions, but I think that just having the empty query is less overhead, and it's efficient because I check for it internally.

ekr-cfa:
@k88hudson-cfa  @jasonasher good to land?

k88hudson-cfa (approved):
LGTM assuming @jasonasher agrees

src/people.rs
```diff
@@ -1204,15 +1207,54 @@ impl ContextPeopleExt for Context {
...
+                    i += (f64::ln(self.sample_range(rng_id, 0.0..1.0)) / f64::ln(1.0 - w)).floor()
+                        as usize
+                        + 1;
+                    w *= self.sample_range(rng_id, 0.0..1.0);
+                }
+            },
+            query.get_query(),
+        );
+
+        match selected {
```
jasonasher, line 1254: FWIW I think you can use `ok_or`

jasonasher (approved):
LGTM.

PR #129, opened 2024-12-18
Updated runner example

k88hudson-cfa:
Based on some feedback from Guido, this is a slightly more complete example


confunguido (approved):
Yes! This is very helpful.

PR #130, opened 2024-12-20
Add additional options to runner

k88hudson-cfa:
I cleaned up some of the `Arg` definitions also (like using `Option` property and `PathBuf` directly)


ekr-cfa (approved):
LGTM

PR #133, opened 2024-12-21
Initial rustyline integration. Fixes #126

### LINKED ISSUE #126: Debugger should use rustylines
As a follow-up to https://github.com/CDCgov/ixa/pull/123, the debugger should use `rustylines`

ekr-cfa:
This adds support for history editing via rustyline.

Before we land this, we should address the behavior under the following conditions:

* Ctrl-D -- usually end of data
* Ctrl-C -- usually interrupt

In the current code prior to this PR, C-c causes termination and C-d causes the prompt to reprint. By contrast, here is LLDB

```
20:52:35|~/dev/cfa/ixa (ekr_issue126_rustylines)$ lldb
(lldb) ^D
20:52:41|~/dev/cfa/ixa (ekr_issue126_rustylines)$ lldb
(lldb) ^C
(lldb) ^C
(lldb)
```

In

### DIFF
--- a/src/debugger.rs
+++ b/src/debugger.rs
@@ -170,8 +170,8 @@ fn start_debugger(context: &mut Context) -> Result<(), IxaError> {
     loop {
         let line = match rl.readline(&format!("t={t} $ ")) {
             Ok(line) => line,
-            Err(rustyline::error::ReadlineError::WindowResized)
-            | Err(rustyline::error::ReadlineError::Interrupted) => continue,
+            Err(rustyline::error::ReadlineError::WindowResized
+                | rustyline::error::ReadlineError::Interrupted) => continue,
             Err(rustyline::error::ReadlineError::Eof) => return Ok(()),
             Err(err) => return Err(IxaError::IxaError(format!("Read error: {err}"))),
         };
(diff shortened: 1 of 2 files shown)


ekr-cfa:
I resolved this as follows:

* ^C acts like an empty line
* ^D acts like `continue`.

@k88hudson-cfa I think this is ready for review.

src/debugger.rs
```diff
@@ -160,24 +161,24 @@ fn build_repl<W: Write + 'static>(output: W) -> DebuggerRepl {
...
+        let line = match rl.readline(&format!("t={t} $ ")) {
+            Ok(line) => line,
+            Err(
+                rustyline::error::ReadlineError::WindowResized
+                | rustyline::error::ReadlineError::Interrupted,
+            ) => continue,
+            Err(rustyline::error::ReadlineError::Eof) => return Ok(()),
+            Err(err) => return Err(IxaError::IxaError(format!("Read error: {err}"))),
+        };
+        rl.add_history_entry(line.clone())
```
k88hudson-cfa, line 180: I think it would be great to maintain history between sessions for the same simulation, but we will need some state in place to make that happen – happy to have that as a followup

k88hudson-cfa (approved):

PR #134, opened 2024-12-23
Network example

bbbruce:
An example using `network.rs`


examples/network-hhmodel/loader.rs
```diff
@@ -0,0 +1,85 @@
...
+use csv::Reader;
+use ixa::context::Context;
+use ixa::define_person_property;
+use ixa::{ContextPeopleExt, PersonId};
+use serde::Deserialize;
+use std::fs::File;
+
+#[derive(Deserialize, Copy, Clone, PartialEq, Eq, Debug, Hash)]
+pub enum AgeGroupValue {
+    U5,
```
k88hudson-cfa, line 12: Personally I'd prefer `Under5` `Under18`, but is this a common convention?

examples/network-hhmodel/loader.rs
```diff
@@ -0,0 +1,85 @@
...
+
+pub fn open_csv(file_name: &str) -> Reader<File> {
+    let current_dir = Path::new(file!()).parent().unwrap();
+    let file_path = current_dir.join(file_name);
+    csv::Reader::from_path(file_path).unwrap()
+}
+
+pub fn init(context: &mut Context) -> Vec<PersonId> {
+    // Load csv and deserialize records
+    let mut reader = open_csv("synthetic_households_us.csv");
```
k88hudson-cfa, line 56: As discussed can we reduce the size of the example populations for the purposes of landing this? Alternatively, we could put this all into a separate repo, but unless we're planning on actively using/maintaining it i think it's probably better here.

k88hudson-cfa (changes requested):

examples/network-hhmodel/loader.rs
```diff
@@ -0,0 +1,85 @@
...
+    id: u16,
+    age_group: AgeGroupValue,
+    sex: SexValue,
+    household_id: u16,
+}
+
+define_person_property!(Id, u16);
+define_person_property!(AgeGroup, AgeGroupValue);
+define_person_property!(Sex, SexValue);
+define_person_property!(HouseholdId, u16);
```
ekr-cfa, line 35: I would consider putting each of these right after the respective structs, but that's a taste issue.

examples/network-hhmodel/parameters.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub incubation_period: f64,
+    pub infectious_period: f64,
+    pub sar: f64,
+    pub shape: f64,
+    pub infection_duration: f64,
+    pub between_hh_transmission_reduction: f64,
+}
+define_global_property!(Parameters, ParametersValues);
+
+pub fn init(context: &mut Context, file_name: &str) -> Result<(), IxaError> {
```
ekr-cfa, line 22: You don't actually need to do this. If you set the property names correctly in the JSON, you can just do `context.load_global_properties()`

examples/network-hhmodel/network.rs
```diff
@@ -0,0 +1,102 @@
...
+            while let Some(person) = members.pop() {
+                for other_person in &members {
+                    context
+                        .add_edge_bidi::<HH>(person, *other_person, 1.0, ())
+                        .unwrap();
+                }
+            }
+        }
+    }
+}
```
ekr-cfa, line 37: You can use the return value of `insert` to avoid having to use `contains`

```suggestion
    for person_id in people {
        let household_id = context.get_person_property(*person_id, HouseholdId);
        if  households.insert(household_id) {
            let mut members = context.query_people((HouseholdId, household_id));
            // create a dense network
            while let Some(person) = members.pop() {
                for other_person in &members {
                    context
                        .add_edge_bidi::<HH>(person, *other_person, 1.0, ())
                        .unwrap();
                }
            }
        }
    }
}
```

ekr-cfa:
One more thought about efficiency.

Your current model regularly queries for all infected people and then tries to expose everyone in their network. And it also does that query three times, once for each edge type. But of course most people aren't infected and yet we still have to iterate over them. I would do one of the following:

1. Explicitly add an index for `DiseaseStatus` that collecting all the infected people is cheap.
2. When someone is infected, schedule a plan to expose their network at `t+1` and keep doing that as long as they are infected.

If we just want to model something with fixed generation intervals then (1) is the easiest change. However, if we want in the future to support random generation intervals, then (2) will be easier to extend to that.

examples/network-hhmodel/loader.rs
bbbruce: It was a convention used in the model I based this on. I agree with you that the other is better and it is easy to refactor as I make the other changes. Honestly it always bothered me a bit in the other that the under 18 are really the 5-18 year olds; so will fix that too.

examples/network-hhmodel/loader.rs
bbbruce: I have reduced the number of households from 5000 to 500 which still gives a decent network but makes better for an example I hope.

examples/network-hhmodel/parameters.rs
bbbruce: OK - simplified - would you recommend doing this differently than a separate module though?

ekr-cfa: I actually meant this: https://github.com/CDCgov/ixa/pull/144

examples/network-hhmodel/loader.rs
```diff
@@ -0,0 +1,86 @@
...
+    use ixa::{context::Context, random::ContextRandomExt};
+
+    const EXPECTED_ROWS: usize = 1606;
+
+    #[test]
+    fn test_init_expected_rows() {
+        let mut context = Context::new();
+        context.init_random(42);
+        init(&mut context);
+        assert_eq!(context.get_current_population(), EXPECTED_ROWS);
```
ekr-cfa, line 84: I would probably add tests for a couple of the people to see that they load properly.

examples/network-hhmodel/main.rs
```diff
@@ -0,0 +1,33 @@
...
+    // Load parameters from json
+    let file_path = Path::new(file!()).parent().unwrap().join("config.json");
+    context.load_global_properties(&file_path).unwrap();
+
+    // Load network
+    network::init(&mut context, &people);
+
+    let mut to_infect: Vec<PersonId> = Vec::new();
+    to_infect.push(context.sample_person(MainRng, ()).unwrap());
+    #[allow(clippy::vec_init_then_push)]
```
ekr-cfa, line 29:
```suggestion
    let to_infect = vec![context.sample_person(MainRng, ()).unwrap()];
```

examples/network-hhmodel/seir.rs
```diff
@@ -0,0 +1,191 @@
...
+    schedule_waiting_event(
+        context,
+        person_id,
+        parameters.shape,
+        parameters.infectious_period,
+        DiseaseStatusValue::R,
+    );
+}
+
+pub fn init(context: &mut Context, initial_infections: Vec<PersonId>) {
```
ekr-cfa, line 103:
```suggestion
pub fn init(context: &mut Context, initial_infections: &Vec<PersonId>) {
```

examples/network-hhmodel/network.rs
```diff
@@ -0,0 +1,100 @@
...
+
+fn load_edge_list<T: EdgeType + 'static>(context: &mut Context, file_name: &str, value: T::Value) {
+    let mut reader = open_csv(file_name);
+
+    for result in reader.deserialize() {
+        let record: EdgeRecord = result.expect("Failed to parse edge");
+        let p1: PersonId = context.query_people((Id, record.v1)).pop().unwrap();
+        let p2: PersonId = context.query_people((Id, record.v2)).pop().unwrap();
```
ekr-cfa, line 43: I would assert that these are each of cardinality 1.

examples/network-hhmodel/u18edges.csv
ekr-cfa, line 1: Is this used?

examples/network-hhmodel/synthetic_households_us.csv
ekr-cfa, line 1: Is this used?

examples/network-hhmodel/u5edges.csv
ekr-cfa, line 1: Is this used?

examples/network-hhmodel/parameters.rs
```diff
@@ -0,0 +1,16 @@
+use std::fmt::Debug;
+
+use ixa::define_global_property;
+use serde::{Deserialize, Serialize};
+
+#[allow(clippy::module_name_repetitions)]
```
ekr-cfa, line 6: Not needed.

examples/network-hhmodel/seir.rs
```diff
@@ -0,0 +1,191 @@
...
+        move |context, event: PersonPropertyChangeEvent<DiseaseStatus>| {
+            match event.current {
+                DiseaseStatusValue::E => schedule_infection(context, event.person_id),
+                DiseaseStatusValue::I => schedule_recovery(context, event.person_id),
+                _ => (),
+            };
+        },
+    );
+
+    // expose the first person to the disease
```
ekr-cfa, line 147:
```suggestion
    // expose the first people to the disease
```

It's true we only start with one person, but that's not part of the API guarantee of this function.

examples/network-hhmodel/seir.rs
```diff
@@ -0,0 +1,191 @@
...
+            .set_global_property_value(Parameters, parameters)
+            .unwrap();
+
+        network::init(&mut context, &people);
+
+        let to_infect = context.query_people((Id, 71));
+
+        init(&mut context, to_infect);
+
+        context.execute();
```
ekr-cfa, line 189: Shouldn't this check that something happens?

examples/network-hhmodel/synthetic_households_us.csv
bbbruce: no - not sure why these legacy CSVV stuck around - will delete

examples/network-hhmodel/loader.rs
bbbruce: is this what you had in mind?

examples/network-hhmodel/parameters.rs
bbbruce: But it complains when I don't have it?

error: item name starts with its containing module's name
 --> examples/network-hhmodel/parameters.rs:7:12
  |
7 | pub struct ParametersValues {
  |            ^^^^^^^^^^^^^^^^
  |
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_name_repetitions
  = note: `-D clippy::module-name-repetitions` implied by `-D warnings`
  = help: to override `-D warnings` add `#[allow(clippy::module_name_repetitions)]`

examples/network-hhmodel/seir.rs
bbbruce: Ensured final state is as expected

examples/network-hhmodel/loader.rs
ekr-cfa: Yes.

examples/network-hhmodel/parameters.rs
ekr-cfa: Try after rebasing because @k88hudson-cfa  added this suppression to Cargo.toml.

OK to leave it though, if you have to.

examples/network-hhmodel/seir.rs
ekr-cfa: LGTM

ekr-cfa (approved):
LGTM

k88hudson-cfa (approved):

ekr-cfa:
I landed this for you.

PR #137, opened 2024-12-23
Debugger should use an internal debugger API for setting breakpoints

k88hudson-cfa:
Rather than just adding a plan, I updated the debugger to be triggered separately from other plans; this makes it possible to actually step through each callback with`next`.

To support setting a breakpoint at `t`, I use `ExecutionPhase:Last`, which triggers on the next loop. I could alternatively create a new phase (for debugging) or a separate time queue altogether... curious if others have thoughts

### DIFF
--- a/examples/cli-debugger/main.rs
+++ b/examples/cli-debugger/main.rs
@@ -1,0 +1,46 @@
+use ixa::runner::run_with_args;
+use ixa::{
+    define_person_property, define_person_property_with_default, ContextPeopleExt,
+    PersonPropertyChangeEvent,
+};
+
+define_person_property_with_default!(IsRunner, bool, false);
+
+// Using run_with_args / run_with_custom_args will parse command line options, including
+// the --debugger interface.
+// Try:
+// cargo run --example cli-debugger -- --debugger      Starts the debugger before
+//                                                     the simulation starts executing.
+// cargo run --example cli-debugger -- --debugger 2.0  Pauses the simulation after all plans
+//                                                     have executed for t=2.0
+//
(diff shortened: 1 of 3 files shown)


ekr-cfa:
Mechanically this seems reasonable, but I wonder if this is quite the correct ergonomics around time-based breakpoints.

I would make two observations:

First, the debugger actually should take priority so when you ask for `Last` it should be *really really last*, not just "in the same equivalence class with other last stuff".

Second, it seems like we probably want a broader range of gestures, and we should probably map them out now. I can think of:

- First
- Last
- Step through to after the next thing at this timepoint (either time-based or event-based) has completed
- Step through to after this timepoint is done (i.e., break on last).

Are there others?
-Ekr

k88hudson-cfa:
> First, the debugger actually should take priority so when you ask for `Last` it should be _really really last_, not just "in the same equivalence class with other last stuff".

What do you think about introducing the concept of `ExecutionPhase:DebugLast` or something to that effect (i.e., continuing to use the plan queue with it's on phase, v.s. a separate queue that is handled separately?) I personally lean towards the former since it avoids building duplicated infrastructure but I'd be happy to hear other arguments


> First
> Last
> Step through to after the next thing at this timepoint (either time-based or event-based) has completed
 > Step through to after this timepoint is done (i.e., break on last).

That sounds good. Maybe like:
```
break --next                  Breaks after the next callback runs
break --next-t --end          Advances to the next t and breaks at the end of it
break --next-t --start        Breaks right before the next t

break -t 2.0                  Breaks at t=2.0 (end by default)
break -t 2.0 --start          Breaks right before t=2.0

```

PR #140, opened 2024-12-29
Some changes to make the debugger more flexible

ekr-cfa:
- Add persistent state (mostly for rustyline support)
- Have `process_command()` (formerly `process_line()`) return a string rather than try to write to `T: Write`.
- Remove the mocks that were needed for testing because of the previous item

### DIFF
--- a/src/debugger.rs
+++ b/src/debugger.rs
@@ -43,7 +43,7 @@ impl Debugger {
         let args = shlex::split(l).ok_or("Error splitting lines")?;
         let matches = self
             .cli
-            .clone()
+            .clone() // cli can only be used once.
             .try_get_matches_from(args)
             .map_err(|e| e.to_string())?;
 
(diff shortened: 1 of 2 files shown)


src/debugger.rs
```diff
@@ -21,70 +22,35 @@ trait DebuggerCommand {
...
+    fn process_command(
+        &self,
+        l: &str,
+        context: &mut Context,
+    ) -> Result<(bool, Option<String>), String> {
         let args = shlex::split(l).ok_or("Error splitting lines")?;
         let matches = self
-            .build_cli()
+            .cli
+            .clone()
```
k88hudson-cfa, line 45: Might be worth commenting this

k88hudson-cfa (changes requested):
For some reason the `population` command no longer seems to be working

k88hudson-cfa (approved):
LGTM

PR #146, opened 2025-01-10
Updated basic-infection, births-deaths to use `ixa::people`. Fixes #141.

### LINKED ISSUE #141: Update older examples to reflect additions to the API
The examples in `examples/` have apparently not always been kept up to date with the current state of the API. For instance, `examples/basic-infection` was written prior to the addition of the infrastructure provided by the `people` module. These examples should be brought into line with what is currently the best way to use Ixa. 

Part of this issue is defining the scope of this task. (Is it just `examples/basic-infection`, or do most or all examples need updates? Do they require accompanying modifications to documentation and/or tutorial text? Etc.) This issue will thus evolve as the scope is better understood.

- [ ] Identify all specific cases of out-of-date code
- [ ] Any uses of `PersonId` as a numeric type outside of Ixa internals should be removed/replaced.
- [ ] `basic-infection` should use the `people` module
  - [x] Code
  - [ ] README.md, which quotes code extensively

RobertJacobsonCDC:
Fixed broken tests on basic-infection and births-deaths. Moved `InfectionStatusEvent` to where it's used, in `infection_manager`. Added an explicit directory for reports-multi-threaded. Other minor cosmetic changes (spelling, grammar, etc.).


.gitignore
```diff
@@ -1,3 +1,6 @@
+.vscode
+.idea
+
```
ekr-cfa, line 3: Generally it would be better to do this in a different PR but OK in this case.

examples/basic-infection/incidence_report.rs
```diff
@@ -1,27 +1,27 @@
...
+use ixa::report::Report;
 use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
 
 #[derive(Serialize, Deserialize, Clone)]
 struct IncidenceReportItem {
     time: f64,
-    person_id: usize,
-    infection_status: InfectionStatus,
+    person_id: String,
```
ekr-cfa, line 14: Do you want to hold this until you have updated PersonId.

examples/basic-infection/infection_manager.rs
```diff
@@ -1,28 +1,33 @@
...
+            person_id,
+            InfectionStatus,
+            InfectionStatusValue::R,
+        );
     });
 }
 
 fn handle_infection_status_change(context: &mut Context, event: InfectionStatusEvent) {
-    if matches!(event.updated_status, InfectionStatus::I) {
+    if InfectionStatusValue::I == event.current {
```
ekr-cfa, line 30: This is some real C programmer energy. Let's just put the variable on the left here and below because you don't need to worry about mistyping = for == in Rust.

examples/basic-infection/infection_manager.rs
```diff
@@ -62,9 +67,13 @@ mod test {
         });
 
         let population_size = 10;
-        for id in 0..population_size {
-            context.create_person();
-            context.set_person_status(id, InfectionStatus::I);
+        for _ in 0..population_size {
+            let person_id = context.add_person(()).unwrap();
+            context.set_person_property::<InfectionStatus>(
```
ekr-cfa, line 72: I know you don't need this, but maybe create them with an explicit status of susceptible to enable local reasoning?

examples/basic-infection/transmission_manager.rs
```diff
@@ -12,55 +11,57 @@ use crate::MAX_TIME;
 define_rng!(TransmissionRng);
 
 fn attempt_infection(context: &mut Context) {
-    let population_size: usize = context.get_population();
-    let person_to_infect: usize = context.sample_range(TransmissionRng, 0..population_size);
+    let population_size: usize = context.get_current_population();
+    let person_to_infect: PersonId = context.sample_person(TransmissionRng, ()).unwrap(); //.sample_range(TransmissionRng, 0..population_size);
```
ekr-cfa, line 15: Do we still need this comment?

ekr-cfa (changes requested):

examples/basic-infection/incidence_report.rs
RobertJacobsonCDC: No, that would leave it in a nonworking state on the main branch. When I update `PersonId` I will remember to return to this because it is partly what motivated me to bring it up. I'll use it as a test case.

examples/basic-infection/infection_manager.rs
RobertJacobsonCDC: Of course you would notice that. Psychologically, after awhile these coping mechanisms become the "right" way to do things in your head, and it feels weird doing it any other way.

examples/basic-infection/transmission_manager.rs
RobertJacobsonCDC: No, it's accidental litter.

examples/basic-infection/incidence_report.rs
```diff
@@ -1,37 +1,40 @@
...
 pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    // In the configuration of report options below, we set `overwrite(true)`, which is not
+    // recommended for production code in order to prevent accidental data loss. It is set
+    // here so that newcomers won't have to deal with a confusing error while running
+    // examples.
     context
         .report_options()
-        .directory(PathBuf::from("./examples/basic-infection/"));
+        .directory(PathBuf::from("./examples/basic-infection/"))
+        .overwrite(true);
```
RobertJacobsonCDC, line 36: @ekr-cfa This and the comment right above explaining it. Both Ryan and I were confused by the error when we were trying to run the examples for the first time. Many new users may be less prepared to debug an issue in a Rust project they haven't seen before than I am. This would be the first place in the code that someone would encounter the issue if they are running the examples in order.

At other locations in the examples I only have a brief comment pointing to the explanation in this first example.

k88hudson-cfa: Yeah I think this is the right call (the comment is clear)

k88hudson-cfa (approved):
LGTM

PR #148, opened 2025-01-10
transmission tree output

paigemiller:
Add an incidence report with who infected who in the simulation rather than just print statements to the command line. Additionally uses new logging functionality rather than print statements.


bbbruce:
@paigemiller - we've landed bbb_hhmodel to main - can you rebase your PR to main and then let's take a look.  We are keeping the examples small - so I think an open question we should discuss on a call soon is if this is important to add to the example (to demostrate the network features) or if we want as they are for isolation guidance - to build upon this in a separate repository for a fully implemented hhmodel with the full synthetic population

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,54 @@
...
+
+pub fn init(context: &mut Context) -> Result<(), IxaError> {
+    let _parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+    context
+        .report_options()
+        .directory(PathBuf::from("./examples/network-hhmodel/output"))
+        .overwrite(true);
```
bbbruce, line 48: This panics if the directory does not exist.

A style suggestion only: think there is no need to use `.` in the relative path.

```suggestion
    fs::create_dir_all("examples/network-hhmodel")?;
    context
        .report_options()
        .directory(PathBuf::from("examples/network-hhmodel/output"))
        .overwrite(true);
```

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,54 @@
+use ixa::context::Context;
+use ixa::error::IxaError;
+use ixa::global_properties::ContextGlobalPropertiesExt;
+use ixa::people::PersonPropertyChangeEvent;
+use ixa::report::ContextReportExt;
+use ixa::ContextPeopleExt;
+use ixa::{create_report_trait, report::Report};
+use std::path::PathBuf;
```
bbbruce, line 8: For suggestion below re: panic on no directory.

```suggestion
use std::{fs, path::PathBuf};
```

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,54 @@
...
+    context
+        .report_options()
+        .directory(PathBuf::from("./examples/network-hhmodel/output"))
+        .overwrite(true);
+    context.add_report::<IncidenceReportItem>("incidence.csv")?;
+    context.subscribe_to_event(|context, event: PersonPropertyChangeEvent<DiseaseStatus>| {
+        handle_infection_status_change(context, event);
+    });
+    Ok(())
+}
```
bbbruce, line 54: Would suggest adding tests - e.g., that the directory and file exist after report started ... check the report has something you expect. Note in my test in `seir.rs` how I infect a specific person and set beta to 1 so that I know everyone will be infected and thus have some invariants to test against.

examples/network-hhmodel/main.rs
```diff
@@ -23,8 +26,13 @@ fn main() {
 
     // Load network
     network::init(&mut context, &people);
+    incidence_report::init(&mut context).unwrap();
 
-    let to_infect = vec![context.sample_person(MainRng, ()).unwrap()];
+    let to_infect: Vec<PersonId> = vec![context
+        .sample_person(MainRng, (AgeGroup, AgeGroupValue::Age18to64))
+        .unwrap()];
+    context.set_person_property(to_infect[0], InfectedBy, Some(to_infect[0]));
```
bbbruce, line 34: Style, but I'd put this in `init.rs`... and I'd change to handle the case where you have multiple people you are infected.  And I'm unclear - if you are using an `Option` why would you set it to a `Some` vs. `None` - the latter seems right to me - or just have a convention and get rid of `Option`.  However, I'm concerned that you would later create a panic when you need to `unwrap` that value to put in the report (I'll find if that's the case and there you could use [`unwrap_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or)).  So, I think you'll have to handle that on printing and decide what you want to print in the report for `None`.

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,54 @@
...
+    context: &mut Context,
+    event: PersonPropertyChangeEvent<DiseaseStatus>,
+) {
+    context.send_report(IncidenceReportItem {
+        time: context.get_current_time(),
+        person_id: event.person_id.to_string(),
+        infection_status: event.current,
+        infected_by: context
+            .get_person_property(event.person_id, InfectedBy)
+            .unwrap()
```
bbbruce, line 35: Yes, this would cause a panic if it is `None` so as I commented in `main.rs` you can use `unwrap_or` so you can provide a default for when the person is one of the index cases in the population.  For example,


```suggestion
            .unwrap_or("NA")
```

could be a possible way to handle.

examples/network-hhmodel/main.rs
```diff
@@ -23,8 +26,13 @@ fn main() {
...
+    #[allow(clippy::vec_init_then_push)]
     seir::init(&mut context, &to_infect);
```
bbbruce, line 37: This did not get pulled from the changes in `main.rs` right during the rebase.

```suggestion
    seir::init(&mut context, &to_infect);
```

examples/network-hhmodel/seir.rs
bbbruce, line 1: Several changes were not brought across from `main` branch to this file.  

For example, the interface to `fn init` has change to take a reference (see the change I also made in your `main.rs`) and this within `fn init`
```
    // expose the first people to the disease
    for ii in initial_infections {
        context.set_person_property(*ii, DiseaseStatus, DiseaseStatusValue::E);
    }
```
as well as new tests.  May be good at some point to show you more about using the merge editor to be sure you feel comfortable going through changes line by line and taking form one version vs. another to mix changes if you don't yet feel comfortable with that during a merge. 

Look at `git diff main bbb_hhmodel_pm -- examples/network-hhmodel/seir.rs`

So you can see what you deleted from `main.rs` that needs to be there now.

bbbruce (changes requested):
Please see requested changes.

examples/network-hhmodel/main.rs
paigemiller: We don't have an `init.rs` in this example right now - do you mean, in `loader.rs` or in `seir.rs` rather than main?

examples/network-hhmodel/incidence_report.rs
paigemiller: We do provide a default for the index person. 

And .unwrap_or() is not compiling for me here. I can look into other options, but if the infected_by person property was empty and being added to the report, we may want it to panic? The point is that each infected person gets a infected_by entry. Will defer to you all!

examples/network-hhmodel/main.rs
k88hudson-cfa: I think actually this *should* be a `Option<PersonId>`, since this is a property that only gets assigned to some people (i.e., when they get infected), so `Some(to_infect[0])` looks right to me

examples/network-hhmodel/incidence_report.rs
paigemiller: Ok changed 2 things to account for the issue of possibly empty values here: 
1. added a condition so that the only time something is getting added to this report is when they change from S->E (not other changes to DiseaseStatus). When they change from S->E, they get InfectedBy value
2. added an error handling for if for some reason, the infectedBy value is somehow still empty. Resolving this comment for now :)

examples/network-hhmodel/incidence_report.rs
paigemiller: Added a test to make sure the output folder gets created for the report.

ekr-cfa:
@paigemiller  is this ready for re-review?

paigemiller:
@ekr-cfa I'm trying to add a test that looks at the content of the incidence report per beau's suggestion. I can't get it to read the csv properly so the test is currently failing.

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,121 @@
...
+        return;
+    }
+
+    context.send_report(IncidenceReportItem {
+        time: context.get_current_time(),
+        person_id: event.person_id.to_string(),
+        infection_status: event.current,
+        infected_by: context
+            .get_person_property(event.person_id, InfectedBy)
+            .expect("Expected person to have infectedBy but none was found")
```
ekr-cfa, line 37:
```suggestion
            .expect("Expected person to have infectedBy but `None` was found")
```

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,121 @@
...
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+    fs::create_dir_all(parameters.output_dir.clone())?;
+    context
+        .report_options()
+        .directory(PathBuf::from(parameters.output_dir))
+        .overwrite(true);
+    context.add_report::<IncidenceReportItem>("incidence.csv")?;
```
ekr-cfa, line 52: The reason you need the `clone()` on likne 46 is because the `parameters` is a borrow of `context`. However, you can rearrange things so that you only need to clone `output_dir` by creating the `PathBuf` first, like so:

```
    let output_dir = PathBuf::from(parameters.output_dir.clone());
    fs::create_dir_all(&output_dir)?;
    context
        .report_options()
        .directory(output_dir)
        .overwrite(true);
```

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+        let path = Path::new(&parameters.output_dir);
+        assert!(path.try_exists().unwrap());
+        let output_path = path.join("incidence.csv");
+        assert!(output_path.try_exists().unwrap());
+
+        let column_names = read_csv_column_names(&output_path).unwrap();
+        assert_eq!(
+            column_names,
+            vec!["time", "person_id", "infection_status", "infected_by"]
```
ekr-cfa, line 118: This test only seems to verify the column names, not that anyone was infected.

A better test would have a separate listener that captured the transmission tree and then verify the results in the file matched.

ekr-cfa (changes requested):
This looks close but I think needs another spin, at least with a somewhat better test.

examples/network-hhmodel/main.rs
```diff
@@ -18,7 +19,7 @@ fn main() {
 }
 
 fn initialize(context: &mut Context) {
-    context.init_random(42);
+    context.init_random(1);
```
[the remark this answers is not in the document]
paigemiller: yeah fair question haha I changed it because 42 didn't result an outbreak, and to make the transmission tree output interesting, I thought we should have an outbreak

examples/network-hhmodel/seir.rs
```diff
@@ -22,6 +23,8 @@ define_rng!(SeirRng);
 
 define_person_property_with_default!(DiseaseStatus, DiseaseStatusValue, DiseaseStatusValue::S);
 
+define_person_property_with_default!(InfectedBy, Option<PersonId>, None);
```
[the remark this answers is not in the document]
paigemiller: I'm not sure I'm following all of this, but conceptually I like the idea of having infected by tied to the disease status values. Thanks for flagging this!

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,121 @@
...
+        return;
+    }
+
+    context.send_report(IncidenceReportItem {
+        time: context.get_current_time(),
+        person_id: event.person_id.to_string(),
+        infection_status: event.current,
+        infected_by: context
+            .get_person_property(event.person_id, InfectedBy)
+            .expect("Expected person to have infectedBy but `None` was found")
```
confunguido, line 37: What would happen here if it's a seeded infection?

paigemiller: Good question, we have had a little bit of discussion on this further up in this PR.  Right now we have this set so that the seeded infected person has infectedBy = self. We had to do this because the reports wouldn't let us have None as a value. Eric suggested above a fix to this but we may hold off on implementing for now: https://github.com/CDCgov/ixa/pull/148#discussion_r1951418725 

More comments: https://github.com/CDCgov/ixa/pull/148#discussion_r1943679030

ekr-cfa: Which report is the issue? Just this one? Because if so, you're already converting to string, so you could convert None to NA.

confunguido: Yeah, I think it might be misleading if self is marked as the infector. I could see that causing issues if I wanted to calculate Rt for instance.

examples/network-hhmodel/seir.rs
```diff
@@ -50,7 +52,13 @@ fn expose_network<T: EdgeType + 'static>(context: &mut Context, beta: f64) {
     for e in edges {
         if context.sample_distr(SeirRng, Bernoulli::new(beta).unwrap()) {
             context.set_person_property(e.neighbor, DiseaseStatus, DiseaseStatusValue::E);
-            println!("Person {} exposed person {}.", e.person, e.neighbor);
+            println!(
+                "Person {} exposed person {} at time {}.",
+                e.person,
+                e.neighbor,
+                context.get_current_time()
+            );
```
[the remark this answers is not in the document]
paigemiller: fixed!

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,200 @@
...
+        let headers = rdr.headers()?;
+        let column_names = headers.iter().map(|s| s.to_string()).collect();
+        Ok(column_names)
+    }
+
+    fn check_values(path: &Path) -> Vec<IncidenceReportItem> {
+        let mut infected_by: Vec<IncidenceReportItem> = Vec::new();
+
+        let mut rdr = csv::Reader::from_path(path).unwrap();
+        let headers = rdr.headers().unwrap();
```
ekr-cfa, line 93: I would check the column names here just to be sure, though I believe serde will be sad if they are wrong.

Then you can remove `read_csv_column_names`

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,200 @@
...
+            .get_person_property(event.person_id, InfectedBy)
+            .is_none()
+        {
+            infected_by_val = "NA".to_string();
+        } else {
+            infected_by_val = context
+                .get_person_property(event.person_id, InfectedBy)
+                .unwrap()
+                .to_string();
+        }
```
ekr-cfa, line 124: See above for suggestions to clean this up.

examples/network-hhmodel/incidence_report.rs
```diff
@@ -0,0 +1,200 @@
...
+            infected_by.push(record);
+        }
+
+        infected_by
+    }
+
+    fn test_infected_by(
+        context: &mut Context,
+        event: PersonPropertyChangeEvent<DiseaseStatus>,
+        infected_by_map: &Rc<RefCell<Vec<IncidenceReportItem>>>,
```
ekr-cfa, line 105: Let's call this something else, as it's not a map.

ekr-cfa (approved):
LGTM. 
I have some minor comments, but you should be able to make these and land.

ChiragKumar9 (approved):
Let's :ship: it!

dinacmistry (approved):
LGTM. I'd like for us to go back to the InfectedBy method in another PR and think about how to generalize this to cases where the environment is the source of infection, or how to explicitly record that an agent was a seeded infection in the simulation with unknown source, or at least the info is not necessary.

bbbruce (approved):
lgtm

k88hudson-cfa (approved):
I think this is good to merge now. Thanks @paigemiller for working through this

PR #166, opened 2025-01-24
Restructure debugger to use the clap derived API.

ekr-cfa:
This is groundwork for the Web API, where I want to use JSON as the interchange language, and this will allow us to use common for the debugger and the Web API once you get past the parser.

### DIFF
--- a/src/debugger.rs
+++ b/src/debugger.rs
@@ -4,7 +4,7 @@ use crate::Context;
 use crate::ContextGlobalPropertiesExt;
 use crate::ContextPeopleExt;
 use crate::IxaError;
-use clap::{ArgMatches, Command, FromArgMatches as _, Parser, Subcommand};
+use clap::{ArgMatches, Command, FromArgMatches, Parser, Subcommand};
 use rustyline;
 
 use std::collections::HashMap;
--- a/src/debugger.rs
+++ b/src/debugger.rs
@@ -4,8 +4,7 @@ use crate::Context;
 use crate::ContextGlobalPropertiesExt;
 use crate::ContextPeopleExt;
 use crate::IxaError;
-use clap::value_parser;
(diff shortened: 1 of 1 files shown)


src/debugger.rs
```diff
@@ -89,35 +89,40 @@ fn available_properties_str(context: &Context) -> String {
...
 
 struct GlobalPropertyCommand;
+
+#[derive(Subcommand, Clone, Debug)]
+enum GlobalPropertySubcommandEnum {
+    /// List all global properties
+    List,
+
+    /// Get the value of a global property
+    Get { property: String },
```
k88hudson-cfa, line 99: I think it would be good to keep the help text for arguments

src/debugger.rs
```diff
@@ -126,51 +131,50 @@ impl DebuggerCommand for GlobalPropertyCommand {
...
         }
     }
 }
 
 /// Adds a new debugger breakpoint at t
 struct NextCommand;
+#[derive(Parser, Debug)]
+enum NextSubcommand {
+    /// Continue until the given time and then pause again
+    Next { next_time: f64 },
```
k88hudson-cfa, line 143: Same here

src/debugger.rs
```diff
@@ -4,8 +4,7 @@ use crate::Context;
 use crate::ContextGlobalPropertiesExt;
 use crate::ContextPeopleExt;
 use crate::IxaError;
-use clap::value_parser;
-use clap::{Arg, ArgMatches, Command};
+use clap::{ArgMatches, Command, FromArgMatches as _, Parser, Subcommand};
```
k88hudson-cfa, line 7: Do you actually need the `as _`?

k88hudson-cfa (approved):
The derived stuff is a lot nicer LGTM

PR #167, opened 2025-01-27
Add a Web API and a generic external API

ekr-cfa:
* Refactor the debugger so it relies on a core of an external API module
* Add a Web API on top of the external API module


src/web_api.rs
```diff
@@ -0,0 +1,365 @@
...
+            EmptyArgs,
+            PopulationExtensionRetval
+        );
+
+        register_api_handler!(
+            api_data,
+            next,
+            NextCommandExtension,
+            NextExtensionArgs,
+            NextExtensionRetval
```
ekr-cfa, line 225: Can we use the associated types?

ekr-cfa:
Fixes #168. Fixes #169.

src/web_api.rs
```diff
@@ -0,0 +1,369 @@
...
+
+#[derive(Clone)]
+struct ApiEndpointServer {
+    sender: mpsc::Sender<ApiRequest>,
+}
+
+async fn process_cmd(
+    State(state): State<ApiEndpointServer>,
+    Path(path): Path<String>,
+    Json(payload): Json<serde_json::Value>,
```
k88hudson-cfa, line 59: What do you think about making the body optional for commands that don't require any arguments (like `continue` for example?) It seems weird to have send an empty json object

src/web_api.rs
```diff
@@ -0,0 +1,369 @@
...
+use axum::{http::StatusCode, routing::post, Router};
+use serde_json::json;
+use std::collections::HashMap;
+use std::thread;
+use tokio::sync::mpsc;
+use tokio::sync::oneshot;
+
+type ApiHandler = dyn Fn(&mut Context, serde_json::Value) -> Result<serde_json::Value, IxaError>;
+
+macro_rules! register_api_handler {
```
k88hudson-cfa, line 15: Can you explain why this has to be a macro? Maybe I'm missing something but it seems like this could probably be a regular function

src/web_api.rs
```diff
@@ -0,0 +1,369 @@
...
+                    response,
+                });
+            }
+        };
+
+        // Special case the functions which require exiting
+        // the loop.
+        if req.cmd == "next" {
+            // This was already type checked in the handler so .unwrap() cannot fail.
+            let next::Args::Next { next_time } = serde_json::from_value(req.arguments).unwrap();
```
k88hudson-cfa, line 157: I don't think this is true – the simulation seems to panic here when json doesn't parse correctly

k88hudson-cfa (changes requested):
This is really cool, can't wait to see it with a front end. I think the only major issue here is the simulation to panic when it receives malformed JSON (it should just respond with an error)

src/web_api.rs
```diff
@@ -0,0 +1,369 @@
...
+                let _ = req.rx.send(ApiResponse {
+                    code: StatusCode::OK,
+                    response,
+                });
+            }
+        };
+
+        // Special case the functions which require exiting
+        // the loop.
+        if req.cmd == "next" {
```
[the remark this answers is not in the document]
ekr-cfa: We could but I'm not sure it's advisable. They come in as strings, so we'd just be mapping a string to an enum for its trip across the channel from the web server to here. Can you elaborate on your thinking?

src/web_api.rs
ekr-cfa: Yep. Fixed and added a test case.

src/web_api.rs
ekr-cfa: It doesn't, though I had to do some turbofishing.

src/web_api.rs
ekr-cfa: Cleaned this up some.

src/web_api.rs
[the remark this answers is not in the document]
k88hudson-cfa: I guess I was thinking about the implicit relationship between the commands on the web side with the commands in debugger.rs, but that's a good point about mapping the enum to strings... I don't really have a great idea for how to do it

k88hudson-cfa (approved):
🚢  it

PR #174, opened 2025-01-27
Added logging infrastructure. Fixes #158.

### LINKED ISSUE #158: Logging Design Discussion
(Tracing is a separate feature. This issue only addresses tracing _messages_ that would be logged somewhere.)

# Levels vs. Verbosity

The standard logging paradigm in Rust is to have five "levels" in decreasing level of priority:

1.  [`error!`](https://docs.rs/log/latest/log/macro.error.html),
2.  [`warn!`](https://docs.rs/log/latest/log/macro.warn.html), 
3. [`info!`](https://docs.rs/log/latest/log/macro.info.html), 
4. [`debug!`](https://docs.rs/log/latest/log/macro.debug.html), and
5. [`trace!`](https://docs.rs/log/latest/log/macro.trace.html).

The [log](https://docs.rs/log/latest/log/) crate is the standard interface to a variety of supported logging backends. Typically you can set a "maximum log level" to filter log messages.

The problem with this paradigm is that these "levels" are not (I argue) linearly ordered by importance. Rather, they are types of messages that should not have an order relationship. Why should a trace message be less important than a debug message? Why is warn more important than info? Is the audience for debug messages even the same as for info messages?

Confusing these message types with verbosity levels is a category error, in my opinion. A more useful paradigm would have both different message types and a verbosity level. This allows for example:

- Set very fine grained debug messages at a high verbosity level and more coarse grained debug messages at a lower verbosity level. 
- You might emit all fatal errors at verbosity 1 while minor recoverable errors are emitted at verbosity level 3. 
- Maybe every trace event is logged to a file but only trace events of verbosity 2 and lower are output to the console. 

The following is a design I used in another project. I was only logging to the console, but obviously log messages can be routed anywhere, with different filters for different routes. I use the phrases "threshold" and "verbosity level" interchangably. 

# Overview

The `log` module provides logging capabilities with customizable thresholds and log levels. The log level describes what _kind_ of messages are to be logged, and the numeric threshold is a verbosity level, which describes the _verbosity_ of the logger.

Here is a simple example.

```rust
use ixa::log::*;

fn main() {
    // Set the global threshold to 1. Any messages logged with greater threshold
    // will not be emitted.
    set_global_logging_threshold(1);

    // Log messages with different channels and levels
    error!(2, "This is an error message.");     // Not emitted
    warning!(2, "Check your input.");           // Not emitted
    info!(1, "Processing started.");            // Emitted
    debug!(0, "Variable values are correct.");  // Emitted
    trace!(0, "Step through the logic here.");  // Emitted
}
```

## Threshold

The global logging threshold is a numerical value, with higher values meaning more verbose logging. This global value is the same for all "levels" (info, warning, error, etc.). Individual log entries are logged "at" a given threshold
and are only emitted if their level is _at most_ the global threshold level. In other words, only messages logged at a threshold less than or equal to the global threshold are emitted. A message with threshold 0 is always emitted.

You can set and get the global logging threshold as follows:

```rust
use ixa::log::{set_global_logging_threshold, get_global_logging_threshold};

// Set the verbosity to 3.
set_global_logging_threshold(3);
println!("The global logging threshold is {}", get_global_logging_threshold());
// Messages logged at levels greater than 3 will not be emitted until the verbosity is set to another value.
// ...
// (Re)set the verbosity to 5.
set_global_logging_threshold(5);
// Messages logged at any nonnegative level will now be emitted from here on.
```

## Levels

Available levels are: Error, Warning, Info, Debug, Trace. Messages of a particular level are prefixed with
the (color coded) level name.

# Macros

The following macros are provided for logging at different levels:

 - `error!`
 - `warning!`
 - `info!`
 - `debug!`
 - `trace!`

syntax:

```rust
// With threshold
level!(threshold, "format string", args...);

// Without threshold (indicates threshold of 0, always emitted)
level!("format string", args...);
```

 - `threshold`: A `u8` value representing the threshold for the log message.
 - `"format string"`: A format string, similar to `println!`.
 - `args...`: Arguments to be formatted into the format string.

Examples:

```rust
use ixa::log::{info, set_global_logging_threshold};

fn main() {
    set_global_logging_threshold(3);
    let value = 42;
    // This message will be logged because its threshold (2) <= global threshold (3)
    info!(2, "Processing value: {}", value);
    // This message will not be logged because its threshold (4) > global threshold (3)
    info!(4, "This message will not be logged.");
    // Increase the global threshold
    set_global_logging_threshold(5);
    // Now this message will be logged
    info!(4, "This message will now be logged. The meaning of life, the universe, and everything is {}.", 42);
}
```

# Feature Summary

 - **Threshold Values:** Use thresholds to categorize log messages based on importance or verbosity.
   - Lower threshold values indicate higher importance.
   - The global logging threshold controls which messages are logged.
 - **Default Threshold:** If the threshold argument is omitted in the macro, it defaults to 0.
 - **Dynamic Threshold Adjustment:** Use `set_global_logging_threshold` to change the logging threshold at runtime.
 - **Automatic Logger Initialization:** The logging macros handle logger initialization automatically; no explicit initialization is required.
 - **Thread Safety:** The global logging threshold is managed using atomic operations, ensuring thread safety.

RobertJacobsonCDC:
Added `trace!` statements to `basic-infection` and enabled logging.
Added `trace!` statements throughout Ixa proper.
Added a default set of module log level filters, as it turns out rustyline logs trace messages, which are noisy. Modified `examples/basic-infection` to accept command line arguments.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -16,8 +16,8 @@
 //! ```
 //!
 //! Logging is _disabled_ by default. Logging messages can be enabled by passing the command line
-//! option `--enable-logging`. Log messages can also be controlled programmatically. Logging can be
-//! enabled/disabled from code using the functions:
+//! option `--log-level <level>`. Log messages can also be controlled programmatically. Logging
+//! can be enabled/disabled from code using the functions:
 //!
 //!  - `enable_logging()`: turns on all log messages
 //!  - `disable_logging()`: turns off all log messages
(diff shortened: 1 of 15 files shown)


RobertJacobsonCDC:
I have a draft for log-related debugger commands on the `RobertJacobsonCDC_158_logging_debugger` branch.

RobertJacobsonCDC:
**Some notes about this implementation.**

A natural question is whether logging configuration should be stored in a `DataPlugin` and attached to a `Context` instance. A logger is global, not specific to a `Context` instance, and there doesn't seem to be an advantage to accessing logging facilities through a `Context`.  Consequently, logging configuration isn't stored in a `Context`.

Another question is how to balance flexibility and features against simplicity. This question recommends having an API that is agnostic to backend implmentation. There is a complexity cliff in the logging ecosystem: There are the simple "just works" crates that require just a few lines, and there are these byzantine crates with loads of abstractions and features. If we need more features, we can swap out the backend without having to refactor client code.

src/log.rs
```diff
@@ -0,0 +1,214 @@
...
+use std::sync::Mutex;
+
+// Logging disabled.
+const DEFAULT_LOG_LEVEL: LevelFilter = LevelFilter::Off;
+// Automatically determine if output supports color.
+const DEFAULT_LOG_STYLE: WriteStyle = WriteStyle::Auto;
+// Default module specific filters
+const DEFAULT_MODULE_FILTERS: [(&str, LevelFilter); 1] = [
+    ("rustyline", LevelFilter::Off),
+    // ("ixa", LevelFilter::Off),
```
RobertJacobsonCDC, line 56: Since `rustyline` logs noisy `trace` messages, I think we need to have `rustyline` specifically filtered out by default. But it's not clear if we should have internal `ixa` messages filtered out by default if the user enables logging. I would think not, but...

src/log.rs
```diff
@@ -0,0 +1,214 @@
...
+    {
+        let log_configuration = get_log_configuration();
+        log_configuration.module_level.remove(module_path);
+    }
+    set_logger();
+}
+
+/// Sets the level filters for a set of modules according to the provided map. Use this instead of
+/// `set_module_filter()` to set filters in bulk.
+#[allow(clippy::implicit_hasher)]
```
RobertJacobsonCDC, line 158: Clippy has a lint that wants every `HashMap` to be generic on the `HashBuilder`. (?! That seems bonkers to me. We wouldn't require everything that allocates to be generic on the allocator, would we?) There is no circumstance I can imagine in which client code would want a specialty hasher for this use case.

src/log.rs
```diff
@@ -0,0 +1,214 @@
...
+            builder.filter(Some(module), *filter);
+        }
+
+        builder.build()
+    }
+}
+
+/// Enables the logger with no global level filter / full logging. Equivalent to
+/// `set_log_level(LevelFilter::Trace)`.
+pub fn enable_logging() {
```
k88hudson-cfa, line 117: I'll add a comment in the runner section but I think I'd probably be in favor of keeping the programmatic API simple (like just `set_log_level`, not `enable_logging`/`disable_logging`) and having the syntactic sugar be at the command line

src/log.rs
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        Some(handle) => {
+            // Replace the existing logger.
+            if let Err(error) = handle.replace(logger) {
+                error!("failed to set logger: {}", error);
+            }
+        }
+    }
+}
```
k88hudson-cfa, line 214: Could you maybe add a couple of tests?

src/runner.rs
```diff
@@ -43,6 +49,7 @@ impl BaseArgs {
             output_dir: None,
             file_prefix: None,
             force_overwrite: false,
+            enable_logging: false,
```
k88hudson-cfa, line 52: This definitely a matter of taste but I think I would lean towards having `error` on by default, and then `--log-level={level}`. I could also see cargo-style  `-v` or `--verbose` (maybe for debug?) / `-vv` or `--very-verbose` (for trace) 

As a follow-up: after looking at your comments with `rustyline` etc. above, I do think we are going to want to allow module-level log settings if people need them (either through a command line option or and environment variable)

k88hudson-cfa (changes requested):
I think as a general principle conventional is good here in terms of the interface. This will be good to put in front of people and see what they think :)

src/log.rs
RobertJacobsonCDC: Do you mean have `set_module_filter` and others as a command line argument? Or do you mean leave the command line argument as it is (just `--enable-logging`) and leave out all of the programmatic API except for `set_log_level`?

src/log.rs
RobertJacobsonCDC: I knew I was forgetting something. 🤦

src/log.rs
RobertJacobsonCDC: (Clarified in chat.)

RobertJacobsonCDC:
@k88hudson-cfa Rebased and switched from `env_logger` to `log4rs`, which supports changing level filters at runtime.

src/log.rs
```diff
@@ -0,0 +1,360 @@
...
+pub fn set_module_filters<S: ToString>(module_filters: &[(&S, LevelFilter)]) {
+    let mut log_configuration = get_log_configuration();
+    log_configuration.set_module_filters(module_filters);
+}
+
+/// Fetches a mutable reference to the global `LogConfiguration`.
+fn get_log_configuration() -> MutexGuard<'static, LogConfiguration> {
+    // Silence lint about mutable global variables.
+    #[allow(static_mut_refs)]
+    unsafe {
```
k88hudson-cfa, line 261: Can you actually use a `RefCell` here instead (like something similar to what we're doing with `GLOBAL_PROPERTIES`)? I was hoping we could avoid adding `unsafe` here

k88hudson-cfa (changes requested):
API-wise I think this looks good, thanks for reworking it – I had a comment about mutable globals, othweise I think this is good to land

src/log.rs
```diff
@@ -0,0 +1,360 @@
...
+
+#[cfg(test)]
+mod tests {
+    // Silence lint about mutable global variables related to `TEST_MUTEX`.
+    #![allow(static_mut_refs)]
+    use super::{get_log_configuration, remove_module_filter, set_log_level, set_module_filters};
+    use log::{error, trace, LevelFilter};
+    use std::sync::{LazyLock, Mutex};
+
+    // Force logging tests to run serially for consistent behavior.
```
k88hudson-cfa, line 274: Would something like this be a good idea maybe? You'd probably know better than me on what's normal here https://crates.io/crates/serial_test

RobertJacobsonCDC: I think the simplest and easiest thing is the best thing. The `serial_test` crates adds another dependency and just uses a mutex under the hood anyway.

But my mutex doesn't need to be mut, so I don't need unsafe.

src/log.rs
RobertJacobsonCDC: 🤦 That's vestigial. I've removed all unsafe blocks. The globals don't need to be mut.

k88hudson-cfa (approved):
🚢 it

PR #175, opened 2025-01-28
Add a CSRF secret. Fixes #172.

### LINKED ISSUE #172: Add an anti-CSRF secret to the Web API
Right now the Web API is too impoverished to do much dangerous stuff, but we should add an anti-CSRF secret to the URL to make it future proof.

ekr-cfa:
@k88hudson-cfa you should hold off on reviewing this until you have reviewed #169, because this stacks on #169.

### DIFF
--- a/src/web_api.rs
+++ b/src/web_api.rs
@@ -190,7 +190,8 @@ impl ContextWebApiExt for Context {
         }
 
         // Start the API server
-        let mut rng = StdRng::seed_from_u64(0);
+        //        let rng = StdRng::from_os_rng();
+        let mut rng = StdRng::from_rng(rand::rngs::OsRng).unwrap();
         let mut random: [u8; 16] = [0; 16];
         rng.fill_bytes(&mut random);
         let secret = uuid::Builder::from_random_bytes(random)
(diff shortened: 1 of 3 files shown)


src/web_api.rs
```diff
@@ -184,13 +189,16 @@ impl ContextWebApiExt for Context {
...
 
         // Start the API server
-        let (ready_tx, ready_rx) = oneshot::channel::<Result<(), IxaError>>();
-        thread::spawn(move || serve(api_to_ctx_send, port, ready_tx));
-        let ready = ready_rx.blocking_recv().unwrap();
-        #[allow(clippy::question_mark)]
-        if ready.is_err() {
-            return ready;
-        }
+        let mut rng = StdRng::from_rng(rand::rngs::OsRng).unwrap();
```
k88hudson-cfa, line 192: Why do you need `StdRng` here as opposed to using `OsRng.fill_bytes` directly?

k88hudson-cfa (approved):
Looks good, I had one question about the use of `StdRng`

src/web_api.rs
ekr-cfa: Because I wrote this expecting to be able to use StdRng directly and then discovered the doc was wrong and didn't go back and clean up.

PR #176, opened 2025-01-28
Promoted `people` module to a directory

RobertJacobsonCDC:
Promoted `people` module to a directory, split into submodules in multiple files.
Moved `Tabulator` to top level module.
Relocated tests accordingly.


src/tabulator.rs
```diff
@@ -0,0 +1,163 @@
+/*!
+
+A `Tabulator` is a mechanism by which reports know their columns.
+
+*/
```
ekr-cfa, line 5: Please use //! style comments. Also, it's more than just "know their columns". It collects the reports.

src/people/mod.rs
```diff
@@ -1024,8 +996,8 @@ mod test {
         Child,
         Adult,
     }
-    define_global_property!(Threshold, u8);
-    define_derived_property!(IsEligible, bool, [Age], [Threshold], |age, threshold| {
+    define_global_property!(ThresholdP, u8);
```
ekr-cfa: Why did you rename these?

ekr-cfa (commented):
This generally looks fine, but I have one comment and one question.

src/tabulator.rs
RobertJacobsonCDC: Prepending each line with `//!` is problematic for some screen readers, so I would advocate for a block style for purposes of accessibility. But that's a separate conversation we can have outside of this PR.

src/people/mod.rs
RobertJacobsonCDC: Because of issue #179. Should I add a code comment?

src/tabulator.rs
ekr-cfa: Our policy is to use the Rust Style Guide, which is clear on this point:

https://doc.rust-lang.org/nightly/style-guide/#doc-comments

src/people/mod.rs
ekr-cfa: No I don't think that's necessary. I'm just making sure we don't have spurious changes because they're hard to detect in a big diff.

ekr-cfa (approved):
LGTM

PR #183, opened 2025-01-30
Add support for serving static files

ekr-cfa:
We'll eventually probably want to have some way to pick the directory, but I haven't worked out the details yet, and this is enough for development


k88hudson-cfa (approved):

k88hudson-cfa:
I'm assuming you intended to leave this empty for now

PR #184, opened 2025-01-30
Add an API to get properties for individual people

ekr-cfa:
Query is still TODO

### DIFF
--- a/src/external_api.rs
+++ b/src/external_api.rs
@@ -194,7 +194,7 @@ pub(crate) mod people {
                         return Err(IxaError::IxaError(format!("No person with id {id}")));
                     }
 
-                    let value = context.get_person_property_by_name(&property, PersonId(*id))?;
+                    let value = context.get_person_property_by_name(property, PersonId(*id))?;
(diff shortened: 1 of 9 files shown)


src/external_api.rs
```diff
@@ -157,3 +157,90 @@ pub(crate) mod r#continue {
...
+
+pub(crate) mod people {
+    use crate::people::{external_api::ContextPeopleExtCrate, ContextPeopleExt, PersonId};
+    use crate::Context;
+    use crate::IxaError;
+    use serde::{Deserialize, Serialize};
+
+    #[derive(Deserialize)]
+    pub(crate) enum ArgsEnum {
+        Get { id: usize, property: String },
```
k88hudson-cfa, line 169: Do you want to maybe use `person_id` since that's what we have been using everywhere else? Also, why not have this actually be `PersonId`

k88hudson-cfa (approved):
Small suggestion but I think this looks good

src/external_api.rs
ekr-cfa: Fixed.

PR #186, opened 2025-01-31
updated devcontainer files to have a working rust codespaces

cdc-as81:


ekr-cfa:
I'm not sure you need to build a new docker container here. There is an existing Rust devcontainer from Microsoft:

https://github.com/devcontainers/images/tree/main/src/rust

That's what's being references in our .devcontainer file  (https://github.com/CDCgov/ixa/blob/main/.devcontainer/devcontainer.json) (I think). For some reason you seem to need to force a rebuild to get it effective in your codespace, but once you do, Rust development works fine. Have you tried getting that working rather than building your own container?


If we do need to build a container, I don't think it should change people's shells from the default (bash) to zsh. That's something that should be done by setting your own settings, which can be made to apply to all of your codespaces. Instructions here: https://bea.stollnitz.com/blog/codespaces-terminal/

ekr-cfa (approved):
LGTM

PR #192, opened 2025-02-05
Add an external tabulate API.

ekr-cfa:
This exposes an external API to allow you to tabulate person properties. The major difference from the internal API is that it takes property names as strings. This API is plumbed up to the Web API but could also be plumbed to the debugger.

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -351,18 +351,23 @@ impl ContextPeopleExt for Context {
         F: Fn(&Context, &[String], usize),
     {
         let type_ids = tabulator.get_typelist();
+        tabulator.setup(self).unwrap();
 
-        // First, update indexes
-        {
-            let data_container = self.get_data_container(PeoplePlugin)
-                .expect("PeoplePlugin is not initialized; make sure you add a person before accessing properties");
-
-            for t in &type_ids {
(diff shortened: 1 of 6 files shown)


src/people/context_extension.rs
```diff
@@ -352,17 +352,22 @@ impl ContextPeopleExt for Context {
...
-            let data_container = self.get_data_container(PeoplePlugin)
-                .expect("PeoplePlugin is not initialized; make sure you add a person before accessing properties");
-
-            for t in &type_ids {
-                if let Some(mut index) = data_container.get_index_ref_mut(*t) {
-                    let methods = data_container.get_methods(*t);
-                    index.index_unindexed_people(self, &methods);
+        if let Some(data_container) = self.get_data_container(PeoplePlugin) {
+            for type_id in &type_ids {
+                tabulator.setup(self).unwrap();
```
k88hudson-cfa, line 357: Does this actually need to happen for every `type_id`?

k88hudson-cfa (approved):
I think this looks good mostly, had one question about the setup  in the context_extension code

src/people/context_extension.rs
ekr-cfa: Oh yeah good catch.

PR #193, opened 2025-02-05
Added query_and utility

k88hudson-cfa:
My use case here is that I'd like to support being able to construct queries across function boundaries, for example:

```rust
fn get_susceptible_people<Q>(context: &Context, q) where Q: Query {
    let base_query = ((Alive, true), (InfectionStatus, InfectionStatusValue::Susceptible));
    context.query_people(query_add(q, base_query));
}

let other_properties = ((RiskCategory, RiskCategory::High), (Region, CA));
let people = get_susceptible_people(context, extra_query); 

```

### DIFF
--- a/src/people/query.rs
+++ b/src/people/query.rs
@@ -9,12 +9,12 @@ use std::any::TypeId;
 /// we implement Query for tuples of up to size 20, that's invisible
 /// to the caller. Do not use this trait directly.
 pub trait Query {
-    fn setup(context: &Context);
+    fn setup(&self, context: &Context);
     fn get_query(&self) -> Vec<(TypeId, IndexValue)>;
 }
 
 impl Query for () {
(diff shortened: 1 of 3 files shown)


src/people/query.rs
```diff
@@ -9,12 +9,12 @@ use std::any::TypeId;
 /// we implement Query for tuples of up to size 20, that's invisible
 /// to the caller. Do not use this trait directly.
 pub trait Query {
-    fn setup(context: &Context);
+    fn setup(&self, context: &Context);
```
k88hudson-cfa, line 12: This was needed to make Query object-safe

ekr-cfa (commented):
This seems sensible but I wonder if we can clean up the implementation.

ekr-cfa (approved):
LGTM

PR #197, opened 2025-02-06
Fix argument parsing for time-varying-infection

ekr-cfa:
This should have been caught in CI. I'm going to add some CI for examples in a separate PR.


RobertJacobsonCDC (approved):
Looks good. We should also check that the benchmarks run in CI, but I can do that in a different issue.

PR #201, opened 2025-02-06
Allow programs to extend web api

ekr-cfa:


k88hudson-cfa (approved):
Awesome, that was straightforward :)

PR #203, opened 2025-02-07
Implement clone for QueryAnd

k88hudson-cfa:
This is needed for situations where you need to iterate over multiple queries given some query from a function (such as when you're looking for a next contact)


ekr-cfa (approved):
LGTM

PR #206, opened 2025-02-07
Remove spurious comma. Fixes #199

### LINKED ISSUE #199: Error when using derived property
I get the following error when trying to do `cargo run`. The issue seems to come from trying to use `define_derived_property`. To me, the error appears to say that there is an extra "," at line 126 of `src/people/property.rs`. Removing this comma fixed the issue for me.

Image

ekr-cfa:
@em-gr does this fix your issue?

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -602,6 +602,15 @@ mod tests {
         age >= threshold
     });
 
+    // This isn't used, it's just testing for a compile error.
+    define_derived_property!(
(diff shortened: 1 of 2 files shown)


k88hudson-cfa (approved):
LGTM

Em-Gr:
Hello, 
Thank you for looking into this so quickly! Unfortunately, this does not appear to solve my error. Removing the same comma as suggested in your fix I now get the following error: 
image

PR #210, opened 2025-02-10
Re-export rand, ctor, paste

### LINKED ISSUE #207: Using ixa macros requires rand, ctor, paste
It's not ideal that users have to manually include rand, ctor, and paste as dependencies – is there some way we could re-export from ixa maybe? 

Test case (this demonstrates importing ixa and then requiring extra dependencies):
https://github.com/k88hudson-cfa/ixa-test-repo

k88hudson-cfa:
Fixes #207

### DIFF
--- a/src/global_properties.rs
+++ b/src/global_properties.rs
@@ -114,8 +114,8 @@ macro_rules! define_global_property {
             }
         }
 
-        paste::paste! {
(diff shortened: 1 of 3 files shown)


RobertJacobsonCDC (approved):
Looks good to me.

PR #211, opened 2025-02-10
Debugger front end refactor

k88hudson-cfa:
- Moved the global `generation` value into a hook
- I removed creating new charts for now
- Some basic styles

I have a [branch ](https://github.com/CDCgov/ixa/compare/k88hudson-react-setup?expand=1) that adds a build system/typescript etc. but I converted it back for now to unblock people using this while I try to get it working with cargo somehow


static/src/App.js
```diff
@@ -0,0 +1,119 @@
...
+      }
+      setSettings(newSettings);
+    })();
+  }, [generation]);
+
+  return html`
+    <div className="settings">
+      <h2>Global Settings</h2>
+      <ul>
+        ${Object.entries(settings).map(
```
ekr-cfa, line 48: If you're going to do this with a dict you want to sort the keys so things don't come out in random order.

static/src/App.js
```diff
@@ -0,0 +1,119 @@
...
+        )}
+      </ul>
+    </div>
+  `;
+}
+
+function App() {
+  const { currentTime, goNext } = useAppState();
+
+  async function doApiCall() {
```
ekr-cfa, line 83: Why is this called "doApiCall()"? It's not a generic API call function

static/src/PeopleCharts.js
```diff
@@ -0,0 +1,140 @@
...
+              ${p}
+            </label>
+          `
+        )}
+      </div>
+      ${
+        selectedProperties.length
+          ? html`
+              <${PeopleTable} properties=${selectedProperties} />
+              <${PeopleGraph} properties=${selectedProperties} />
```
ekr-cfa, line 48: * I don't think we should couple the table and graph together. The table takes up a lot of space and it;s going to be worse if we have >1 property
* I don't love having the checkmarks change the current graph. If you try selecting and deselecting you will notice you lose history, which is annoying but it's also just going to be kind of weird even if you keep the history for a given configuration that when you uncheck a property the history is lost but then you check it and it comes back. Having each graph have an immutable configuration removes that issue.

static/src/useAppState.js
```diff
@@ -0,0 +1,32 @@
...
+export function SimulationContextProvider({ children }) {
+  const [generation, setGeneration] = useState(0);
+  const [currentTime, setCurrentTime] = useState(0);
+
+  useEffect(() => {
+    getTime().then((t) => setCurrentTime(t));
+  }, [generation]);
+
+  async function goNext() {
+    nextTime(currentTime + 1.0);
```
ekr-cfa, line 20: BUG: You aren't waiting on `nextTime()` and so when `setGeneration()` is called it will trigger API calls and now you have a race condition with respect to `nextTime()` starting.

BUG: `nextTime()` returns right after the server acknowledges the request, but if it takes a really long time to run the next tick, then the reload queries may time out. This is why I busy wait on the time incrementing.

ekr-cfa (changes requested):

static/src/useAppState.js
k88hudson-cfa: Ah good catch, this should have an `await`

static/src/App.js
k88hudson-cfa: Added a comment in case someone runs into this in the future

static/src/PeopleCharts.js
```diff
@@ -1,70 +1,139 @@
...
       setSelectedProperties(
         selectedProperties.filter((existing) => existing !== p)
       );
     }
-  };
+  }
+
+  function onSubmit(e) {
+    e.preventDefault();
+    if (selectedProperties.length === 0) {
```
ekr-cfa, line 57: For the future, do you think this should produce an error?

ekr-cfa (approved):
LGTM other than needing the busy wait. We discussed offline and happy to review again when you have it. Approving proactively.

static/src/PeopleCharts.js
k88hudson-cfa: Yes, I'll add a comment

ekr-cfa (approved):
LGTM2

PR #212, opened 2025-02-11
Replace `Hash` with `Serialize` for `IndexValue`s. Fixes #189

### LINKED ISSUE #189: Make it easier to define float-type person properties
After working with this awhile, I think it's pretty clear that the ergonomics on floats for person properties are not great.
Currently, because person properties have to implement `Hash`, the way we suggest handling this is to use `OrderedFloat<f64>`.

Say you try to define a person property related to a time:

```rs
define_person_property!(TimeOfInfection, f64);
```

First, the error message isn't that helpful. It doesn't tell you *why* it needs to implement hash, and it doesn't
tell you how to fix it, which is to use `OrderedFloat`
Image

Second, you have to add  `ordered-float` separately, which isn't great.
```rust
define_person_property!(TimeOfInfection, OrderedFloat<f64>);
```

Third, if you have to actually work with times, you have to remember to convert back and forth, which is kind of annoying

```rust
 let time = self.get_person_property(person_id, TimeOfInfection).0 - current_time
```

### So what do we do?

At the very least we could improve the error message and include a type with ixa so you don't have to import anything extra, but I think actually it's worth considering whether we could just let people use `f64` and complain if they try to use them for a query (generally, it's hard to think of when indexing/querying on an f64 would make sense anyway)

RobertJacobsonCDC:
This changes replaces the requirement that `PersonProperty::Value` implement `Hash` with the requirement that `PersonProperty::Value` implement `Serialize`. This allows for `f64` to be used in properties without the need to be wrapped in `OrderedFloat`.

### DIFF
--- a/src/people/property.rs
+++ b/src/people/property.rs
@@ -1,16 +1,16 @@
 use crate::people::data::PersonPropertyHolder;
 use crate::{Context, PersonId};
+use serde::Serialize;
 use std::fmt::Debug;
-use std::hash::Hash;
 
 /// An individual characteristic or state related to a person, such as age or
 /// disease status.
 ///
-/// Person properties should defined with the [`define_person_property!()`],
+/// Person properties should be defined with the [`define_person_property!()`],
 /// [`define_person_property_with_default!()`] and [`define_derived_property!()`]
 /// macros.
 pub trait PersonProperty: Copy {
-    type Value: Copy + Debug + PartialEq + Hash;
+    type Value: Copy + Debug + PartialEq + Serialize;
     #[must_use]
     fn is_derived() -> bool {
         false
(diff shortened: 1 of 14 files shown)


src/people/property.rs
```diff
@@ -1,16 +1,16 @@
...
 /// An individual characteristic or state related to a person, such as age or
 /// disease status.
 ///
-/// Person properties should defined with the [`define_person_property!()`],
+/// Person properties should be defined with the [`define_person_property!()`],
 /// [`define_person_property_with_default!()`] and [`define_derived_property!()`]
 /// macros.
 pub trait PersonProperty: Copy {
-    type Value: Copy + Debug + PartialEq + Hash;
+    type Value: Copy + Debug + PartialEq + Serialize;
```
ekr-cfa, line 13: BTW, we're eventually going to want Deserialize to allow for external_api setting of people properties, but we can do that later.

ekr-cfa (approved):
LGTM

PR #213, opened 2025-02-11
added create_presicion_report

cdc-as81:


src/report.rs
```diff
@@ -230,6 +305,7 @@ impl ContextReportExt for Context {
     fn send_report<T: Report>(&self, report: T) {
         let writer = &mut self.get_writer(report.type_id());
         report.serialize(writer);
+        writer.flush().unwrap();
```
k88hudson-cfa, line 308: We don't want to flush every time we send reports since this we could be sending a high volume of reports a lot during a simulation – instead, make sure that the writer goes out of scope on the test  (dropping `context` should work)

cdc-as81: removed

k88hudson-cfa (approved):
LGTM, thanks!

PR #214, opened 2025-02-11
Added to GitHub Workflow to run examples in CI

RobertJacobsonCDC:
Modified `build-test.yaml` to run this script in CI.


RobertJacobsonCDC:
I'm not a fan of how this script works right now, because it exits when the first command fails to run. I'd rather it run every example and return an error code if any single example fails.

RobertJacobsonCDC:
Here's an alternative approach that runs every example and exits with an error code if any single example fails.

```bash
#!/bin/bash

# Initialize a variable to track failures
has_error=0

# Function to run a command and log failures
run_command() {
    echo "Running: $1"
    $1
    if [ $? -ne 0 ]; then
        echo "Command failed: $1"
        has_error=1
    fi
}

# Run each command
run_command "cargo run --example basic-infection"
run_command "cargo run --example births-deaths"
# ... etc. ...
run_command "cargo run --example time-varying-infection ./examples/time-varying-infection/input.json"

# Exit with the appropriate code based on whether any command failed
if [ $has_error -eq 1 ]; then
    echo "One or more commands failed."
    exit 1
else
    echo "All commands executed successfully."
    exit 0
fi
```

ekr-cfa:
I agree that it's undesirable to have the whole job fail when a single example fails. In addition, I think we want to have individual failures visible in the GitHub interface. I think this pushes us towards each example being a separate job in GitHub actions using the GitHub actions "matrix" feature. 

Here's a partial example:

https://github.com/CDCgov/ixa/pull/216

This is unfinished. There are a few areas for improvement here:

1. As you can see, I'm using two jobs connected by an artifact so we only need to build once. However, for some reason (timestamps?) we're actually getting a rebuild. This needs some investigation.
1. You need some way to give time-varying-infection an argument for the config file. I assume you need some list of the arguments for each example.
1. It should be possible to auto-detect all the examples, e.g., run `cargo --example` and then use a script to feed this into the matrix. Of course that only works if they don't take arguments

At minimum we need to fix (1) and (2).  However, I think this is a more promising angle than a script.

ekr-cfa:
This is what it looks like: https://github.com/CDCgov/ixa/actions/runs/13291542297

RobertJacobsonCDC:
Thanks for this work! I'll add "WIP" to this PR and keep working on this. 

I've been getting a list of examples with this bash one-liner:

```bash
EXAMPLES=$(cargo run --example 2>&1 | tail -n +3)
```

But you're right, this only works for examples without arguments.

RobertJacobsonCDC:
Cached artifacts appear to be saved and restored correctly, but I cannot observe any difference in compile times. 

 - Other examples across the web do not cache `target/debug/.fingerprint` nor do they update timestamps, so these steps may be unnecessary in a working solution. 
 - It's cache, not cash.
 - Note that cargo says "compiling" event when it doesn't actually compile, so use `--verbose`.

RobertJacobsonCDC:
Examples are running. Caching of dependencies and their build artifacts has been bumped to the backlog as a separate task for now.

cdc-as81 (approved):
LGTM

Issue #200, opened 2025-02-06
Static directory should be relative to ixa, not current directory


k88hudson-cfa:
Currently the web server uses `"static"` which uses current directory not the root of `ixa` – I think what we want is `PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("static")`

PR #219, opened 2025-02-12
Adjusting web path to project root #200

cdc-as81:
also added pre-commit to the project codespace.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,9 +1,5 @@
 {
 	"image": "mcr.microsoft.com/devcontainers/rust:latest",
-	"features": {
-		"ghcr.io/prulloac/devcontainer-features/pre-commit:latest": {}
-	},
-	"postCreateCommand": "pre-commit install",
 	"customizations": {
 		"vscode": {
 			"settings": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,9 @@
 {
 	"image": "mcr.microsoft.com/devcontainers/rust:latest",
+	"features": {
+		"ghcr.io/prulloac/devcontainer-features/pre-commit:latest": {}
+	},
+	"postCreateCommand": "pre-commit install",
 	"customizations": {
 		"vscode": {
 			"settings": {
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -1,5 +1,9 @@
 {
 	"image": "mcr.microsoft.com/devcontainers/rust:latest",
+	"features": {
+		"ghcr.io/prulloac/devcontainer-features/pre-commit:latest": {}
+	},
+	"postCreateCommand": "pre-commit install",
```
k88hudson-cfa, line 6: I think I would change this to a print a message "Run pre-commit install in your terminal to install pre-commit hooks" rather than doing it for you automatically

src/web_api.rs
```diff
@@ -101,17 +101,22 @@ async fn serve(
     }
 
     // build our application with a route
+    let path = format!("{}/{}", env!("CARGO_MANIFEST_DIR"), "static/");
+    let local_path = std::path::Path::new(&path);
```
k88hudson-cfa, line 105: maybe call this `static_assets_path` or something?

k88hudson-cfa (changes requested):
Some minor suggestions. Thanks for picking this up

.devcontainer/devcontainer.json
cdc-as81: I would argue the opposite. as a developer, I prefer to wait another 20 seconds and have the dev env setup at the start, without me needing to run a command, every time I create a codespace.
If you think this is an issue, then I would suggest having 2 flavors of codespace, a plain one, and a loaded one.

k88hudson-cfa: I hear you on not wanting to do it every time, I just really think that commit hooks are a matter of personal preference and should be optional. Perhaps something like? [this](https://www.reddit.com/r/git/comments/a5mpff/setting_up_precommit_hooks_to_existing_and_all/ ) would work via [dotfiles](https://docs.github.com/en/codespaces/setting-your-user-preferences/personalizing-github-codespaces-for-your-account#enabling-your-dotfiles-repository-for-codespaces)?

If that's really not going to work, I'd be open to the two configs I think

k88hudson-cfa (approved):

Issue #209, opened 2025-02-10
Allow us to set log levels per module via command line


k88hudson-cfa:
Right now you are able to specify per-module log levels via the API, but we'd also be able to do so through the command line (i.e., by implementing a new command in runner). See [the log-level command](https://github.com/CDCgov/ixa/blob/9f3d50e9a93d769306c6e51c8d940514c5544290/src/runner.rs#L153)

cdc-as81:
how about something like:
cargo run ... -- --modules_log_level "ixa:debug, rustyline:error"

k88hudson-cfa:
yep that sounds fine

PR #231, opened 2025-02-19
Set log levels for modules #209

cdc-as81:


k88hudson-cfa (approved):
LGTM with a small fix

PR #232, opened 2025-02-19
added build command for runner_test_debug or runner_test_custom_args …

cdc-as81:
…when needed by tests


k88hudson-cfa (changes requested):
How about refactor into setup function that returns `assert_cmd::Command::cargo_bin("runner_test_debug")` ?

k88hudson-cfa (approved):

PR #237, opened 2025-02-24
changed log-level modules to accept equal sign instead of colon

cdc-as81:


RobertJacobsonCDC (changes requested):
I'm a little concerned that the syntax is confusing:

```
cargo run -- --log-level=ixa=off,model=trace
```

Is that one command line argument or two? What's the relative operator precedence of `=` and `,`? (The answer is context dependent.) I just find it hard to visually parse.

Two possible alternatives in no particular order:

We _could_ stick with colons. Module path separators in Rust are always two colons: `package_name::outer_module::inner_module`. The parsing logic would be pretty simple: If you see a colon and the next character is also a colon, it's a path separator. Otherwise, it's a module filter separator. One could argue it's still a little confusing visually, though:

```
cargo run -- --log-level=ixa::people:off,model::people:trace
```

Another possibility is to use something like function call syntax in which the log level appears in parentheses:

```
cargo run -- --log-level=ixa::people(off),model::people(trace)
```

Feel free to disagree with me. Maybe everyone else thinks your syntax is superior, who knows?

cdc-as81:
how about:
`cargo run -- --log-level ixa=off model=trace`

RobertJacobsonCDC:
> how about: `cargo run -- --log-level ixa=off model=trace`

Oh, yeah, I guess that works.

RobertJacobsonCDC (approved):

Issue #239, opened 2025-02-25
Clean up basic-infection example


RobertJacobsonCDC:
PR-sized chunk of #182 specific to the first example, "basic infection."

RobertJacobsonCDC:
The source code to the example models are included (using {{#include path/to/file:start_line:end_line}} syntax) in the text, which keeps the text and the example source in sync. Right now the example code is in the models subdirectory, but they need to live in the examples directory in the project root.

PR #241, opened 2025-02-25
Initial draft of The Ixa Book in `docs/book`.

RobertJacobsonCDC:
Tasks required before merge:

- [ ] ~Move examples to project root~ Broken out to #239 
- [x] Quick start and Your First Model aren't quite done yet.
- [ ] ~Render in CI~ separate Issue/PR, #243 
- [x] Probably doesn't pass pre-commit checks.
- [x] Rebase onto main

Addresses Issue #242.


docs/book/src/first_model/2_setup.md
```diff
@@ -0,0 +1,110 @@
...
+        }  
+        Err(e) => {  
+            error!("Simulation exited with error: {}", e);  
+        }  
+    }  
+}
+```
+Don't let this code intimidate you—it's really quite simple. The first line says we want to use symbols from the `ixa` library in the code that follows.  In `main()`, the first thing we do is call  `run_with_args()`. The `run_with_args()` function takes as an argument a closure inside which we can do additional setup before the simulation is kicked off if necessary. The only "setup" we do is log a `trace!` message that we are initializing the model.
+
+> [!INFO] Closures
```
k88hudson-cfa, line 72: Nice

docs/book/book.toml
```diff
@@ -0,0 +1,17 @@
+[book]
+authors = ["Robert Jacobson"]
```
k88hudson-cfa, line 2: What is this used for?

docs/book/models/disease_model/src/people.rs
```diff
@@ -0,0 +1,24 @@
...
+pub enum InfectionStatusValue {
+  S,
+  I,
+  R,
+}
+
+define_person_property_with_default!(
+    InfectionStatus,        // Property Name
+    InfectionStatusValue,   // Type of the Property Values
+    InfectionStatusValue::S // The default value assigned to each person
```
k88hudson-cfa, line 15: Maybe "when a person is added to the simulation"

k88hudson-cfa (changes requested):
This is an awesome start.  I think we're going to want to iterate on what's in the first model, but let's leave it mostly as is for the scope of this PR

Super excited to put this in front of people

docs/book/src/first_model/2_setup.md
```diff
@@ -0,0 +1,110 @@
...
+```bash
+cd ~/Code
+cargo new --bin disease_model
+```
+
+Cargo creates a directory named `disease_model` with a project skeleton for us. Open the newly created `disease_model` directory in your favorite IDE, like [VSCode](https://code.visualstudio.com/) (free) or [RustRover](https://www.jetbrains.com/rust/).
+
+```
+🗂️ disease_model/
+├── .git/
```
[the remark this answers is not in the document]
RobertJacobsonCDC: Yeah, I debated about whether to acknowledge the existence of Git at all. It's out of scope, but... what does it look like on a Windows machine?

docs/book/book.toml
RobertJacobsonCDC: I genuinely don't know. It was in the example I copied&pasted from. In any case, it shouldn't be Robert Jacobson. It should be "The Ixa Developers" or something.

docs/book/src/1_quick_start.md
```diff
@@ -0,0 +1,52 @@
...
+        }  
+        Err(e) => {  
+            error!("Simulation exited with error: {}", e);  
+        }  
+    }  
+}
+```
+
+## Using Ixa From Source
+The recommended way to use Ixa is by including it as a dependency in your model project's `Cargo.toml` file like any other dependency, but there are situations when you may want the latest pre-release Ixa source code.
```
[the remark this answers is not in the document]
RobertJacobsonCDC: Yeah, Gil gave the same feedback. I think anyone who really is interested in the Ixa source code itself doesn't need instructions on how to access it. And someone who wants bleeding edge `main` versus `release` just points to a different branch. So I think we can ax the whole "from source" section.

docs/book/src/first_model/1_your_first_model.md
```diff
@@ -0,0 +1,44 @@
...
+
+In this model, each individual susceptible person has an exponentially distributed time until they are infected. This means the time between successive infection events follows an exponential distribution. The rate of infection is typically expressed as a **force of infection**, which is a measure of the risk of a susceptible individual contracting the disease. In the case of a food-borne illness, this force is constant, meaning each susceptible individual faces a fixed probability per unit time of becoming infected, independent of the number of people already infected.
+## High-level view of how Ixa functions
+
+This diagram gives a high-level view of how Ixa works:
+
+![An Agent Based Model](../assets/AnAgentBasedModel.svg)
+
+Don't expect to understand everything in this diagram straight away. The major concepts we need to understand about models in Ixa are:
```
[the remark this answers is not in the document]
RobertJacobsonCDC: There might be a better place for this in the book as it evolves. I don't think we necessarily need a whole thing on what an agent-based disease model is, but I'm imagining a brand new team member, or maybe someone whose only experience is with ODE models or something, trying to figure out what it is they are being asked to help create.

But I'm definitely open to changing this if we think it's the right thing.

docs/book/book.toml
RobertJacobsonCDC: Also in `Cargo.toml` change it to `"John Doe <jdoe@example.com>"`.

docs/book/src/first_model/2_setup.md
```diff
@@ -0,0 +1,110 @@
...
+cargo run
+```
+## Command line arguments
+The model doesn't do anything yet—it doesn't even emit the log messages we included. We can turn those on to see what is happening inside our model during development with the following command line argument:
+```bash
+cargo run -- --log-level=trace
+```
+This turns on messages emitted by Ixa itself, too. If you only want to see messages emitted by `disease_model`, you can specify the module in addition to the log level:
+```bash
+cargo run -- --log-level=disease_model:trace
```
RobertJacobsonCDC, line 93: Update command line argument syntax for filtering log messages.

docs/book/models/disease_model/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "disease_model"
+description = "A basic SIR disease model using the Ixa agent-based modeling framework"
+authors = ["Robert Jacobson <rjacobson@cdc.gov>"]
```
RobertJacobsonCDC, line 4: John Doe

docs/book/src/0_introduction.md
```diff
@@ -0,0 +1,11 @@
+# Introduction
+
+Ixa (Interactive eXecution of ABMs) is a Rust framework for building modular agent-based discrete event models for large-scale simulations. While its primary application is modeling disease transmission, its flexible design makes it suitable for a wide range of simulation scenarios.
+
+<img align="left" alt="Ixa crab" src="./assets/ixa_logo.svg" width="200px" />
```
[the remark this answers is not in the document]
RobertJacobsonCDC: Also the diagram(s), which aren't checked in yet. Maybe put them in an "Ixa Internals" section just so they have a home?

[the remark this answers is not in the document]
RobertJacobsonCDC: Unfortunately, mdBook can't do that for some reason. But git is symlink aware, so I just made a link.

paigemiller:
I think this will be really helpful for people. Some thoughts as I was going through it: 

* I wished there was a description of cargo earlier, maybe comparing it to python equivalents of package mangers
* in the installing ixa quick start, src/main.rs is confusing calling it a disease model when it doesn't do anything. I would call it faux_model or something to emphasize that it does nothing
* I'm not sure logging and printing is helpful, without an explanation. it just sort of looks like hard-to-read code until the explanation call out comes later. Could you show the option of println!() and then later introduce ixa specific logging
* early on, can you give resources on type hinting and any helpful extra set up instructions to make peoples experiences better
* I like the simple food model, I think this is a good choice and I'm assuming a simple SI model could be a good follow up?
* I do think the diagram looks very nice, but is a bit more complicated looking than it needs to be. Suggest simplifying a bit more and removing some text if possible.
* In the model, why do you need to define another population_size variable rather than just using POPULATION? 
* I still think closures are confusing but the callout boxes are helpful. 
* A rust glossary with a quick and dirty explanation of !, ?, ownership, mut, borrow, _, &, some/none might be helpful to people who think this syntax is confusing to see early on or be able to return to as a reference as they are learning ixa/rust

RobertJacobsonCDC:
In task list form:

- [ ] I wished there was a description of cargo earlier, maybe comparing it to python equivalents of package mangers
- [ ] in the installing ixa quick start, src/main.rs is confusing calling it a disease model when it doesn't do anything. I would call it faux_model or something to emphasize that it does nothing
- [ ] I'm not sure logging and printing is helpful, without an explanation. it just sort of looks like hard-to-read code until the explanation call out comes later. Could you show the option of println!() and then later introduce ixa specific logging
- [ ] early on, can you give resources on type hinting and any helpful extra set up instructions to make peoples experiences better
- [ ] I like the simple food model, I think this is a good choice and I'm assuming a simple SI model could be a good follow up?
- [ ] I do think the diagram looks very nice, but is a bit more complicated looking than it needs to be. Suggest simplifying a bit more and removing some text if possible.
- [ ] In the model, why do you need to define another population_size variable rather than just using POPULATION?
- [ ] I still think closures are confusing but the callout boxes are helpful.
- [ ] A rust glossary with a quick and dirty explanation of !, ?, ownership, mut, borrow, _, &, some/none might be helpful to people who think this syntax is confusing to see early on or be able to return to as a reference as they are learning ixa/rust

k88hudson-cfa (approved):

PR #245, opened 2025-02-26
Return an Option from sample_person

### LINKED ISSUE #191: sample_person should return an Option, not a Result
I think it's pretty weird that sample_person returns an IxaError when there are no people left ... we should probably return an Option instead

k88hudson-cfa:
Fixes #191


RobertJacobsonCDC (approved):
This is an easy one. 👍

PR #246, opened 2025-02-26
Upgraded rand crate dependency to v0.9.0. Fixes #238.

### LINKED ISSUE #238: Upgrade rand dependency to 9.0
Issue #208 identified rand 0.9 as incompatible with Ixa. We decided not to upgrade at that time.

Client code often depends on rand_distr, which itself depends on rand 9.0 unless you downgrade rand_distr, and it gives a confusing error that doesn't make it clear what the problem is.

RobertJacobsonCDC:
- Upgraded rand crate dependency to v0.9.0, rand_distr to v0.5.1, and adjusted code throughout to accommodate breaking changes.
- Moved roots dependency to dev dependencies.
- Added "or higher" version syntax to web interface dependencies. 
- Removed unused ordered_float dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,14 +9,13 @@ homepage = "https://github.com/CDCgov/ixa"
 
 [dependencies]
 fxhash = "^0.2.1"
-rand = "^0.8.5"
+rand = "^0.9.0"
 csv = "^1.3.1"
 serde = { version = "^1.0.217", features = ["derive"] }
 serde_derive = "^1.0.217"
 serde_json = "^1.0.135"
 bincode = "^1.3.3"
 reikna = "^0.12.3"
-roots = "0.0.8"
 ixa-derive = { path = "ixa-derive" }
 seq-macro = "^0.3.5"
 paste = "^1.0.15"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -25,20 +24,20 @@ clap = { version = "^4.5.26", features = ["derive"] }
...
 rustyline = "^15.0.0"
 log = "^0.4.22"
-log4rs = { version = "^1.3.0", features = ["console_appender", "pattern_encoder"] }
-axum = "0.8.1"
-tokio = { version = "1", features = ["full"] }
-reqwest = { version = "0.12.12", features = ["blocking", "json"] }
-uuid = "1.12.1"
-tower-http = { version = "0.6.2", features = ["full"] }
-mime = "0.3.17"
+log4rs = { version = "^1.3.0", default-features = false, features = ["console_appender"] }
```
k88hudson-cfa, line 27: Can you say more about this change?

k88hudson-cfa (approved):
Thanks! One question but otherwise good to merge.

Cargo.toml
RobertJacobsonCDC: Yeah, it turns out we ONLY need the `console_appender` feature (which implies `pattern_encoder`), and in fact if we allow all default features then `log4rs` pulls in the older incompatible `rand` as a dependency. This change doesn't limit us in the future. As I recall, the `rand` dependency is for the rolling log files feature. It's just a matter of import what you use.

Issue #221, opened 2025-02-13
New ixa project template


k88hudson-cfa:
I've been thinking about how it would be nice to have a one-click / single-command way to set up a new Ixa project with some basic stuff (like runner) set up for you.

Some options here: 
* A GitHub template
* via a command line tool
 
Personally I think it would be better to keep the source code for this inside this repository so it's fixed to the corresponding version of Ixa

ekr-cfa:
This seems like a good idea.

PR #248, opened 2025-02-28
added script to setup new ixa projects

cdc-as81:
using the script should be documented in the book:
`export ixa_branch=release && curl -s --variable %B=$ixa_branch --expand-url https://raw.githubusercontent.com/CDCgov/ixa/{{B:url}}/scripts/setup_new_ixa_project.sh | sh -s $ixa_branch`

for testing until this is merged you can run:
`export ixa_branch=cdc-as81-script-to-setup-new-ixa-project-#221 && curl -s --variable %B=$ixa_branch --expand-url https://raw.githubusercontent.com/CDCgov/ixa/{{B:url}}/scripts/setup_new_ixa_project.sh | sh -s $ixa_branch`

### DIFF
--- a/examples/basic/main.rs
+++ b/examples/basic/main.rs
@@ -1,0 +1,11 @@
+use ixa::run_with_args;
+
+fn main() {
+    run_with_args(|context, _, _| {
+        context.add_plan(1.0, |context| {
+            println!("The current time is {}", context.get_current_time());
+        });
+        Ok(())
+    })
+    .unwrap();
+}
--- a/scripts/setup_new_ixa_project.sh
+++ b/scripts/setup_new_ixa_project.sh
@@ -21,7 +21,6 @@ if ! [ -f "Cargo.toml" ]; then
 fi
 
 cargo add --git "https://github.com/CDCgov/ixa" ixa --branch $ixa_branch
-cargo add clap
 
 # add .gitignore from ixa
 curl -o .gitignore https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/.gitignore
@@ -34,5 +33,7 @@ mkdir -p .github/workflows
 curl -o .github/workflows/build-test.yaml https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/template/.github/workflows/build-test.yaml
 curl -o .github/workflows/pre-commit.yaml https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/template/.github/workflows/pre-commit.yaml
 
-# override main.rs with ixa runner example
-curl -o src/main.rs https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/examples/runner/main.rs
+# override main.rs with ixa basic example
+curl -o src/main.rs https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/examples/basic/main.rs
+
+echo "Project setup complete"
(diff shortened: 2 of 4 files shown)


RobertJacobsonCDC (changes requested):
The only thing I would add is a message telling the user that the script completed successfully. 

**Edit:** Actually, now that I think about it, we should probably produce messages explaining to the user what we are doing at each step. It's easy enough to do and makes the whole process less opaque to the user.

RobertJacobsonCDC:
This is a separate issue, but we should probably re-export `Args` if it's a type we expect users to use in our public API. There's a way to specify that a dependency of Ixa is a "public" dependency which means exactly that, but I'm not sure it's been stabilized.

scripts/setup_new_ixa_project.sh
```diff
@@ -0,0 +1,39 @@
+# to be run in a new project directory
+# Usage: ./setup_new_ixa_project.sh <ixa-branch>
+# ixa-branch: the branch of ixa to use, default is release
+ixa_branch="release"
```
k88hudson-cfa, line 4: Can you add a comment about this not being allowed to contain special characters, or else make sure you are URL encoding

scripts/setup_new_ixa_project.sh
```diff
@@ -0,0 +1,39 @@
...
+# add the pre-commit hook from ixa
+curl -o .pre-commit-config.yaml https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/.pre-commit-config.yaml
+
+# add github action from ixa
+mkdir -p .github/workflows
+curl -o .github/workflows/build-test.yaml https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/template/.github/workflows/build-test.yaml
+curl -o .github/workflows/pre-commit.yaml https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/template/.github/workflows/pre-commit.yaml
+
+# override main.rs with ixa basic example
+curl -o src/main.rs https://raw.githubusercontent.com/CDCgov/ixa/$ixa_branch/examples/basic/main.rs
```
k88hudson-cfa, line 37: Ideally if this gets an error response it should exit not print the error

examples/basic/main.rs
RobertJacobsonCDC, line 1: Do we want to synchronize this with the [Quick Start](https://github.com/CDCgov/ixa/blob/RobertJacobsonCDC_the_ixa_book/docs/book/src/1_quick_start.md) and [First Model](https://github.com/CDCgov/ixa/blob/RobertJacobsonCDC_the_ixa_book/docs/book/src/first_model/2_setup.md#executing-the-ixa-model) examples in the book? In fact, mdBook can include source files (or specific lines therefrom), so if we change the skeleton, the text of the book changes to coincide.

k88hudson-cfa (changes requested):
This is mostly working as expected a couple of comments

scripts/setup_new_ixa_project.sh
```diff
@@ -0,0 +1,39 @@
+# to be run in a new project directory
+# Usage: ./setup_new_ixa_project.sh <ixa-branch>
+# ixa-branch: the branch of ixa to use, default is release
+ixa_branch="release"
+if [ -n "$1" ]; then
+    ixa_branch=$1
+fi
+
```
k88hudson-cfa, line 8: Can you print "using ixa branch..." (including if it's the default)

scripts/setup_new_ixa_project.sh
cdc-as81: url encoding added!

scripts/setup_new_ixa_project.sh
cdc-as81: added!

examples/basic/main.rs
cdc-as81: see description

k88hudson-cfa (approved):
Thanks!

RobertJacobsonCDC (approved):

Issue #244, opened 2025-02-25
Add exit command


k88hudson-cfa:
Lets add a command (that is listed in help) which exits the simulation directly from the debugger;

WE should also change the behavior of Ctrl+D to do this instead of continue.

RobertJacobsonCDC:
I should have caught this one, but the implementation should use `Context::shutdown_requested` for a clean exit instead of [`std::process::exit`](https://doc.rust-lang.org/std/process/fn.exit.html), because:

> Note that because this function never returns, and that it terminates the process, no destructors on the current stack or any other thread’s stack will be run.

k88hudson-cfa:
We can discuss in the weekly meeting, but I think I disagree with this – I don't think we want a clean exit here

RobertJacobsonCDC:
Closing as per discussion.

PR #249, opened 2025-02-28
Breakpoint, "next", and "halt" implementation. Fixes #163.

### LINKED ISSUE #163: Add support for breakpoints in debugger
Breakpoints should persist for the session.
 
- **`breakpoint set <t>]`**

  Set a breakpoint a given time

  Example: `breakpoint set 4.0`

- **`breakpoint list`**

  List all active breakpoints.

- **`breakpoint delete [<id> | --all]`**

  Delete the breakpoint with the specified id.
  Providing the `--all` option removes all breakpoints.

  Example: `breakpoint delete 1`

RobertJacobsonCDC:
# Breakpoints and Next

- It reuses `plan::Queue` for `Context::breakpoints_scheduled`, but I implemented `next_time()` and `peek()` on it, the latter of which required that `Entry<P>` be `pub(crate)`. 

- `Context::request_debugger()` and `Context::cancel_debugger_request()` are a setter and resetter respectively for `Context::break_requested`, which works like `Context::shutdown_requested()` but takes priority.

- Breakpoints scheduled at a particular time run before the next scheduled plan if the time of the next scheduled plan is `<=` or `<`, according to whether the breakpoint was scheduled with `ExecutionPhase::First` or `ExecutionPhase::Last`. (What about `ExecutionPhase::Normal`?)

- The "next" command that takes a time has been renamed to "breakpoint", and a new "next" command was created. In other words: `breakpoint 1.2` sets a breakpoint at time `1.2`, while `next` executes the next thing.

- Method `schedule_debugger()` is now implemented directly on `Context`, because the data member `breakpoints_scheduled` is in `Context`. There is now no need for `ContextDebuggerExt`

  

Vanity changes:

- Format times to 6 decimal places when printed in debugger, trace statements. I think I should also strip trailing zeros past the decimal point.
- Added a "halt" debugger command (Related: #244 )

I'm not sure I'm happy with this implementation. My least favorite thing about this is it kind of makes a mess of `plan::Queue`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -2,10 +2,10 @@
 //!
 //! Defines a `Context` that is intended to provide the foundational mechanism
 //! for storing and manipulating the state of a given simulation.
-use crate::{HashMap, HashMapExt};
 use crate::debugger::enter_debugger;
 use crate::plan::{PlanId, PlanSchedule, Queue};
 use crate::{error, trace};
+use crate::{HashMap, HashMapExt};
 use std::fmt::{Display, Formatter};
 use std::{
     any::{Any, TypeId},
@@ -76,7 +76,7 @@ pub struct Context {
     callback_queue: VecDeque<Box<Callback>>,
     event_handlers: HashMap<TypeId, Box<dyn Any>>,
     data_plugins: HashMap<TypeId, Box<dyn Any>>,
-    breakpoints_scheduled: Queue<(), ExecutionPhase>,
+    breakpoints_scheduled: Queue<Box<Callback>, ExecutionPhase>,
     current_time: f64,
     shutdown_requested: bool,
     break_requested: bool,
@@ -107,10 +107,16 @@ impl Context {
     /// # Errors
     /// Internal debugger errors e.g., reading or writing to stdin/stdout;
     /// errors in Ixa are printed to stdout
-    pub fn schedule_debugger(&mut self, time: f64, priority: Option<ExecutionPhase>) {
+    pub fn schedule_debugger(
+        &mut self,
+        time: f64,
+        priority: Option<ExecutionPhase>,
+        callback: Box<Callback>,
+    ) {
         trace!("scheduling debugger");
         let priority = priority.unwrap_or(ExecutionPhase::First);
-        self.breakpoints_scheduled.add_plan(time, (), priority);
+        self.breakpoints_scheduled
+            .add_plan(time, callback, priority);
     }
 
     /// Register to handle emission of events of type E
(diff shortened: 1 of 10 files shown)


RobertJacobsonCDC:
Also, this needs some tests. Converted to a draft.

RobertJacobsonCDC:
Modify debugger commands to match #163.

RobertJacobsonCDC:
There are two failing tests that still need to be figured out.

src/context.rs
```diff
@@ -214,7 +245,10 @@ impl Context {
     /// cancelled or executed.
     pub fn cancel_plan(&mut self, plan_id: &PlanId) {
         trace!("canceling plan {:?}", plan_id);
-        self.plan_queue.cancel_plan(plan_id);
+        let result = self.plan_queue.cancel_plan(plan_id);
```
RobertJacobsonCDC, line 248: This used to panic if the `plan_id` didn't exist, but now users are typing `plan_id`'s into the terminal a nonexistent `plan_id` does not necessarily mean a logic error in the program.

src/context.rs
```diff
@@ -280,33 +314,125 @@ impl Context {
...
+
     /// Execute the simulation until the plan and callback queues are empty
     pub fn execute(&mut self) {
         trace!("entering event loop");
         // Start plan loop
         loop {
-            if self.shutdown_requested {
+            if self.break_requested {
+                self.break_requested = false;
+                enter_debugger(self);
```
RobertJacobsonCDC, line 368: This execution path is not used for scheduled breakpoints. It exists so client code can request a breakpoint without scheduling one. ("break now").

src/debugger.rs
```diff
@@ -263,54 +319,29 @@ fn start_debugger(context: &mut Context, debugger: &mut Debugger) -> Result<(),
 
         match debugger.process_command(line, context) {
             Ok((quit, message)) => {
+                if let Some(message) = message {
+                    println!("{message}");
```
RobertJacobsonCDC, line 323: The original code wrote to `std::io::stdout`. Maybe the intention was to make sure the buffer gets flushed after printing the message. But `println` and `eprintln` (but not `print`) implicitly flush the buffer.

src/debugger.rs
```diff
@@ -156,27 +156,71 @@ impl DebuggerCommand for GlobalPropertyCommand {
     }
 }
 
-struct NextCommand;
+/// Exits the debugger and ends the simulation.
+struct HaltCommand;
```
RobertJacobsonCDC, line 160: Added `halt` command here and in the web API but not in the web frontend.

src/external_api.rs
```diff
@@ -110,26 +237,45 @@ pub(crate) mod next {
...
-        },
+        /// Execute the next callback or scheduled plan and return to debugger
+        Next,
     }
     #[derive(Serialize)]
     pub(crate) struct Retval {}
+    #[allow(unused)]
     pub(crate) struct Api {}
     impl super::ExtApi for Api {
         type Args = Args;
```
k88hudson-cfa, line 248: I think this actually has to be EmptyArgs

RobertJacobsonCDC: Yep!

k88hudson-cfa (approved):
LGTM with rebase

PR #253, opened 2025-03-04
changed handling of ctrl-D in debugger to exit program

cdc-as81:


src/debugger.rs
```diff
@@ -244,7 +249,7 @@ fn start_debugger(context: &mut Context, debugger: &mut Debugger) -> Result<(),
                 rustyline::error::ReadlineError::WindowResized
                 | rustyline::error::ReadlineError::Interrupted,
             ) => continue,
-            Err(rustyline::error::ReadlineError::Eof) => return Ok(()),
+            Err(rustyline::error::ReadlineError::Eof) => exit_debugger().to_string(),
```
k88hudson-cfa, line 252: What is `to_string` doing here?

src/debugger.rs
```diff
@@ -231,6 +231,11 @@ fn init(context: &mut Context) {
     }
 }
 
+fn exit_debugger() -> String {
```
k88hudson-cfa, line 234: I think this should work if you return a never type, like `!`

src/debugger.rs
cdc-as81: removed

k88hudson-cfa (approved):
LGTM

PR #269, opened 2025-03-07
fixed setup script to not use variables with curl

cdc-as81:


RobertJacobsonCDC:
The issue this PR solves is, git branch names can have "special" characters, and if a user asks the script to fetch from a branch with a special character it can break the script, potentially in a confusing way. 

The original solution used a command line option to `curl`, but it turns out we have users on systems with older versions of `curl` that don't have that feature. 

This PR solves the problem by URL encoding the branch name with a function in pure bash and `printf` (which is POSIX standard).

This might seem like an unlikely problem to run into, but it happened to the PR's author himself!

RobertJacobsonCDC (approved):
Looks good to me!

Issue #272, opened 2025-03-10
Replace Set and Hashmap in people/network with deterministic versions


k88hudson-cfa:
This is what is causing the behaviour observed in #265; queries use ` HashSet`, which returns people in an undeterministic order.

k88hudson-cfa:
It would be nice to add some docs about this generally too, for people implementing models

RobertJacobsonCDC:
PR #273

Issue #265, opened 2025-03-06
Possible non-deterministic behavior with random?


k88hudson-cfa:
We're observing some behavior in https://github.com/cdcent/cfa-ixa-household-model/issues/27 where the times of infection are stable but the ids of the people are non-deterministic.

Some possible theories
- Something about loading the population from a csv?
- Something is broken in rng?
- Using multiple Rng instances across different files ends up with non-deterministic behavior?

RobertJacobsonCDC:
If we don't have tests for the built-in rngs to make sure they are deterministic, we should.

k88hudson-cfa:
This could be related to `HashMap`/`Set` order being non-deterministic

RobertJacobsonCDC:
I think `HashMap`/`HashSet` ordering is unspecified but deterministic.

RobertJacobsonCDC:
#222 might fix this, actually. As part of that ticket we should either 
1. audit all uses of `HashMap`/`HashSet` to make sure they aren't used in a way that introduces nondeterminism (basically any time they are iterated over), or
2. swap out the default `HashMap`/`HashSet` with a deterministic version every single place we use it.

The first option is easy, but it's also easy to accidentally introduce nondeterminism later. If we just have a rule that we always use the deterministic versions everywhere, we won't have to worry about it.

Issue #222, opened 2025-02-13
Use a fast hashing algorithm for structures in the hot path


RobertJacobsonCDC:
A cryptographic hash function is unnecessary for simulation purposes. DDOS resistance is unlikely to be a concern. Consider [`fxhash`](https://crates.io/crates/fxhash) (FireFox & rustc), [`ahash`](https://crates.io/crates/ahash), or [`gxhash`](https://github.com/ogxd/gxhash) (x86 & ARM intrinsics).

RobertJacobsonCDC:
Note that this change not only improves performance, it also ensures determinism, because the standard library uses a random seed as a DoS mitigation.

I refactored the imports of `HashMap` and `HashSet` in Ixa to import from Ixa's root module so that it is easy to swap out implementations if we need to. I did this for *every* instance of a use of these structures in the codebase, not just those cases that obviously affect the determinism of the model. This seemed safer and easier to maintain going forward.

There are many choices of alternative hash functions. My short list is:

- fxhash: widely used, well-established, hash algorithm used in Firefox
- gxhash: widely used, mature, but does not have determinism feature
- ahash: young, not widely used, but fast, and has determinism feature built in

The fxhash crate would be the obvious choice, but [it is unmaintained.](https://github.com/rustsec/advisory-db/issues/2185) Because of its maturity, I'd prefer gxhash. I have [submitted a PR](https://github.com/ogxd/gxhash/pull/114) to the gxhash project to add a feature flag to enable determinism. The rustc-hash crate was recently recommended to me, but I haven't investigated it yet.

RobertJacobsonCDC:
PR #273

PR #273, opened 2025-03-11
All instances of `HashMap` and `HashSet` now deterministic.

RobertJacobsonCDC:
All instances of `HashMap` and `HashSet` now import from the crate root so that it is easy to replace implementation. 
Deterministic implementations of both data structures are provided in `hashing.rs`. 
The same deterministic algorithm is used in `random.rs` for name offset in RNGs.
Fixed a bug in `get_rng()` where adding the seed offset might saturate.
Allowed "implicit hasher" for clippy.

This PR addresses the following issues:

- #222 
- #265 
- #272

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,6 @@ license = "Apache-2.0"
 homepage = "https://github.com/CDCgov/ixa"
 
 [dependencies]
-fxhash = "^0.2.1"
 rand = "^0.8.5"
 csv = "^1.3.1"
 serde = { version = "^1.0.217", features = ["derive"] }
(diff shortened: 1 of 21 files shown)


k88hudson-cfa (changes requested):
Can we add a lint rule to disallow HashMap/HashSet https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_types with a suggested alternative?

k88hudson (approved):

k88hudson-cfa (approved):

RobertJacobsonCDC:
> Can we add a lint rule to disallow HashMap/HashSet [rust-lang.github.io/rust-clippy/master/index.html#disallowed_types](https://rust-lang.github.io/rust-clippy/master/index.html#disallowed_types) with a suggested alternative?

Added as issue #278.

PR #277, opened 2025-03-17
feat: add release-plz action

cdc-as81:
By default, version and changelog should be updated automatically by release-plz.
if we need to set a version, we can run locally:
`release-plz set-version 1.2.3  `

Note that the action creates a PR, we can edit the PR before merge.

### DIFF
--- a/.github/workflows/release-plz.yaml
+++ b/.github/workflows/release-plz.yaml
@@ -9,10 +9,6 @@ on:
     branches:
       - release
 
-env:
-  GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
-  CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
-
 jobs:
 
   # Release unpublished packages.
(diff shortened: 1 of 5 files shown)


.github/workflows/release-plz.yaml
```diff
@@ -0,0 +1,79 @@
...
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      # Generating a GitHub token, so that PRs and tags created by
+      # the release-plz-action can trigger actions workflows.
+      - name: Generate GitHub token
+        uses: actions/create-github-app-token@21cfef2b496dd8ef5b904c159339626a10ad380e # v1.11.6
```
k88hudson-cfa, line 24: Can you add a comment about how to keep this up to date? / who to contact for more info

k88hudson-cfa (approved):
This looks good to me to give it a try. Can you file a follow up about figuring out how we keep this up to date with new versions of actions? 

If we want to specify a specific release version (v.s. using commits)/ edit the changelog, how do we do that?

.github/workflows/release-plz.yaml
cdc-as81: I fixed all imported actions, as this is safer. Updating would be manual, need to go to the action github and bring the git-hash for the version you want to update.

PR #298, opened 2025-03-24
feat(doc) adding markdown to pre-commit

cdc-as81:
Added markdown plugin to pre-commit.
Note that in order to add it I had to cleanup many md files.


.markdownlint.json
```diff
@@ -0,0 +1,4 @@
+{
+    "no-inline-html": false,
+    "line-length": false
```
RobertJacobsonCDC, line 3: The lint is supposed to limit line lengths to 100 characters. Should this be the integer `100` instead of a bool?

src/global_properties.rs
```diff
@@ -57,7 +57,7 @@ pub fn add_global_property<T: GlobalProperty>(name: &str)
 where
     for<'de> <T as GlobalProperty>::Value: serde::Deserialize<'de> + serde::Serialize,
 {
-    trace!("Adding global property {}", name);
+    trace!("Adding global property {name}");
```
RobertJacobsonCDC, line 60: I actually disagree with this clippy lint.  I vote we disable it. Do you have any opinion on this? 

I'd also like to disable the one about documenting panics in doc comments, not because I disagree with it, but because I feel like it's overkill to have it on every thing function. But maybe that's off topic right now.

RobertJacobsonCDC (changes requested):

.markdownlint.json
cdc-as81: I currently disabled 'line-length' since it includes a lot of changes. This will come as part2.

src/global_properties.rs
cdc-as81: I actually like the named arguments, but let's talk about it.

PR #300, opened 2025-03-26
fix: updated setup script to use released ixa in cargo

cdc-as81:


RobertJacobsonCDC (approved):
Works for me.

scripts/setup_new_ixa_project.sh
```diff
@@ -1,23 +1,23 @@
 # to be run in a new project directory
```
RobertJacobsonCDC, line 1: Maybe capitalize the "to" and use a ".".

Issue #180, opened 2025-01-29
Visibility of internal implementation of the `people` module.


RobertJacobsonCDC:
While refactoring the `people` module I found some items with public visibility that I think should be private to `people`. The items are:

```rust
people::index::process_indices
people::index::Index
people::index::IndexValue
people::query::Query
people::data::PersonPropertyHolder
``` 

These are where they live after the refactor. All of these were `pub` in the original `people` module. I think they can all be `pub(super)` (visible internally to `people`) in the refactor.

k88hudson-cfa:
Some of these are needed from macros, could use comment

RobertJacobsonCDC:
Folding into ongoing work on #303, so closing in favor of those related issues.

PR #303, opened 2025-04-03
fix: Internal refactor of queries and indexing. Fixes #301.

### LINKED ISSUE #301: Bug in querying derived person properties prior to their explicit initialization.
Over in ixa-epi land, @KOVALW suggested that I make some code more efficient by using the sample people based on a query functionality. In the code, I am sampling by a derived person property (`InfectionStatus`, derived from `InfectionData`), and I have not yet accessed the value of that person property for anybody. When I do so, I get an error. In particular, I get `BorrowMut` error on line 330 of people's context extension: https://github.com/CDCgov/ixa/blob/bf12b0e4903bed79c5cba6d4505102fda57fae4c/src/people/context_extension.rs#L330.

Indexing `InfectionStatus` did not resolve the issue. Instead, I resolved the issue by getting the value of `InfectionStatus` for a random person prior to running the query: https://github.com/CDCgov/ixa-epi-isolation/pull/80/commits/22310e4d1e96a5ffdce2cfd20ab443bbc191f0ed.

Here's a reproducible example with the GitHub actions showing the error too: https://github.com/CDCgov/ixa-epi-isolation/pull/80/commits/8de533a90f11b145bc18cead4fd25f88a38020d2.

RobertJacobsonCDC:
# Refactor

## Commit 1: Lookup value moved to property

- **Looking up a value (`get_person_property`) and computing a value (`PersonProperty::compute`) has been combined into `PersonProperty::compute`**

This changes the semantics of `compute`. The computation of the initial value is now in it's own method `initial_value`.

Both `compute` and `initial_value` return an `Option<T>`, which is consistent with the semantics we agreed on some time ago. 

The justification is that the property (or the modeler who defined it) should contain the logic about it's initial value and initialization behavior. Also, a few commits from now registration and dependencies will be moved to `PersonProperty`.

## Commit 2: Removed redundant `+'static` bound on `PersonProperty`

- The `+ 'static` is redundant for `PersonProperty`, as every `PersonProperty` is already required to be static. So I changed all instances of `PersonProperty + 'static` to just `PersonProperty`.

## Commit 3: Dependency collection, property registration

- **Logic to collect dependencies moved to `PersonProperty`**

`PersonProperty::dependencies` has been renamed to `collect_dependencies` to reflect expanded function.

`PersonProperty::dependencies` now returns itself if it is not derived.

- **Unnecessary indirection for property registration removed**

We use dynamic dispatch to register derived and non derived properties differently. 

The internal registration functions were moved into `ContextPeopleExtInteral`, as they will never be called by client code.

### Moved query execution to `Query`

The reason to have `Data::methods` is to map `TypeId` to functionality in `T: Property` in the absense of the particular `T`. However, `Query` always has access to all of its `T`'s. 

Also, it seems more natural to have querying functionality within `Query` instead of an add-on to `Context`.

### Moved tabulation to `Tabulator`

For the same reasons, and in the same way as `Query::execute_query`.

### Moved borrow of `Methods` closer to use

The bug of #301 is caused by a double borrow of `Data::methods`. Moving the borrow closer to its use minimizes the potential error surface.

## Commit 5: Replaced `std::any::TypeId::of::<T>()` with `type_of::<T>()`.

Half the code used this convention already. This just makes it consistent. 

## Commit 6: initialization of person properties even lazier

The default behavior was to _insert_ an initial value for a `PersonId` (which might be `None`) into the property vector whenever the property is first accessed for a person. If the vector is not large enough, it is extended by inserting `None` in the new slots. It turns out this is expensive. Instead, if a `PersonId` does not have a value stored in the property vector, the initial value defined for that property is immediately returned, that is, `get_person_property` _does not_ insert the initial value into the data store for the person. Client code that wants to do that override the appropriate method in their implementation of the `PersonProperty` trait.

This optimization significantly improves performance for some access patterns. It also allows the removal of the creation of index during property registration, which fixes the bug of #301.

# Not Done

I didn't do everything I would have liked. Some things of significance are:

- `Context::get_person_property()` should return an option.
- Remove usages of `RefCell<T>` - Needs discussion
- `query::QueryAnd` not implemented. Multiple ways to do this, but do we need it?
- Remove unnecessary `PersonPropertyHolder`

## Other Issues and Notes

- Some items had more visibility that they require. I restricted visibility as much as possible for many such cases. This addresses #180.
- `ContextPeopleExt::tabulate_person_properties` does a check for the existence of `PeopePlugin` at a time when a panic would have already occurred if `PeoplePlugin` wasn't initialized.
- `tabulate_person_properties` was rewritten to mirror `Query::execute_query()`. 
- The version of `impl Tabulator for Vec<(TypeId, String)> is fallible. What should we do on error?

## Refactors affecting public API

These haven't been done, either.

- Remove tag types: `(PersonProperty, PersonPropertyValue) -> PersonProperty(value)`
- Remove the `String` from the hash map's value in `Index`'s implementation: `HashMap<IndexValue, (String, HashSet<PersonId>)>` becomes `HashMap<IndexValue, HashSet<PersonId>>`. (In separate branch.)


jasonasher:
I didn't have time yet to go through this in depth but I think there are some conceptual issues here.

Your note about for Commit 6 about making person properties even lazier wouldn't be correct. Suppose I have a person property that just stores a pseudorandom number between 0 and 1 for each person. If the client requests this value the first time then the initializer must be run - precisely once - to compute this random value and then store it. If you do not store it then the second time you will get a different pseudorandom value, violating the semantics of the property system.

I have some larger thoughts about some of the other refactors and I wonder if there are other semantic changes being introduced. I think this PR needs to be replaced with first one that addresses #301 alone with the smallest number of changes to the rest of the code to address that issue (as well as appropriate tests). Then one could file issues for some of the other changes you have in mind, as I think they will likely require discussion.

src/lib.rs
```diff
@@ -72,10 +71,19 @@ pub mod web_api;
...
+pub use seq_macro::seq;
 
 // Deterministic hashing data structures
 pub use crate::hashing::{HashMap, HashMapExt, HashSet, HashSetExt};
 
+pub use std::any::TypeId;
+
+#[must_use]
+#[inline(always)]
+pub fn type_of<T: 'static>() -> TypeId {
```
k88hudson-cfa, line 83: Can we just use `TypeId::of`?

k88hudson-cfa (changes requested):
Let's take out the semantic changes to initialization

k88hudson-cfa (changes requested):
I think you're right that a bunch of stuff here needs to be refactored/cleaned up but I think I'd like to merge a smaller targeted fix for the bug first (#312) and then revisit this in pieces; ideally with some more tests.

Removing indexes during property registration seems like good idea to me (also, if you want to break out removing those redundant statics we can merge that too) – the semantic changes to initialization I'm still not so sure about, but we can look at it separately.

Sorry it took so long to understand what was going on here, it was tough to get my head around!

RobertJacobsonCDC:
Closing as per discussion.
- The work in this PR will be broken out into multiple more manageable PRs.
- The bug this PR addresses has been addressed in a more targeted PR #312.

PR #309, opened 2025-04-18
feat: Support multi-property indexes

cdc-as81:
Adding support for multi property index

usage:
if you have the following properties defined: Age, County, Height
in your function you can call:
   define_multi_property_index!(AgeCountyHeight, [Age, County, Height]);

and now you can just query the defined properties, like:
  let result = context.query_people(((Height, 140), (County, 1), (Age, 28)));

### DIFF
--- a/src/people/index.rs
+++ b/src/people/index.rs
@@ -1,17 +1,21 @@
 use super::methods::Methods;
 use crate::{Context, ContextPeopleExt, PersonId, PersonProperty};
-use crate::{HashMap, HashSet, HashSetExt};
+use crate::{HashMap, HashMapExt, HashSet, HashSetExt};
 use bincode::serialize;
 use serde::Serialize;
+use std::any::TypeId;
+use std::cell::RefCell;
+use std::collections::hash_map::DefaultHasher;
+use std::hash::{Hash, Hasher};
+use std::sync::LazyLock;
+use std::sync::Mutex;
 
-#[derive(Clone, PartialEq, Eq, Hash, Debug)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, serde::Serialize)]
 // The lookup key for entries in the index. This is a serialized version of the value.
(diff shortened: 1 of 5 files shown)


src/people/index.rs
```diff
@@ -27,8 +34,10 @@ impl IndexValue {
 
             IndexValue::Fixed(u128::from_le_bytes(tmp))
         } else {
-            // Otherwise, store it as `IndexValue::Variable`
-            IndexValue::Variable(serialized_data)
+            let mut hasher = DefaultHasher::new();
```
k88hudson, line 37: Ah, we actually didn't want to use hashing because it's not collision-proof. Maybe we can have the macro generate a tuple and ditch the sorting for now?

k88hudson (changes requested):

src/people/index.rs
cdc-as81: as discussed, we will have fixed(U128), med64, med256 and hash for size > u8x256.

src/people/index.rs
```diff
@@ -97,6 +120,66 @@ impl Index {
...
+
+pub struct MultiIndex {
+    register: Box<IndexCallback>,
+    type_id: TypeId,
+}
+
+// / A static map of multi-property indices. This is used to register multi-property property indices
+// / when they are first created. The map is keyed by the property IDs of the properties that are
+// / indexed. The values are the `MultiIndex` objects that contain the registration function and
+// / the type ID of the index.
```
k88hudson-cfa, line 134:
```suggestion
// A static map of multi-property indices. This is used to register multi-property property indices
// when they are first created. The map is keyed by the property IDs of the properties that are
// indexed. The values are the `MultiIndex` objects that contain the registration function and
// the type ID of the index.
```

k88hudson-cfa (approved):
LGTM. As discussed, we are merging this for now and filing a follow-up to replace the Indexer with something that would avoid or more appropriately handle possible collisions (such as storage/comparison of the actual values).

PR #310, opened 2025-04-22
Change range of lines of code to display define_rng!

eqmooring:
I believe that the line
https://github.com/CDCgov/ixa/blob/4be6bca8e436481a935720caefefcd7fdac034ff/docs/book/models/disease_model/src/transmission_manager.rs#L10

is missing from the following code chunk:

image


k88hudson-cfa:
@cdc-as81 was this fixed in another PR?

cdc-as81 (approved):
LGTM

PR #312, opened 2025-04-25
Fix double borrow during property registration

k88hudson-cfa:
AFAICT the bug seemed to be introduced by ee3c219bed26fd8251ea8644b5ecd263c0186023 when we started registering all properties during get_person_property etc.

I agree we should take a look at refactoring/cleaning a bunch of this up but it would be good to add some more tests about initialization semantics/indexing before we do that.


RobertJacobsonCDC (approved):
This is really great work! You even tested for the gotcha case of a derived dependency. Nice! 

I can't break it, looks good to me.

PR #325, opened 2025-05-04
chore: Fix benchmark ci

k88hudson-cfa:
This PR addresses an issue where `cargo bench` doesn't run the benchmarks in`ixa-bench` on main right now, so the task is failing


k88hudson:
Argh, the action we're using hasn't been updated so I still can't use the `package` option

PR #326, opened 2025-05-05
New benchmark action

### LINKED ISSUE #328: Pull request benchmark action is failing
The issue is that run `ixa bench` without a package name or `--workspaces` doesn't actually run any benchmarks

k88hudson-cfa:
Fixes #328 

There were some issues with `criterion-compare-action` supporting workspaces (as well as being out-of-date, using a  old versions of node + `checkout`) so I wrote a replacement. The output isn't as nice but it should still be readable.

Needs #327 first


k88hudson-cfa:
itsworking

RobertJacobsonCDC (approved):
Looks good to me. 

We might revisit caching the build artifacts at some point.

PR #341, opened 2025-05-09
feat: Added `ixa-fips` library to the repo and workspace

RobertJacobsonCDC:
It is useful to have this library available for implementing the Settings feature in Ixa core. 

- ~Where does the `fips` crate live? As a member of Ixa's workspace?~ Split crate into `ixa-fips` and the ASPR layer.

ToDo:

- [X] Examples in doc comments `use fips::*` instead of `use ixa_fips::*`. 
- [X] Remove ASPR (to its own crate).
- [X] Improve documentation: clarify relationship to the standard, explain what the codes are and link to the standard where appropriate.
- [x] Add tests to `parser.rs` for codes from real datasets.


ixa-fips/src/aspr/mod.rs
```diff
@@ -0,0 +1,181 @@
...
+
+        // ToDo: Give a reasonable representation for these categories.
+        SettingCategory::Unspecified | SettingCategory::CensusTract => Err(std::fmt::Error),
+    }
+    // The category code and "data" field are not represented in this format.
+    // write!(f, "{:01}", fips_code.category_code())?;
+    // write!(f, "{:03}", fips_code.data())?;
+}
+
+fn format_as_fips_code_string(fips_code: FIPSCode) -> String {
```
k88hudson, line 144: Maybe move this into the test block?

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,276 @@
+/*!
+
+An enum for states as represented by FIPS Geographic Region Codes. Note that the `FIPSCode` encoded type only uses six
```
k88hudson, line 3: I think it would be good to link to the spec here and what version you're using – maybe the 2020 version here https://www.census.gov/library/reference/code-lists/ansi.html?

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,276 @@
...
+
+impl USState {
+    /// Returns true is `self` is a state or District of Columbia
+    pub fn is_state(&self) -> bool {
+        let v = *self as StateCode;
+        v <= 56u8 && ![3u8, 7, 14, 43, 52].contains(&v)
+    }
+
+    /// Returns true if `self` is a state (not including District of Columbia)
+    pub fn is_proper_state(&self) -> bool {
```
k88hudson, line 124: I would probably leave this out unless you have a use case

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,276 @@
...
+            USState::TN => "TN",
+            USState::TX => "TX",
+            USState::UT => "UT",
+            USState::VT => "VT",
+            USState::VA => "VA",
+            USState::WA => "WA",
+            USState::WV => "WV",
+            USState::WI => "WI",
+            USState::WY => "WY",
+            _ => unimplemented!("Only states are supported."),
```
k88hudson, line 208: Why not include other geographic codes here?

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,276 @@
...
+*/
+
+use crate::StateCode;
+use std::fmt::Display;
+
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub enum USState {
+    AL = 1,
+    AK = 2,
+    AmericanSamoa = 3, // (FIPS 5-1 reserved code)
```
k88hudson, line 16: FIPS 5-1 seems to be an outdated version of the spec, these do have two-letter abbreviations you could use

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,276 @@
...
+}
+
+impl Display for USState {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
k88hudson, line 220: I'd probably do some tests against the data provided by the census

k88hudson (changes requested):
Having this as an optional feature of ixa I think would be really nice, although I'd probably only put the FIPS utilities in here and move the APSR stuff to somewhere you can test against the actual dataset. I did a first pass but can do a more thorough review on the APPR code if you move that out.

The situation with which spec is currently active was a bit confusing to be able to review this for accuracy – looks like probably what's described here: https://www.census.gov/library/reference/code-lists/ansi.html, let me know what you were thinking. Let me know and I can do another pass

ixa-fips/src/states.rs
RobertJacobsonCDC: I'll fold this into the tests for `parser.rs`.

ixa-fips/src/aspr/archive.rs
```diff
@@ -0,0 +1,470 @@
...
+pub const CBSA_ALL_DIR: &str = "cbsa_all_work_school_household";
+pub const CBSA_ONLY_RESIDENTS_DIR: &str = "cbsa_only_residents";
+// Either of the next two can be affixed to either of the two above directories
+pub const NON_CBSA_RESIDENTS_DIR: &str = "non_CBSA_residents";
+pub const MULTI_STATE_DIR: &str = "Multi-state";
+
+// Path to the ASPR data directory
+const DEFAULT_ASPR_DATA_PATH: &str = "../CDC/data/ASPR_Synthetic_Population";
+// ToDo: Get the ASPR data path from an environment variable.
+static ASPR_DATA_PATH: Lazy<RwLock<PathBuf>> =
```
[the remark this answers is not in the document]
RobertJacobsonCDC: Marking the ASPR related stuff resolved, but only because we removed ASPR. I'll keep them in my notes.

ixa-fips/src/states.rs
RobertJacobsonCDC: Give Jason's input in the dev meeting, I went the other direction and removed everything but the minimal US States and D.C. We could go a step farther and just use a numeric code for states like we do for other fields, but it's really convenient to interact with these things using the state rather than some number you'd have to know, and this subset of the standard has been very stable. What do you think?

ixa-fips/src/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+/// Parses the first six digits of `input` as a FIPS census tract code. Enforces the
+/// requirement that the value fit into 20 bits (a tautology in this case).
+pub fn parse_tract_code(input: &str) -> FIPSParseResult<u32> {
+    parse_decimal_digits_to_bits(6, 20, input).map(|(rest, value)| {
+        // The `parse_decimal_digits_to_bits` function guarantees `value` fits in 20 bits.
+        (rest, value as u32)
+    })
+}
+
+// #[allow(unused_imports)]
```
k88hudson-cfa, line 173: Remove

ixa-fips/src/parser.rs
```diff
@@ -0,0 +1,374 @@
...
+/// Parses the first two decimal digits of `input` into a `USState` enum variant.
+///
+/// This method is only intended to parse states for which `state.is_state()` is
+/// true. In particular, it enforces the constraint that the decimal value be
+/// representable by 6 binary bits (so values <= 63).
+pub fn parse_state_code(input: &str) -> FIPSParseResult<USState> {
+    parse_decimal_digits_to_bits(2, 6, input).map(|(rest, value)| {
+        // The `parse_decimal_digits_to_bits` function guarantees `value` fits in 6
+        // bits, so this unwrap always succeeds.
+        let state = unsafe { USState::decode(value as u8).unwrap_unchecked() };
```
k88hudson-cfa, line 150: Given that we want to this to work generically, wouldn't it make more sense for this to return a `u8` and have the enum/display versions be something you can use optionally?

k88hudson-cfa (changes requested):

ixa-fips/src/parser.rs
k88hudson-cfa: Also: can you say more about why you went with `transmute`/unsafe for this

RobertJacobsonCDC: I just hadn't pushed the latest commit yet. Both issues resolved.

ixa-fips/src/states.rs
```diff
@@ -0,0 +1,132 @@
...
+}
+
+impl USState {
+    /// Returns true if `self` is a state or District of Columbia
+    pub fn is_state(&self) -> bool {
+        USState::is_state_code(*self as StateCode)
+    }
+
+    /// Returns true if the given state code is a state or District of Columbia
+    pub fn is_state_code(value: StateCode) -> bool {
```
k88hudson-cfa, line 73: Let's change this. + decode to use `FromRepr` since we have strum anyway

k88hudson-cfa (approved):
LGTM with the change to use `FromRepr`

PR #344, opened 2025-05-09
feat: Settings module

RobertJacobsonCDC:
Implements Settings ("groups"). This implementation is a lot simpler than the one in [epi-isolation](https://github.com/CDCgov/ixa-epi-isolation/blob/main/src/settings.rs). 

Some issues for discussion:

- Functions that exist in [[Epi-isolation](https://github.com/CDCgov/ixa-epi-isolation/blob/main/src/settings.rs)](https://github.com/CDCgov/ixa-epi-isolation/blob/main/src/settings.rs) but not implemented here:
  - `ContextSettingExt::calculate_total_infectiousness_multiplier_for_person` is only called in tests. Is this method needed?
  - a public getter for the `alpha`s - only used in tests
  - `get_setting_members` - only used in tests and private implementation
  - `get_contact` for a given person and setting -- only used to sample using itinerary
- Nothing enforces the constraint that $\displaystyle \sum_{ratio \in itinerary}ratio = 1$. We can either enforce the constraint or normalize. Right now we do neither.
- Do we need custom `calculate_multiplier()` functions for each setting category (a.k.a. `SettingType`)? Because all of them are computed using the same formula: $(n-1)^\alpha$ where `n` is the number of people in the setting. Can we just have a single function of $n$ and $\alpha$?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ pedantic = { level = "warn", priority = -1 }
 module-name-repetitions = "allow"
 implicit_hasher = "allow"
 missing_panics_doc = "allow"
+missing_errors_doc = "allow"
 uninlined_format_args = "allow"
 
 [lib]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "ixa-bench",
   "ixa-derive",
   "ixa-integration-tests",
+  "ixa-fips"
 ]
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ license.workspace = true
 edition.workspace = true
 
 [dependencies]
+approx = "^0.5.1"
 rand = "^0.8.5"
 csv = "^1.3.1"
 serde = { version = "^1.0.217", features = ["derive"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,6 +48,7 @@ uuid = "^1.12.1"
 tower-http = { version = "^0.6.2", features = ["full"] }
 mime = "^0.3.17"
 rustc-hash = "^2.1.1"
+fips = { path = "../fips" }
 
 [dev-dependencies]
 rand_distr = "^0.4.3"
--- a/examples/basic-infection/Cargo.toml
+++ b/examples/basic-infection/Cargo.toml
@@ -15,6 +15,9 @@ rand = "^0.8.5"
 rand_distr = "^0.4.3"
 paste = "^1.0.15"
 
+[lints]
+workspace = true
+
 [[bin]]
 name = "basic_infection"
 path = "main.rs"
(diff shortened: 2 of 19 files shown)


RobertJacobsonCDC:
The most important thing to get right is the public API:

```rust
pub trait ContextSettingExt {
    /// Associates an alpha value to a `SettingCategoryCode`. If a value of alpha was already set for
    /// the given category, returns the previous value.
    fn set_alpha_for_setting_category(
        &mut self,
        setting_category: SettingCategoryCode,
        alpha: f64,
    ) -> Option<f64>;

    /// Corresponding getter to `set_alpha_for_setting_category()`.
    fn get_alpha_for_setting_category(
        &mut self,
        setting_category: SettingCategoryCode,
    ) -> Option<f64>;

    /// Sets an `Itinerary` for the given person, inserting the person as a member of the settings
    /// in the given `Itinerary`. Returns the old `Itinerary` if the method replaced an existing
    /// itinerary (i.e. an itinerary was already set for this person), `None` otherwise.
    fn set_itinerary_for_person(
        &mut self,
        person_id: PersonId,
        itinerary: Itinerary,
    ) -> Option<Itinerary>;

    /// Corresponding getter to `set_itinerary_for_person()`.
    fn get_itinerary_for_person(&mut self, person_id: PersonId) -> Option<&Itinerary>;

    /// For the given person, computes the inner product $<R, M>$ where $R$ is the vector of ratios
    /// for each setting and $M$ is the vector of multipliers for each setting.
    ///
    /// Recall that the "multiplier" for a setting is computed as
    ///     `((n_members - 1) as f64).powf(alpha)`.
    fn calculate_total_infectiousness_multiplier_for_person(&self, person_id: PersonId) -> f64;

    /// For a given person, use the person's itinerary and associated setting properties
    /// to sample a setting and a contact from that setting. If the person has no
    /// itinerary, or if the person is isolated (alone) in the setting, returns `None`.
    fn draw_contact_from_itinerary(&self, person_id: PersonId) -> Option<(PersonId, SettingId)>;
}
```

Notice that there are special `Itinerary` and `ItineraryEntry` types. These are convenience types that enforce the constraint that the same setting cannot appear in an itinerary twice.

confunguido:
Some thoughts about your questions. See below. 

> * The public `Context.draw_contact_*` method returns a pair `(PersonId, SettingId)` on success. Should we expose the "draw setting" and "draw contact from setting" methods publicly, or is returning the pair enough?

I think we should expose both methods publicly. If I'm not mistaken, in the most updated version of ixa-epi-isolation, we split this in two: draw setting and draw contact from setting. 

 >`ContextSettingExt::calculate_total_infectiousness_multiplier_for_person` is only called in tests. Is this method needed?

It is needed. It's not only implemented in the tests, it is used in `infectiousness_manager.rs`

>   * `get_setting_members` - only used in tests and private implementation
We will need to get setting members to implement interventions such as contact tracing. For now, this is only required in the tests as an example.

PR #346, opened 2025-05-13
Property Refactor Phase 1: Lookup value moved to property

RobertJacobsonCDC:
- **Looking up a value (`get_person_property`) and computing a value (`PersonProperty::compute`) has been combined into `PersonProperty::compute`**

This changes the semantics of `compute`. The computation of the initial value is now in it's own method `initial_value`.

Both `compute` and `initial_value` return an `Option<T>`, which is consistent with the semantics we agreed on some time ago. However, we just call `unwrap` for the public API until we are ready to make that breaking change. The default behavior of `compute` inserts the initial value into the data store for the person. The whole conversation about changing this behavior to make initialization "lazier" has been sidelined for now. However, because this change moves initialization out of `Context` and into the `PersonProperty`, it does make it _possible_ for client code to implement this "lazier" behavior. 

The justification for this refactor is that the property (or the modeler who defined it) should be responsible for the logic about it's initial value and initialization behavior. Also, a few commits from now registration and dependencies will be moved to `PersonProperty`. (More on that later.) The general principle being served is that `ContextPeopleExt` should define the public API but not private implementation that more logically belongs to other components.

- **`PersonProperty::register_dependencies` renamed to `PersonProperty::dependencies`, because the actual registration occurs in the `Context` extension.**

- **Added the `'static` constraint to the `PersonProperty` trait.** 

A lot of functions require `PersonProperty + 'static`, so anything that implements `PersonProperty` must also be static in practice anyway. We clean up redundent `'static` lifetime constraints in another PR.


src/lib.rs
```diff
@@ -76,3 +76,15 @@ pub use rand;
...
+mod tests {
+    use assert_cmd::cargo::CargoError;
+    pub fn run_external_runner(runner_name: &str) -> Result<assert_cmd::Command, CargoError> {
+        assert_cmd::Command::new("cargo")
+            .args(["build", "--bin", runner_name])
+            .ok()
+            .unwrap_or_else(|_| panic!("Failed to build {runner_name}"));
+        assert_cmd::Command::cargo_bin(runner_name)
+    }
+}
```
RobertJacobsonCDC, line 90: Oops, this shouldn't be here.

RobertJacobsonCDC:
Closed, since we are going to reimplement entities and properties soon.

PR #348, opened 2025-05-19
build: added cache to the github build

cdc-as81:


.github/workflows/build-test.yaml
```diff
@@ -14,6 +14,15 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
+      - uses: actions/cache@v4
```
k88hudson-cfa, line 17: What is your thinking about implementing this way v.s. using something like [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache)? 

Also: I would imagine caching the registry + compiled dependencies  between builds would save most of the time, is that not the case?

k88hudson-cfa (changes requested):

.github/workflows/build-test.yaml
cdc-as81: changing the cache key to a hash of Cargo.toml

k88hudson-cfa (approved):

PR #350, opened 2025-05-27
chore: Members inherit fields from the workspace. Fixes #336.

### LINKED ISSUE #336: What fields of Ixa subcrates should default to the value of the workspace?
What fields of Ixa subcrates should default to the value of the workspace? A nonexhaustive list of possibilities:

```toml
[package]
name = "ixa-subcrate"
description = "An example of an Ixa subcrate"

authors.workspace = true # Currently not defined for the Ixa workspace
version.workspace = true # Should we synchronize versions?
license.workspace = true
edition.workspace = true
reposity.workspace = true
homepage.workspace = true

[lints]
# Each subcrate must opt in to workspace lints
workspace = true
```

The `version` field deserves particular consideration. Should we synchronize the versions of the aubcrates to the workspace?

RobertJacobsonCDC:
Inherit fields from the workspace for the values of the fields repository, license, edition, and homepage from the workspace.

Inheriting dependencies is a separate PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2021"
 repository = "https://github.com/CDCgov/ixa"
 license = "Apache-2.0"
 homepage = "https://github.com/CDCgov/ixa"
+authors = ["The Ixa Developers <cfa@cdc.gov>"]
 
 [package]
 name = "ixa"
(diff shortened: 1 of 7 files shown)


cdc-as81 (approved):
LGTM

PR #372, opened 2025-06-04
Added the `'static` constraint to `PersonProperty`

RobertJacobsonCDC:
We have `PersonProperty + 'static` all over our public API. This PR adds the `'static` constraint to the `PersonProperty` trait itself and replaces all instances of `PersonProperty + 'static` with just `PersonProperty`:

```rust
pub trait PersonProperty: Copy + 'static {
    // ...
}
```


cdc-as81 (approved):
LGTM

PR #373, opened 2025-06-05
chore: Added a more helpful error message for duplicate global property. Fixes #179.

### LINKED ISSUE #179: Confusing error when defining the same global property more than once.
In `ixa::global_properties` we have the function `add_global_property(..)`:

```rust
#[allow(clippy::missing_panics_doc)]
pub fn add_global_property<T: GlobalProperty>(name: &str)
where
    for<'de> <T as GlobalProperty>::Value: serde::Deserialize<'de> + serde::Serialize,
{
    let properties = GLOBAL_PROPERTIES.lock().unwrap();
    assert!(properties
        .borrow_mut()
        .insert(
            // etc.
        )
        .is_none());
}
```

This is called in `define_global_property!` macro expansions. Suppose hypothetically that you have a module with tests that define `define_global_property!(Threshold, u8)`, and in a completely different module you have a different set of tests that also define `define_global_property!(Threshold, u8)`. The tests will compile, but when it runs the code panics with an unhelpful error message.

I suggest we replace the `assert!` with a check that emits an error. This is also a good candidate for log messages once logging lands.

(This issue is unrelated to the fact that Cargo parallelizes tests by default.)

RobertJacobsonCDC:
This is an old one that was never made into a PR for some reason.

### DIFF
--- a/src/global_properties.rs
+++ b/src/global_properties.rs
@@ -59,7 +59,7 @@ where
 {
     trace!("Adding global property {name}");
     let properties = GLOBAL_PROPERTIES.lock().unwrap();
-    assert!(properties
+    properties
         .borrow_mut()
         .insert(
             name.to_string(),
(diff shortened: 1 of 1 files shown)


cdc-as81 (approved):
LGTM

PR #380, opened 2025-06-10
feat: Use a fast pseudorandom number generator

cdc-as81:
replacing StdRng with SmallRng
https://docs.rs/rand/latest/rand/rngs/struct.SmallRng.html#


k88hudson-cfa:
I think @RobertJacobsonCDC is a better reviewer here given his expertise :)

PR #385, opened 2025-06-11
Add PluginContext trait for Context

k88hudson-cfa:
This is an implementation of the first part of a new plugin design based on explorations here https://k88hudson-cfa.github.io/ixa-playground/context-as-trait/index.html based on feedback. I implemented it in a way that should be backwards-compatible, but allows people to migrate.

After this, we should follow up with migrating other context extensions; I migrated `Random` as a sanity check

### DIFF
--- a/src/random.rs
+++ b/src/random.rs
@@ -1,6 +1,6 @@
 use crate::context::Context;
 use crate::hashing::hash_str;
-use crate::{HashMap, HashMapExt};
+use crate::{HashMap, HashMapExt, PluginContext};
 use log::trace;
 use rand::distributions::uniform::{SampleRange, SampleUniform};
 use rand::distributions::WeightedIndex;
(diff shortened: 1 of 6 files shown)


RobertJacobsonCDC (approved):
Looks good. This will really clean things up.

src/random.rs
```diff
@@ -164,18 +114,26 @@ impl ContextRandomExt for Context {
...
     }
 
+    /// Gets a random sample from the random number generator associated with the given
+    /// `RngId` by applying the specified sampler function. If the Rng has not been used
+    /// before, one will be created with the base seed you defined in `set_base_random_seed`.
+    /// Note that this will panic if `set_base_random_seed` was not called yet.
     fn sample<R: RngId + 'static, T>(
         &self,
-        _rng_id: R,
+        _rng_type: R,
```
jasonasher, line 123: Tiny comment (and I know this was merged) - shouldn't this still be _rng_id like the other methods (same for sample_distr)?

PR #413, opened 2025-06-18
fix: person_property debug trait

cdc-as81:


RobertJacobsonCDC (changes requested):
If we unify `__define_person_property_common` with `__define_person_property_option`, we won't need to keep the two macros in sync when the code changes.

RobertJacobsonCDC (approved):
LGTM. 🚀

PR #416, opened 2025-06-18
feat: Added a progress bar feature. Fixes 408.

RobertJacobsonCDC:
There is a tricky bit to make sure the progress bar doesn't interfere with logging.

Questions:
- Should we have a command line flag that enables a progress bar for the timeline?
- I put it behind a feature flag, but we could include it unconditionally and leave it to the client code to enable it in code.
- Should the feature be enabled by default? The two additional dependencies are pretty small.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -1,0 +1,146 @@
+//! Provides functions to set up and update a progress bar.
+//!
+//! A progress bar has a label, a maximum progress value, and its current progress, which
+//! starts at zero. The maximum and current progress values are constrained to be of type
+//! `usize`. However, convenience methods are provided for the common case of a progress bar
+//! for the timeline that take `f64` time values and rounds them to nearest integers for you.
+//!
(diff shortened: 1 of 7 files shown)


src/progress.rs
```diff
@@ -0,0 +1,146 @@
...
+//! use crate::progress_bar::{increment_custom_progress};
+//!
+//! // You might already have this event defined for other purposes.
+//! pub type InfectionStatusEvent = PersonPropertyChangeEvent<InfectionStatus>;
+//!
+//! // This will handle the status change event, updating the progress bar
+//! // if there is a new infection.
+//! fn handle_infection_status_change(context: &mut Context, event: InfectionStatusEvent) {
+//!   // We only increment the progress bar when a new infection occurs.
+//!   if (InfectionStatusValue::Susceptible, InfectionStatusValue::Susceptible)
```
RobertJacobsonCDC, line 65: I think this should be: `if (InfectionStatusValue::Susceptible, InfectionStatusValue::Infected)`.

PR #427, opened 2025-06-27
Added `execution_stats.rs` that implements collecting profiling data.

RobertJacobsonCDC:
Made use of `trace!` and `info!` logging macros more consistent. The execution statistics are printed at simulation end by default if using the runner. The command line argument `--no-stats` disables it.

Questions / Issues:

- Tracking maximum population is tricky. We can use the max of population at start and population at end, or we can track population size dynamically... Do we assume existence of the `People` module (an initilized data container)? 

  **Right now:** if the data container exists, we use `current_population` at simulation end, because this value is monotonically increasing (serves as next `PersonId`). If data container doesn't exist, we just excluding people related stats.

- Polling memory used by current process - at what frequency? It is an expensive system call. **Right now:** 1 Hz.

- The wall and CPU time is *not* paused when the debugger is in use. 

- When using the runner, the stats are printed by default. However, when a `Context` is created, `print_execution_statistics` is false by default. This way stats are not printed when a `Context` is created programmatically, e.g. in tests.

- When `print_execution_statistics` is false, the stats are logged with `info!` instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,9 @@ fern = "^0.7.1"
 wasm-bindgen = "^0.2.100"
 progress_bar = "^1.2.1"
 anyhow = "^1.0.28"
+sysinfo = "^0.35.2"
+humantime = "^2.2.0"
+bytesize = "^2.0.1"
 
 [package]
 name = "ixa"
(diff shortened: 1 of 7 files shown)


RobertJacobsonCDC:
It turns out that `std::time::Instant::now()` is unsupported on wasm32. But we can do this instead:

```rust
use wasm_bindgen::prelude::*;
use web_sys::window;

#[wasm_bindgen]
pub fn get_high_res_time() -> f64 {
    let perf = window().unwrap().performance().unwrap();
    perf.now() // Returns time in milliseconds as f64
}
```

It's a little annoying, because the types are slightly different on wasm32 compared to other platforms, but it's straightforward enough to fix.

k88hudson-cfa (approved):
Personally I would have it off by default but I don't have strong feelings about it

PR #432, opened 2025-07-01
docs: Remove "SIR" terminology from Ixa book

eqmooring:
[The Ixa Book](https://ixa.rs/book) guides readers through a simple transmission model with a constant force of infection. Because this model has susceptible, infected, and recovered compartments, it is referred to as an SIR model. However, I would suggest that we drop this terminology and explain parenthetically how this model is in fact different from the canonical SIR model (i.e., the Kermack-McKendrick model). I believe these changes would help make the text clearer for persons with a prior background infectious disease modeling. This PR is entirely about documentation and does not include any substantive changes to the example model or code.


RobertJacobsonCDC (approved):
Looks good! Thanks!

PR #437, opened 2025-07-02
feat: Adds `filter_people`, which removes people from a vector. Fixes #435.

### LINKED ISSUE #435: Make a `match_people` method to complement `match_person`
There are situations where you have a vector of people in hand and want to filter that vector according to a query. We can do this faster internally than a modeler can using the public API.

RobertJacobsonCDC:
Adds `filter_people`, which removes people from a vector who do not match a given query. This is a multi-person version of `match_person`. It cuts the runtime by 2/3rds relative to calling `match_person` in a loop.

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -86,6 +86,10 @@ pub trait ContextPeopleExt {
     ///
     /// The syntax here is the same as with [`Context::query_people()`].
     fn match_person<T: Query>(&self, person_id: PersonId, q: T) -> bool;
+
+    /// Similar to `match_person`, but more efficient, it removes people
+    /// from a list who do not match the given query.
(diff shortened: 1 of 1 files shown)


KOVALW (commented):
The function itself LGTM! I'm having a hard time seeing why this is a method on context though, and not a `pub fn`.

 Could you add a test like

```rust
    #[test]
    fn test_filter_people() {
        let _ = context
            .add_person((Age, 40))
            .unwrap();
        let _ = context
            .add_person((Age, 42))
            .unwrap();
        let _ = context
            .add_person((Age, 42))
            .unwrap();
        let all_people = context.query_people(());

        assert_eq!((context.filter_people(&mut all_people, (Age, 42)).len(), 2);
        assert!((context.filter_people(&mut all_people, (Age, 43)).is_empty());
```

KOVALW (approved):
LGTM

PR #442, opened 2025-07-07
fix: Silences a few lints new in the latest Rust release. Fixes #440.

### LINKED ISSUE #440: Take care of some lints from new rustc release.
PRs #438 and #427 (and probably others) have lint failures triggered by unrelated code, presumably due to a new rustc release. I have some fixes in #427, but they should be in a separate PR, I think.

RobertJacobsonCDC:
Also, a couple of module members did not correctly inherit the lint exceptions from the workspace.

The new lint exceptions are:
 - `struct_field_names` - Says that field names of structs should now have the name of the struct as a prefix.
 - `unnecessary_debug_formatting` - If we instead use `some_path.display()`, "escaped characters will no longer be escaped and surrounding quotes will be removed." But we prefer escape characters and quotes for our use case.
 - `struct_excessive_bools` - Detects when more than three fields of a struct are of type `bool`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,6 +145,12 @@ missing_panics_doc = "allow"
 missing_errors_doc = "allow"
 uninlined_format_args = "allow"
 result_unit_err = "allow"
+struct_field_names = "allow"
+unnecessary_debug_formatting = "allow"
+struct_excessive_bools = "allow"
+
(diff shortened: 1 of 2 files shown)


cdc-as81 (approved):
LGTM

PR #443, opened 2025-07-07
fix: Unorphands `numeric.rs`, which was accidentally orphaned in #333. Fixes #438.

RobertJacobsonCDC:


RobertJacobsonCDC:
Failing lints are silenced in #442.

Issue #459, opened 2025-07-12
A `justfile` for running tasks locally?


RobertJacobsonCDC:
Do we want a `justfile` to run tasks in the local dev environment? [Just](https://just.systems/man/en/) is a command / task runner similar in spirit to `make` but without the historical baggage or complexity of a build-system. "Tasks" are defined with "recipes", and tasks are composable. A `justfile` for ixa could contain:

- one-time environment setup tasks like installing MDBook, pro-commit, wasm-pack, etc.
- granular build tasks for wasm, examples, etc., and a "build everything" task
- a test task that serves as a single source of truth for what should be tested
- a task to run ALL examples correctly (see #458)
- `clean-target` (i.e. `cargo clean`), `clean-docs`, `clean-book`, `clean-wasm`
- a `ci` task that runs everything that runs in CI but locally (single source of truth)

RobertJacobsonCDC:
[Here's an example of integrating just with pre-commit.](https://github.com/casey/just/blob/master/examples/pre-commit.just)

PR #460, opened 2025-07-12
feat: A `justfile` with a suite of tasks for local development. Implements #459.

RobertJacobsonCDC:
An implementation of a `justfile` described in #459. ~This PR is just for discussion.~ I think we decided we like the idea. We can iterate on the implementation.

List all available recipes with descriptions:

```bash
just --list
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,4 +1,9 @@
 ########################################
+# Set environment variables for Rust builds
+########################################
+export RUSTFLAGS := "-D warnings"
+
+########################################
 # Setup Tasks (one-time developer use)
 ########################################
 


RobertJacobsonCDC:
The improvement to the `basic-infection` benchmark—twice!—is certainly suspicious, isn't it?

.gitignore
```diff
@@ -29,10 +29,15 @@ Cargo.lock
 # html
 examples/**/*.html
 examples/**/*.csv
+# These outputs can end up in the root dir depending on how the example is invoced.
```
RobertJacobsonCDC, line 32: invoked

PR #464, opened 2025-07-14
feat: Implemented global plugin automatic registry. Fixes #403.

### LINKED ISSUE #403: Implement global registration and initializers for data plugins
* Plugins can be registered globally (via ctor or similar) and initialized lazily with a OnceCell.
* Each plugin provides a value that is constructed on demand the first time it is accessed via an initializer.
* Initializers get an immutable context reference

Example usage:
```rust
define_data_plugin!(A, usize, |context| {
    println!("Initializing A");
    *context.get_data(B) + 1
});
```
This will enable plugin modularity without requiring manual registration in every context instance, but without doing extra initialization for plugins that are not used.

Dependency resolution happens automatically as data plugins are accessed.

See https://k88hudson-cfa.github.io/ixa-playground/context-trait-global/index.html

RobertJacobsonCDC:
Tasks remaining before merge:

- [x] Remove `run_with_plugin`, which I think is now unnecessary
- [x] Remove instances of `context.get_data_container_mut()` that exist only to initialize the plugin.

I usually delete redundant Benchmark-Results comments, but here it's useful to compare results across several runs to have confidence that any difference in performance is within the noice threshold.


RobertJacobsonCDC:
For `ctors`:

> This library supports WASM targets, but the MSRV for this target is 1.85

Wasm targets are failing in CI.

src/global_properties.rs
```diff
@@ -115,13 +115,15 @@ macro_rules! define_global_property {
...
 
         $crate::paste::paste! {
-            #[$crate::ctor::ctor]
-            fn [<$global_property:snake _register>]() {
-                let module = module_path!();
-                let mut name = module.split("::").next().unwrap().to_string();
-                name += ".";
-                name += stringify!($global_property);
-                $crate::global_properties::add_global_property::<$global_property>(&name);
+            $crate::ctor::declarative::ctor!{
```
RobertJacobsonCDC, line 118: The `#[ctor]` magic can't be re-exported for some reason, but the "declarative" macro can, so it's necessary to wrap uses of `#[ctor]` in `$crate::ctor::declarative::ctor!{...}`.

RobertJacobsonCDC:
Also test with the web API feature:

```
cargo test --no-default-features --features "web_api logging"
```

k88hudson-cfa (changes requested):
Yay! I think this is mostly good to go, my comments are all minor

k88hudson-cfa (approved):
LGMT

PR #466, opened 2025-07-15
fix: revert sample_person to not call sample_people

cdc-as81:


RobertJacobsonCDC (approved):
Yep, looks good. Nice catch!

PR #469, opened 2025-07-16
fix: Silenced a handful of warnings that are only emitted for the `web_api` feature.

RobertJacobsonCDC:
Also added missing generated docs output to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,9 @@ examples/**/*.csv
 # docs
 website/book/
 website/doc/
+website/debug/
+website/.rustc_info.json
+website/.rustdoc_fingerprint.json
 
 #exceptions:
 !.vscode/settings.json
--- a/src/execution_stats.rs
+++ b/src/execution_stats.rs
@@ -13,6 +13,7 @@ use wasm_bindgen::prelude::*;
 #[cfg(target_arch = "wasm32")]
 use web_sys::window;
 
+#[cfg_attr(target_arch = "wasm32", allow(dead_code))]
 /// How frequently we update the max memory used value.
 const REFRESH_INTERVAL: Duration = Duration::from_secs(1);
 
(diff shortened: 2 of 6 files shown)


RobertJacobsonCDC (commented):
Rebase and you'll find more warnings in `execution_stats.rs`:

```
warning: constant `REFRESH_INTERVAL` is never used
  --> src/execution_stats.rs:17:7
   |
17 | const REFRESH_INTERVAL: Duration = Duration::from_secs(1);
   |       ^^^^^^^^^^^^^^^^

warning: field `last_refresh` is never read
  --> src/execution_stats.rs:52:5
   |
41 | pub struct ExecutionProfilingCollector {
   |            --------------------------- field in this struct
...
52 |     last_refresh: f64,
   |     ^^^^^^^^^^^^

warning: method `poll_memory` is never used
   --> src/execution_stats.rs:111:8
    |
66  | impl ExecutionProfilingCollector {
    | -------------------------------- method in this implementation
...
111 |     fn poll_memory(&mut self) {
    |        ^^^^^^^^^^^
```

EDIT: Appears for the web-pack build specifically.

RobertJacobsonCDC:
> Rebase and you'll find more warnings in `execution_stats.rs`:

What is happening in this case is that we don't poll for the maximum memory used `target_family = "wasm"`, because there is no support for getting memory used by the current process on that platform. So I conditionally compile the `poll_memory` method for non-wasm targets. BUT I want to keep `ExecutionProfilingCollector::last_refresh` and `execution_stats::REFRESH_INTERVAL` for potential future use, as they are not intrinsically related to unsupported features in the same way, so in those cases I just `allow(dead_code)`. I think that will be the least confusing for future me.

In any case, these warnings are unrelated to the `web_api` feature.

PR #477, opened 2025-07-26
feat: Changed `Context::data_plugins` into a `Vec`, implemented `DataPlugin::index` method.

### LINKED ISSUE #224: Use fast data structures for small sets
In some cases, a simple vector of items is much faster than a `HashMap` or `HashSet`. In general, this will be the case for 

- Small sets (hash time)
- Many sets (creation time)

This task is to identify which uses of `HashMap` / `HashSet` are
 - always small
 - in the hot path, and
 - represent a nonnegligible amount of the running time.

For the cases identified, replace them with a faster alternative. (Jason Asher has done some benchmarking to determine the fastest alternative in his `rust-nucleus` experiment.)

RobertJacobsonCDC:
This partly addresses #224, but I believe the same pattern can be applied elsewhere in the codebase, so we shouldn't close #224.

Each implementer of `DataPlugin` knows the index into `Context::data_plugins` where an instance of itself should be stored. This index is initialized lazily upon first access in a thread-safe way.

### DIFF
--- a/src/data_plugin.rs
+++ b/src/data_plugin.rs
@@ -50,6 +50,8 @@ pub fn initialize_data_plugin_index(plugin_index: &AtomicUsize) -> usize {
     // another instance of this plugin in another thread just initialized the index prior to us
     // obtaining the lock. If the index has been initialized beneath us, we do not update
     // `NEXT_DATA_PLUGIN_INDEX`, we just return the value `plugin_index` was initialized to.
+    // For a justification of the data ordering, see:
+    //     https://github.com/CDCgov/ixa/pull/477#discussion_r2244302872
     match plugin_index.compare_exchange(usize::MAX, candidate, Ordering::AcqRel, Ordering::Acquire)
     {
         Ok(_) => {
@@ -134,7 +136,10 @@ macro_rules! define_data_plugin {
 mod tests {
     use super::*;
     use crate::Context;
+    use std::sync::{Arc, Barrier};
+    use std::thread;
 
+    // We attempt an out-of-bounds index with a plugin
     #[test]
     #[should_panic(
         expected = "No data plugin found with index = 1000. You must use the `define_data_plugin!` macro to create a data plugin."
@@ -161,7 +166,7 @@ mod tests {
         println!("{}", container.len());
     }
 
-    // We attempt a collision with this plugin
+    // We attempt a collision with a plugin
     define_data_plugin!(LegitDataPlugin, Vec<u32>, vec![]);
     #[should_panic(
         expected = "TypeID does not match data plugin type. You must use the `define_data_plugin!` macro to create a data plugin."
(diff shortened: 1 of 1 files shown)


src/data_plugin.rs
```diff
@@ -24,11 +26,53 @@ pub fn get_plugin_ids() -> Vec<TypeId> {
...
+pub fn initialize_data_plugin_index(plugin_index: &AtomicUsize) -> usize {
+    // Acquire a global lock.
+    let mut guard = NEXT_DATA_PLUGIN_INDEX.lock().unwrap();
+    let candidate = *guard;
+
+    // Try to claim the candidate index. Here we guard against the potential race condition that
+    // another instance of this plugin in another thread just initialized the index prior to us
+    // obtaining the lock. If the index has been initialized beneath us, we do not update
+    // `NEXT_DATA_PLUGIN_INDEX`, we just return the value `plugin_index` was initialized to.
+    match plugin_index.compare_exchange(usize::MAX, candidate, Ordering::SeqCst, Ordering::SeqCst) {
```
k88hudson-cfa, line 53: Does failure really need to be `Ordering::SeqCst`? If it's just loading the value it can probably be `Relaxed` right?

src/data_plugin.rs
```diff
@@ -43,13 +87,30 @@ macro_rules! __define_data_plugin {
             fn init($ctx: &impl $crate::PluginContext) -> Self::DataContainer {
                 $body
             }
+
+            fn index() -> usize {
+                // This static must be initialized with a compile-time constant expression.
+                // We use `usize::MAX` as a sentinenl to mean "uninitialized". This
```
k88hudson-cfa, line 93:
```suggestion
                // We use `usize::MAX` as a sentinel to mean "uninitialized". This
```

src/data_plugin.rs
```diff
@@ -24,11 +26,53 @@ pub fn get_plugin_ids() -> Vec<TypeId> {
...
+
 /// A trait for objects that can provide data containers to be held by `Context`
 pub trait DataPlugin: Any {
     type DataContainer;
 
     fn init(context: &impl PluginContext) -> Self::DataContainer;
+
+    /// Returns the index into `Context::data_plugins`, the vector of data plugins, where
+    /// the instance of this data plugin can be found.
+    fn index() -> usize;
```
k88hudson-cfa, line 75: Personally I would call this `type_index` or something less general than `index`

src/data_plugin.rs
```diff
@@ -24,11 +26,53 @@ pub fn get_plugin_ids() -> Vec<TypeId> {
...
+    // Acquire a global lock.
+    let mut guard = NEXT_DATA_PLUGIN_INDEX.lock().unwrap();
+    let candidate = *guard;
+
+    // Try to claim the candidate index. Here we guard against the potential race condition that
+    // another instance of this plugin in another thread just initialized the index prior to us
+    // obtaining the lock. If the index has been initialized beneath us, we do not update
+    // `NEXT_DATA_PLUGIN_INDEX`, we just return the value `plugin_index` was initialized to.
+    match plugin_index.compare_exchange(usize::MAX, candidate, Ordering::SeqCst, Ordering::SeqCst) {
+        Ok(_) => {
```
k88hudson-cfa, line 54: What do you think about checking the TypeId for registration here / giving people a good error message about using the macro? Right now if you don't register the failure case is that you eventually hit an index that doesn't exist, which seems less good

k88hudson-cfa (changes requested):
Looks good, I think the only thing that seems weird to me is the failure state for if you impl the trait without registering

src/data_plugin.rs
RobertJacobsonCDC: Let me experiment with how it fails and get back to you. I was thinking that pretty much all error cases end up occurring here:

```rust
thing.downcast_mut::<T::DataContainer>()
     .expect("TypeID does not match data plugin type")
```

But if you get a weird case where client code doesn't use the macro or something, you might also get an index out of bounds error, I guess.

src/data_plugin.rs
RobertJacobsonCDC: Yeah, I debated with myself on this, too. But I don't think `type_index` is better, and it does have the context of being a trait method of `DataPlugin`.  How about `index_within_context`?

src/data_plugin.rs
RobertJacobsonCDC: In any case, we can probably improve that error message.

k88hudson-cfa: Via discussion offline – I was thinking of the one weird case here (hitting max entries) but @RobertJacobsonCDC  pointed out some other stuff that could be weird if you don't use the macro (hard coding `index` or something). so yeah I agree, let's just live with it

src/data_plugin.rs
```diff
@@ -67,3 +129,69 @@ macro_rules! define_data_plugin {
         $crate::__define_data_plugin!($data_plugin, $data_container, |_context| $default);
     };
 }
+
+#[cfg(test)]
```
k88hudson-cfa, line 133: Personally I would put a put a test in here that tests the regular case, i.e., declaring a bunch of plugins and accessing them across various threads

k88hudson-cfa (approved):
Thanks, I would add a couple more test cases before merging but LGTM

PR #489, opened 2025-08-07
feat: Publicly exposed `Context::get_execution_statistics`. Fixes #488.

### LINKED ISSUE #488: Publicly expose `ExecutionStatistics`
Make a public method `Context::get_execution_stats` for consumption by the `profiling` plugin. Maybe we eventually offload printing execution stats to a plugin as well, but this is enough for now.

RobertJacobsonCDC:
This PR adds `Context::get_execution_statistics(&mut self) -> ExecutionStatistics` and also a method on `PluginContext` that delegates to it. The purpose is to expose the collected statistics for consumption by profiling modules like the one proposed for epi-isolation. 

**Future Direction**
If/when the profiling module is ported back to ixa core, displaying or logging a context's`ExecutionStatistics` can be moved into that module. (Right now it's a free function in `execution_stats.rs`.) We can convert the execution statistics collector into a data plugin in which the mechanism used to refresh the execution stats is a periodic plan.

### DIFF
--- a/examples/basic-infection/src/incidence_report.rs
+++ b/examples/basic-infection/src/incidence_report.rs
@@ -2,10 +2,10 @@ use crate::infection_manager::InfectionStatusEvent;
 use crate::people::InfectionStatusValue;
 use ixa::prelude::*;
 use ixa::{trace, PersonId};
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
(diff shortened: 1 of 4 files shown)


k88hudson-cfa (approved):

k88hudson-cfa:
ty!

PR #492, opened 2025-08-11
docs: Added a chapter on Reports to the Ixa Book.

RobertJacobsonCDC:
This also changes the `MAX_TIME` of `basic-infection` from the arbitrary value `303` to the arbitrary value `300` for aesthetic reasons.


k88hudson-cfa (approved):
💅

PR #510, opened 2025-08-28
fix: Derived properties with Global Property dependencies. Fixes #475.

### LINKED ISSUE #475: Derived properties with Global Property dependencies
A derived property depending on a global property fetches a reference to the global property value, dereferences the property value, and passes the value to the closure defining the derived property. This is a problem in the case that the global property value is not `Copy`. Right now global properties are immutable, so it probably makes more sense to just pass an immutable reference to the global property value.

It looks like making this changed doesn't have catostrphic consequences—everything compiles and the code seems to work. But it breaks a bunch of tests and examples. Also, someone with a better understanding of the history of this code should check my assumptions, here.

cdc-as81:
…ed rust lint that does not exist

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -663,7 +663,7 @@ mod tests {
     }
     define_global_property!(ThresholdP, u8);
     define_derived_property!(IsEligible, bool, [Age], [ThresholdP], |age, threshold| {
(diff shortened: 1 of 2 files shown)


RobertJacobsonCDC (approved):
🚢-it.

PR #515, opened 2025-08-29
fix: remove wasm-test from just prepush, updated dev codespace install

cdc-as81:


justfile
```diff
@@ -290,5 +290,5 @@ clean: clean-target clean-wasm clean-docs clean-book clean-examples
 # environment, say, before pushing.
 
 # Run locally everything that runs in CI
-prepush $RUSTFLAGS="-D warnings": precommit build-all-targets build-wasm-pack test test-examples test-wasm \
```
RobertJacobsonCDC, line 293: Could we instead have another variant of prepush for the CodeSpaces environment?

RobertJacobsonCDC (changes requested):

RobertJacobsonCDC (approved):

PR #517, opened 2025-09-02
feat: Use bump allocator for allocations during querying

RobertJacobsonCDC:
Allocations are made during querying that are difficult to avoid without major changes. We can speed things up quite a bit by using a bump allocator during querying that we then reset.

**EDIT:** The benchmarks from github-actions suggest a much more modest improvement than my own benchmark. Maybe try it on your machine(s) to see if the added complexity is worth it.

```bash
just baseline-create bump
just baseline-compare bump
```

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -9,10 +9,11 @@ use crate::{
     PersonPropertyChangeEvent, RngId, Tabulator,
 };
 use crate::{HashMap, HashMapExt, HashSet, HashSetExt};
+use bumpalo::{collections::Vec as BumpVec, Bump};
 use log::warn;
 use rand::Rng;
 use std::any::TypeId;
-use std::cell::Ref;
+use std::cell::{Ref, RefMut};
 
 /// A trait extension for [`Context`] that exposes the people
 /// functionality.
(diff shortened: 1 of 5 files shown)


src/people/context_extension.rs
```diff
@@ -555,89 +576,92 @@ impl ContextPeopleExtInternal for Context {
...
+            //    (2) the overall population if there are no indices.
 
-        let holder: Ref<HashSet<PersonId>>;
-        let to_check: Box<dyn Iterator<Item = PersonId>> = if indexes.is_empty() {
-            Box::new(data_container.people_iterator())
-        } else {
-            indexes.sort_by_key(|x| x.len());
+            let holder: Ref<HashSet<PersonId>>;
+            let to_check: Box<dyn Iterator<Item = PersonId>> = if indexes.is_empty() {
+                Box::new(data_container.people_iterator())
```
RobertJacobsonCDC, line 632: I couldn't figure out how to allocate these two instances of `Box` using `bumpalo::boxed::Box`. The compiler complained about a missing `Sized` constraint.

RobertJacobsonCDC:
The performance gains are not as much as they first appeared. We can revisit the idea in the future, but for now it's not worth the complexity.

PR #518, opened 2025-09-02
feat: Multi-Properties and Multi-Indexing Refactor

RobertJacobsonCDC:
This PR implements "multi-properties," which are `PersonProperty` types that are just tuples of existing `PersonProperty` types. They can be indexed and queried just like any other `PersonProperty`. The idea is that this concept will replace multi-indices.

The `PersonProperty` trait is endowed with a `PersonProperty::Value` and `PersonProperty::CanonicalValue` associated types and the `PersonProperty::make_canonical()` method and its inverse `PersonProperty::make_uncanonical()`, for implementors that store a different value in its index than its natural value. Multi-properties use these types and methods to store their values in component-sorted order so that multi-properties with the same component properties in a different order can use the same index. Also, queries that query for the same property set as a multi-property can also use the multi-property index.


RobertJacobsonCDC:
Benchmark results seem to have a high variance, but average times seem to be on par. The `bench_query_population_indexed_property` benchmark is consistently a little faster.

src/people/query.rs
```diff
@@ -54,83 +111,69 @@ macro_rules! impl_query {
...
-/// define_person_property!(Age, u8);
-/// define_person_property!(Alive, bool);
-/// let context = Context::new();
-/// let q1 = (Age, 42);
-/// let q2 = (Alive, true);
-/// context.query_people(QueryAnd::new(q1, q2));
-/// ```
-
-#[derive(Copy, Clone)]
-pub struct QueryAnd<Q1, Q2>
```
RobertJacobsonCDC, line 87: We never use `QueryAnd`, and I don't think it's worth the maintenance cost. We can add if back if and when we have a use case for it.

ixa-bench/large_dataset/large_dataset.rs
```diff
@@ -49,7 +50,7 @@ fn bench_query_population_derived_property(context: &mut Context) {
 }
 
 pub fn criterion_benchmark(c: &mut Criterion) {
-    define_multi_property_index!(Age, SchoolId, WorkplaceId);
+    define_multi_property!(ASW, (Age, SchoolId, WorkplaceId));
```
k88hudson-cfa, line 53: If I'm reading this correctly, this test (and any user code) now needs to call `.add_index(AWS)` now right? That seems like a regression in the user experience. 

This kind of brings up a broader point which is, if the only use case for a multi-property is to help with indexing, (1) why expose the multi-property name in the public API at all (2) don't you want to create the index automatically?

ixa-derive/src/reorder_fn.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    quote! {
+        fn unreorder_by_tag((#( #vars ),*): (#( #value_types_sorted ),*)) -> (#( #value_types_unsorted ),*) {
+            ( #( #unreordered ),* )
+        }
+    }
+}
+
+/// Expands to an `impl $struct_name` block that defines make_canonical and make_uncanonical methods.
+pub fn impl_reoder_fns(input: TokenStream) -> TokenStream {
```
k88hudson-cfa, line 81: Typo

k88hudson-cfa (changes requested):

src/prelude.rs
```diff
@@ -3,7 +3,6 @@ pub use crate::define_data_plugin;
 pub use crate::define_derived_property;
 pub use crate::define_edge_type;
 pub use crate::define_global_property;
-pub use crate::define_multi_property_index;
```
RobertJacobsonCDC, line 6: Forgot to *add* `define_multi_property` to prelude.

Also, surely we meant `PersonId` to be in the prelude... it doesn't seem to be there?

ixa-bench/large_dataset/large_dataset.rs
RobertJacobsonCDC: I'm trying to remember why this is here. Either we had it, then decided to change the tests around to not use it and forgot to remove it, or else we just needed to put a usage of `define_multi_property_index` in an external crate somewhere as a test.

In any case:
1. it isn't used in `large_dataset.rs` or related code,
2. but we do need a usage external to `ixa`,
3. we now put things that need to be tested in an external crate in an integration test, because that makes far more sense than random pieces of code in a benchmark. 🤦‍♂️

PR #522, opened 2025-09-03
feat: Added a `-v` / `--verbose` commandline argument. Fixes #494.

### LINKED ISSUE #494: Improve log filtering ergonomics.
We've found that filtering log messages is cumbersome in practice. Research a better way to do this.

In epi-isolation some people run

```bash
cargo run -- --config input/input.json --log-level=Trace -f | grep epi_isolation
```

instead of

```bash
cargo run -- --config input/input.json --log-level epi_isolation=Trace -f
```

which might just be vestigial or maybe a lack of documentation for that command flag.

RobertJacobsonCDC:
Recall level order: `error!`, `warn!`, `info!`, `debug!` and `trace!`

Issues for discussion:

- Change default to `LogLevel::Error` instead of `LogLevel::Off`.
- Downgrade "Tried to cancel nonexistent plan" to a warning, as there are use-cases where it is convenient to do. Discuss whether we want to turn it off or lower it to `debug`.

So we have:

| Option   | Log Levels Enabled                   |
|----------|--------------------------------------|
| *(none)* | errors only                          |
| `-v`     | errors, warnings, info               |
| `-vv`    | errors, warnings, info, debug        |
| `-vvv`   | errors, warnings, info, debug, trace |

### DIFF
--- a/src/runner.rs
+++ b/src/runner.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
 use crate::context::Context;
@@ -6,13 +6,14 @@ use crate::context::Context;
 use crate::debugger::enter_debugger;
 use crate::error::IxaError;
 use crate::global_properties::ContextGlobalPropertiesExt;
+use crate::log::level_to_string_list;
 #[cfg(feature = "progress_bar")]
 use crate::progress::init_timeline_progress_bar;
 use crate::random::ContextRandomExt;
 use crate::report::ContextReportExt;
 #[cfg(feature = "web_api")]
 use crate::web_api::ContextWebApiExt;
-use crate::{info, set_log_level, set_module_filters, warn, LevelFilter};
+use crate::{set_log_level, set_module_filters, warn, LevelFilter};
 use clap::{ArgAction, Args, Command, FromArgMatches as _};
 
 /// Custom parser for log levels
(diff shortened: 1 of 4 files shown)


src/runner.rs
```diff
@@ -60,6 +60,10 @@ pub struct BaseArgs {
     #[arg(short, long)]
     pub log_level: Option<String>,
 
+    /// Increase logging verbosity (-v, -vv, -vvv, etc.)
```
k88hudson-cfa, line 63: I'd be way more explicit in the help text here about each level

k88hudson-cfa (commented):
> Change default to LogLevel::Error instead of LogLevel::Off

Definitely agree with this 

Re: the UX, I like it a lot better! Maybe we could print out something like this in the help text?

```
+---------+-------+------+-------+-------+-------+
| Level   | ERROR | WARN | INFO  | DEBUG | TRACE |
+---------+-------+------+-------+-------+-------+
| Default |   ✓   |      |       |       |       |
| -v      |   ✓   |  ✓   |   ✓   |       |       |
| -vv     |   ✓   |  ✓   |   ✓   |   ✓   |       |
| -vvv    |   ✓   |  ✓   |   ✓   |   ✓   |   ✓   |
+---------+-------+------+-------+-------+-------+
```
And something like this at the start of logging

```
Current log levels enabled: ERROR, WARN
Run [model] -- --help -v to see more options
```

In terms of other options, I think the only other option I can think of is

```
--warn
--debug
--trace
```
etc.

RobertJacobsonCDC:
In the presence of multiple log level options, we take the *maximum* log level supplied. This is only relevant for the *global* log level, not any per-module log level/filters that might be set (with `--log-level`). So `-v --trace` would mean, "set the global log level to `TRACE`."

PR #538, opened 2025-09-15
fix: Visibility of multi-property support functions used in a public macro was private.

RobertJacobsonCDC:
The functions `add_multi_property_index` and `get_multi_property_value_hash`, which are used in the `define_multi_property_index!` macro, should have been publicly exported in `ixa::people`. This caused a compilation error when building the benchmarks in `large_dataset.rs`. For some reason this was not caught in CI for the original PR but caused CI to fail for subsequent PRs.


cdc-as81 (approved):
LGTM

PR #543, opened 2025-09-16
feat: Extend PluginContext to include base plugins

k88hudson-cfa:
Fixes #542

PR #544, opened 2025-09-16
feat: Add reference model SIR benchmarks

k88hudson-cfa:
Note, this blocked because it is stacked on #543 

Adds SIR benches as criterion benchmarks. You can test this with:

```sh
just bench large_sir
```

Honestly I really prefer the hyperfine output but this seems to work ok; I tried messing around with the criterion settings, but I still see weird noise between runs.


RobertJacobsonCDC (approved):
Works for me!

ixa-bench/src/bench_utils/runner.rs
```diff
@@ -0,0 +1,63 @@
...
+}
+
+#[derive(Parser)]
+#[command(version, about, long_about = None)]
+struct BenchRunnerArgs {
+    #[command(subcommand)]
+    command: BenchRunnerCommand,
+}
+
+pub struct BenchRunner;
```
RobertJacobsonCDC, line 29: Why not just have free functions? Is the idea that we might want to hold state in the future?

justfile
```diff
@@ -165,8 +171,44 @@ test:
...
+# Run a single hyperfine benchmark
+_hyperfine group="" *hyperfine_opts="--warmup 3 --runs 50":
+    @bench_list=$(./target/release/hyperfine list --group {{group}} --one-line) && \
+    hyperfine -N {{hyperfine_opts}} \
+    --parameter-list bench_id $bench_list \
+    --command-name '{{group}}::{bench_id}' \
+    './target/release/hyperfine run --group {{group}} --bench {bench_id}'
+
+# Run a single run of all benchmarks.For CI
+hyperfine-check:
```
RobertJacobsonCDC, line 210: We probably should add it to the GitHub workflow, too.

ixa-bench/src/reference_sir/bench.rs
```diff
@@ -0,0 +1,25 @@
+use super::*;
```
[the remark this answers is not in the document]
RobertJacobsonCDC: I really like it. It'll be nice to have.

PR #548, opened 2025-09-17
feat: Optimize `sample_people` & `sample_person`. Fixes #534.

### LINKED ISSUE #534: Optimize sample people
Getting a random person given some properties is also a very hot path (e.g., for selecting contacts):

Note that we do not need to return the entire set of IDs, we just need to find a matching person. Definitely we should avoid copying any stuff out of the index if it exists. 
 
Sometimes it is better to select any person and reject until a match. Maybe we could give people an easy way to switch between strategies?

RobertJacobsonCDC:
- Moved `proc-macro2` and `proc-macro-crate` to workspace dependencies.
- Added `algorithm_benches.rs` to comparing the performance of different sampling algorithms.
- Added "alternative_algorithm_benches" feature flag to benches to turn off benchmarking unused algorithms. They are convenient to have for comparison with the implemented versions, but we don't use them in ixa.
- Added fast paths in `sample_person()` / `sample_people()` for the indexed case using "known length" algorithm.

This PR will need to be rebased when #544 lands.


ixa-bench/algorithms/algorithm_benches.rs
RobertJacobsonCDC, line 1: Oops, I forgot to copy over the benchmarks for the ixa implementation itself.

ixa-bench/algorithms/algorithm_benches.rs
```diff
@@ -0,0 +1,205 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use ixa::rand::{seq::index::sample as choose_range, seq::IteratorRandom, thread_rng, Rng};
+use std::hint::black_box;
+
+/// A wrapper around any iterator that prevents it from
+/// exposing `ExactSizeIterator`, `DoubleEndedIterator`, etc.
```
RobertJacobsonCDC, line 6: Say why this is needed.

ixa-bench/algorithms/algorithm_benches.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+
+pub fn criterion_benchmark(criterion: &mut Criterion) {
+    let mut rng = thread_rng();
+    // The number of items to choose out of the data set for multiple sampling
+    let mut counts: Vec<usize> = Vec::with_capacity(1000);
+    for _ in 0..1000 {
+        counts.push(rng.gen_range(5..100));
```
RobertJacobsonCDC, line 35: Is this a reasonable range for the number of samples we will want to take out of a query result set?

ixa-bench/algorithms/algorithm_benches.rs
```diff
@@ -0,0 +1,205 @@
...
+    // This algorithm is significantly slower than the "known length" algorithm (factor
+    // of 10^4). The reservoir algorithm from `rand` reduces to the "known length`
+    // algorithm when the iterator is an `ExactSizeIterator`, or more precisely,
+    // when `iterator.size_hint()` returns `(k, Some(k))` for some `k`. Otherwise,
+    // this algorithm is much faster than the `rand` implementation (factor of 100).
+    criterion.bench_function("algorithm_sampling_single_l_reservoir", |bencher| {
+        bencher.iter(|| {
+            // Treat inputs as opaque at the start of the iteration
+            let rng = black_box(&mut rng);
+            let data = black_box(&data);
```
RobertJacobsonCDC, line 68: Maybe move these to the top of the function instead of including them in each closure so that you know every closure is using the same back-boxed input? Check `black_box` docs.

src/people/context_extension.rs
```diff
@@ -379,45 +379,96 @@ impl ContextPeopleExt for Context {
...
+
+        // This is the "unknown length" case. This algorithm is *much*
+        // faster than the reservoir algorithm implemented in `rand`.
+        // This implements "Algorithm L" from KIM-HUNG LI, Reservoir-
+        // Sampling Algorithms of Time Complexity O(n(1 + log(N/n)))
         // https://dl.acm.org/doi/pdf/10.1145/198429.198435
-        // Temporary variables.
         let mut w: f64 = self.sample_range(rng_id, 0.0..1.0);
         let mut ctr: usize = 0;
         let mut i: usize = 1;
```
RobertJacobsonCDC, line 468: Use comprehensible variable names like you did in `algorithm_benches.rs`.

src/people/context_extension.rs
```diff
@@ -444,36 +496,61 @@ impl ContextPeopleExt for Context {
...
+        // This is the "unknown length" case. This algorithm is *much*
+        // faster than the reservoir algorithm implemented in `rand`.
+        // This implements "Algorithm L" from KIM-HUNG LI, Reservoir-
+        // Sampling Algorithms of Time Complexity O(n(1 + log(N/n)))
         // https://dl.acm.org/doi/pdf/10.1145/198429.198435
-        // Temporary variables.
         let mut selected: Option<PersonId> = None;
         let mut w: f64 = self.sample_range(rng_id, 0.0..1.0);
         let mut ctr: usize = 0;
         let mut i: usize = 1;
```
RobertJacobsonCDC, line 545: Same with variable names.

RobertJacobsonCDC:
Split benchmarks into their own commit (the first commit), the optimizations into a second commit. 

Checkout the first commit, run

```bash
just bench sample_people
```

then checkout the second commit, and run it again, and you should get something like this:

Screenshot 2025-09-19 at 1 17 33 PM

The algorithm benches are independent of the second commit, but you can run them with:

 ```bash
just bench algorithm_sampling
```

File: ixa-bench/criterion/algorithm_benches.rs [hunk 6]
```diff
@@ -0,0 +1,205 @@
...
+        counts.push(rng.gen_range(5..100));
+    }
+    // A data set of numbers to simulate sampling from a population (memory access patterns)
+    let mut data: Vec<u8> = Vec::with_capacity(100_000);
+    for _ in 0..100_000 {
+        data.push(rng.gen_range(0..100));
+    }
+
+    // This algorithm is used when the property is indexed, and thus we know the length of the result set.
+    criterion.bench_function("algorithm_sampling_single_known_length", |bencher| {
```
k88hudson-cfa, line 44: Some thoughts here:

- We talked a little about whether it is worth it to leave these in – I do think it probably is, although I would implement these as helper functions in the random module and adjust the tests to reference those. So the goal of this would be to catch major changes in `rand` etc. 
- What's your source for the`u32` for `usize`?

ixa-bench/criterion/algorithm_benches.rs
```diff
@@ -0,0 +1,205 @@
...
+        bencher.iter(|| {
+            // Treat inputs as opaque at the start of the iteration
+            let rng = black_box(&mut rng);
+            let data = black_box(&data);
+
+            // This little trick with `u32` makes this function 30% faster.
```
k88hudson-cfa, line 50: Interesting... source? I wonder if this will change when we upgrade to 0.9 at all?

src/people/context_extension.rs
```diff
@@ -379,61 +379,125 @@ impl ContextPeopleExt for Context {
...
-        // Temporary variables.
-        let mut w: f64 = self.sample_range(rng_id, 0.0..1.0);
-        let mut ctr: usize = 0;
-        let mut i: usize = 1;
+        let mut weight: f64 = self.sample_range(rng_id, 0.0..1.0);
+        let mut position: usize = 0;
+        let mut next_pick_position: usize = 1;
         let mut selected = Vec::new();
 
+        // ToDo(Robert): This will use `iter_query_results` API when it is ready.
```
k88hudson-cfa, line 482:
```suggestion
        // TODO(RobertJacobsonCDC): This will use `iter_query_results` API when it is ready.
```

src/people/context_extension.rs
```diff
@@ -444,40 +508,70 @@ impl ContextPeopleExt for Context {
...
+            if let Some(index) = container
+                .property_indexes
+                .borrow_mut()
+                .get_mut(&multi_property_id)
+            {
+                // Make sure the index isn't stale.
+                index.index_unindexed_people(self);
+
+                if let Some(people_set) = index.get_with_hash(query.multi_property_value_hash()) {
+                    // This little trick with `u32` makes this branch 30% faster.
```
k88hudson-cfa, line 543: See my note in the benchmarks about refactoring these into the random module

Cargo.toml
```diff
@@ -51,7 +51,9 @@ roots = "0.0.8"
 assert_approx_eq = "^1.1.0"
 strum = { version = "^0.27.1", features = ["derive"] }
 quote = "^1.0.38"
-syn = "^2.0.95"
+syn = { version = "^2.0.95",  default-features = false, features = ["derive", "parsing"] }
```
k88hudson-cfa, line 54: What's up with the changes here?

ixa-bench/criterion/algorithm_benches.rs
```diff
@@ -0,0 +1,205 @@
+use criterion::{criterion_group, criterion_main, Criterion};
+use ixa::rand::{seq::index::sample as choose_range, seq::IteratorRandom, thread_rng, Rng};
+use std::hint::black_box;
+
+/// A wrapper around any iterator that prevents it from exposing `ExactSizeIterator`, `DoubleEndedIterator`, etc.
+/// This is needed to test the "slow path" with `rand`'s implementation, as `rand` has an optimization for when
+/// `iterator.size_hint()` returns `(k, Some(k))`.
```
k88hudson-cfa, line 7:
```suggestion
// A wrapper around any iterator that prevents it from exposing `ExactSizeIterator`, `DoubleEndedIterator`, etc.
// This is needed to test the "slow path" with `rand`'s implementation, as `rand` has an optimization for when
// `iterator.size_hint()` returns `(k, Some(k))`.
```

ixa-bench/criterion/sample_people.rs
```diff
@@ -0,0 +1,98 @@
...
+    context.sample_range(SampleBenchRng, 0..10)
+});
+define_person_property!(Property100, u8, |context: &Context, _person: PersonId| {
+    context.sample_range(SampleBenchRng, 0..100)
+});
+
+pub fn criterion_benchmark(criterion: &mut Criterion) {
+    let mut criterion = criterion.benchmark_group("sample_people");
+    let mut rng = thread_rng();
+    let mut context = Context::new();
```
k88hudson-cfa, line 19: I would move this + the to a setup function to isolate each test (you could also do the `blackbox` part) 

```rust
criterion.bench_function("sampling_multiple_known_length", |bencher| {
    let (context, counts) = setup();
    bencher.iter(|| {
        let counts = black_box(&counts);

        for value in counts {
            ...
        }
    });
});
```

src/people/context_extension.rs
```diff
@@ -379,61 +379,125 @@ impl ContextPeopleExt for Context {
...
+                }
+                Some(person) => {
+                    vec![person]
+                }
+            };
+        }
+
+        let current_population = self.get_current_population();
+
+        let requested = std::cmp::min(n, current_population);
```
k88hudson-cfa, line 404: This is a good change, although I do think it's really odd that `sample_people` returns a `Vec` when `sample_person` returns an `Option` – maybe we should do a breaking change here, (or create a new method and deprecate?)

k88hudson-cfa (changes requested):
Thanks, this is great work!

I imagine this might be good to revisit after upgrading to 0.9, but did you consider/compare against https://github.com/rust-random/rand/blob/54ea272a34c73067f315fbbbced558a7e8ba5a8c/src/seq/iterator.rs#L66 at all?

ixa-bench/criterion/algorithm_benches.rs
RobertJacobsonCDC: Good question! I don't know.

I discovered it when the `algorithm_sampling_single_rand_reservoir` benchmark kept beating mine by ~30%. Digging into the implementation of [`rand`'s implementation of `IteratorRandom::choose`](https://docs.rs/rand/latest/src/rand/seq/iterator.rs.html#66-119), I saw that they have a fast path that reduces to my implementation, but it was still faster somehow. It turns out they are using an internal function called [`gen_index`](https://docs.rs/rand/0.8.5/src/rand/seq/mod.rs.html#659) that is implemented like this:

```rust
// Sample a number uniformly between 0 and `ubound`. Uses 32-bit sampling where
// possible, primarily in order to produce the same output on 32-bit and 64-bit
// platforms.
#[inline]
fn gen_index<R: Rng + ?Sized>(rng: &mut R, ubound: usize) -> usize {
    if ubound <= (core::u32::MAX as usize) {
        rng.gen_range(0..ubound as u32) as usize
    } else {
        rng.gen_range(0..ubound)
    }
}
```

This code *is* changed in `rand@0.9.0`, so I definitely need to take another look with the new version.

PR #552, opened 2025-09-20
fix: Handle no-baseline case gracefully

k88hudson-cfa:


RobertJacobsonCDC (approved):
I guess we need to merge it to even test it.

Issue #283, opened 2025-03-17
Profiling infrastructure for indexes


RobertJacobsonCDC:
Create profiling infrastructure to measure speed / efficiency of indexes and queries. This will provide information to support:
- #279 
- #280

RobertJacobsonCDC:
Some research on this revealed:

- multiproperty indexes, "multi-indexes", provide a substantial and meaningful benefit
- multi-indexes can be implemented as a single index on a derived tuple, a feature we already have
- automatically creating indexes, or other features that require gathering statistic on indexing will probably require some refactoring

Given that we already have multi-indexing "for free," we decided to forego the engineering work it would take to do automatic profiling, at least for now.

RobertJacobsonCDC:
Track how many times we are called and produced a huge or difficult to compute result and issue a warning in that case.
Count: how many each query is called, the size. Maybe it's dev build only.
Suggest the user use counter index or iterator API.

RobertJacobsonCDC:
I think it makes sense to leverage the profiling module for this. Let's backport that module from epi-isolation before implementing this. It also makes sense to implement this for entities.

Blocked by:
- entities
- backporting profiling module.

PR #556, opened 2025-09-22
feat: `with_query_results` API. Fixes #540.

### LINKED ISSUE #540: `with_query_results` API
### GET FULL SET OF IDS

Get ALL people matching a query

#### Use case examples

- Get all people in a setting, for some edge in a Network
- Get all people matching some properties to apply a policy
- Fine-grained debugging
- In tests

#### Work to be done

- Deprecate `query_people` (it's a footgun!)
- New API `with_query_results` method that takes a closure or function pointer that is given an
  immutable reference to an index set. This obviates the need for a `Cow`-like type, prevents the
  immutable reference from escaping its scope, and encourages the reference to be dropped before
  any mutable operations on `context` are performed. Client code can create an owned copy if they want it.

```rs
context.with_query_results(((Region, Region:A), (AgeGroup, AgeGroupValue:Senior)), |people: &HashSet<PersonId>| {
   for person in people {
     context.set_property(person, IsolationPolicyEnabled, true);
   }
});
```

([Roadmap](https://github.com/CDCgov/ixa/wiki/Making-Queries-Faster-Roadmap))

RobertJacobsonCDC:
- Added `index_benches.rs`.
- Added `with_query_results` method.
- Changed tests to use `with_query_results` instead of `query_people`

This PR is stacked on #534, but it doesn't have to be.

### DIFF
--- a/examples/network-hhmodel/network.rs
+++ b/examples/network-hhmodel/network.rs
@@ -18,7 +18,10 @@ fn create_household_networks(context: &mut Context, people: &[PersonId]) {
     for person_id in people {
         let household_id = context.get_person_property(*person_id, HouseholdId);
         if households.insert(household_id) {
-            let mut members = context.query_people((HouseholdId, household_id));
+            let mut members: Vec<PersonId> = Vec::new();
+            context.with_query_results((HouseholdId, household_id), &mut |results| {
+                members = results.iter().copied().collect()
+            });
             // create a dense network
             while let Some(person) = members.pop() {
                 for other_person in &members {
@@ -36,10 +39,16 @@ fn load_edge_list<T: EdgeType + 'static>(context: &mut Context, file_name: &str,
 
     for result in reader.deserialize() {
         let record: EdgeRecord = result.expect("Failed to parse edge");
-        let p1_vec = context.query_people((Id, record.v1));
+        let mut p1_vec = Vec::new();
+        context.with_query_results((Id, record.v1), &mut |people| {
+            p1_vec = people.to_owned_vec()
+        });
         assert_eq!(p1_vec.len(), 1);
         let p1 = p1_vec[0];
-        let p2_vec = context.query_people((Id, record.v2));
+        let mut p2_vec = Vec::new();
+        context.with_query_results((Id, record.v2), &mut |people| {
+            p2_vec = people.to_owned_vec()
+        });
         assert_eq!(p2_vec.len(), 1);
         let p2 = p2_vec[0];
         context.add_edge_bidi::<T>(p1, p2, 1.0, value).unwrap();
--- a/examples/network-hhmodel/network.rs
+++ b/examples/network-hhmodel/network.rs
@@ -20,7 +20,7 @@ fn create_household_networks(context: &mut Context, people: &[PersonId]) {
         if households.insert(household_id) {
             let mut members: Vec<PersonId> = Vec::new();
             context.with_query_results((HouseholdId, household_id), &mut |results| {
-                members = results.iter().copied().collect()
+                members = results.to_owned_vec()
             });
             // create a dense network
             while let Some(person) = members.pop() {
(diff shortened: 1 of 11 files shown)


src/hashing.rs
```diff
@@ -35,13 +35,23 @@ impl<K, V> HashMapExt for HashMap<K, V> {
...
+impl<T: Clone> HashSetExt for HashSet<T> {
+    type Item = T;
+
     fn new() -> Self {
         HashSet::default()
     }
+
+    fn to_owned_vec(&self) -> Vec<Self::Item> {
+        self.iter().cloned().collect()
+    }
```
RobertJacobsonCDC, line 54: I added this convenience method when I noticed the ergonomics are poor for the case that we want to make our own owned copy.

ixa-bench/src/reference_sir/sir_ixa.rs
```diff
@@ -70,6 +70,7 @@ impl InfectionLoop for Context {
     }
     fn infected_people(&self) -> usize {
         if self.get_options().queries_enabled {
+            #[allow(deprecated)]
             self.query_people((InfectionStatus, InfectionStatusValue::Infectious))
                 .len()
```
RobertJacobsonCDC, line 75: Shouldn't this use `query_peoplen_count`?

examples/network-hhmodel/network.rs
```diff
@@ -18,7 +18,10 @@ fn create_household_networks(context: &mut Context, people: &[PersonId]) {
     for person_id in people {
         let household_id = context.get_person_property(*person_id, HouseholdId);
         if households.insert(household_id) {
-            let mut members = context.query_people((HouseholdId, household_id));
+            let mut members: Vec<PersonId> = Vec::new();
+            context.with_query_results((HouseholdId, household_id), &mut |results| {
+                members = results.iter().copied().collect()
```
RobertJacobsonCDC, line 23: `results.to_owned_vec()`

ixa-bench/src/reference_sir/sir_ixa.rs
RobertJacobsonCDC: Also line 333 below.

k88hudson (approved):
Mostly small comments, overall looks good!

k88hudson-cfa (approved):

src/people/context_extension.rs
```diff
@@ -561,6 +579,57 @@ impl ContextPeopleExt for Context {
 
         selected
     }
+
+    fn with_query_results<Q: Query>(&self, query: Q, scope: &mut dyn FnMut(&HashSet<PersonId>)) {
+        // Special case the empty query, which creates a set containing the entire population.
+        if query.type_id() == TypeId::of::<()>() {
```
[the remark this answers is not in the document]
RobertJacobsonCDC: I'm keeping notes on this idea on #283.

ixa-bench/criterion/index_benches.rs
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    context.index_property(Property10);
+    context.index_property(Property100);
+    context.index_property(MProperty);
+
+    criterion.bench_function("with_query_results_single_indexed_property", |bencher| {
+        bencher.iter(|| {
+            for number in &numbers {
+                black_box(context.with_query_results(
```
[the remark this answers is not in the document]
RobertJacobsonCDC: Under the hood, a closure is an anonymous struct, in this case a constant having no fields. The purpose of `black_box` is to hide the fact that the arguments to `context.with_query_results` are constants from the compiler to prevent the compiler from deciding that the output of the function call will always be the same. In this case, it's pretty unlikely that the compiler would hoist the function call out of the loop anyway, but the rule of thumb is to wrap the function inputs and the function output.

The one I'm not sure about is the `black_box` _inside_ the closure, though. That one probably isn't necessary.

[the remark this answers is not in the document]
RobertJacobsonCDC: Changed the second parameter to `black_box(&mut |_| {})`.

PR #562, opened 2025-09-26
(Experiment) feat: Use mise for tasks and tool installation

k88hudson-cfa:
I'm experimenting with a different approach for tool management / tasks here.
Note that there was some weirdness in files getting formatted differently (probably because of version stuff?)  

The task running part of mise is pretty similar to just, but it replaces a lot of the manual tool installation with mise's tool registry.

I got the tasks down to ~172 lines, although I think we could probably do even better! 

Everything should be working with the Docker file, codespace, and Github actions


cdc-as81 (approved):
LGTM

PR #572, opened 2025-10-07
fix: mise should not run wildcard tasks

cdc-as81:


mise.toml
```diff
@@ -118,15 +126,19 @@ run = [
 [tasks.prettier]
 description = "Run prettier on all supported files"
 run = '''prettier --{{arg(name='action', default='check')}} "**/*.{md,json,yml,yaml}" \
-  --ignore-path ./.gitignore --ignore-path ./.markdownlintignore
+  --ignore-path ./.gitignore --ignore-path ./.markdownlintignore --ignore-path ./.prettierignore
```
RobertJacobsonCDC, line 129: Ah, I remember now that we use `.markdownlintignore` for both markdown lint tools. Instead of adding `.prettierignore`, add the contents of `.prettierignore`  to `.markdownlintignore`, because we want generated markdown to be ignored by both tools.

mise.toml
```diff
@@ -104,7 +109,10 @@ run = './target/release/hyperfine'
 
 [tasks.check]
 description = "Formatting and linting"
-run="mise run check:*"
+run = [
+  "mise run check:rust",
+  "mise run check:markdown"
+]
```
RobertJacobsonCDC, line 115: Ok, so I think there are [two ways to do this](https://mise.jdx.dev/tasks/running-tasks.html#wildcards) so that `mise` automatically runs all the `check:` tasks. The first way is to specify them as a dependency:

```toml
[tasks.check]
description = "Formatting and linting"
depends = ["check:*"]
```

The second way is to quote the pattern:

```toml
[tasks.check]
description = "Formatting and linting"
run = 'mise run "check:*"'
```

These seem to work on my machine. If you can get one of these to work, it'd be better than having to manually list everything.

RobertJacobsonCDC (changes requested):

PR #589, opened 2025-10-23
chore: move macros to macros files

cdc-as81:
moved macros to macros files. no changes were done to the macros.

### DIFF
--- a/.github/workflows/bench-compare.yaml
+++ b/.github/workflows/bench-compare.yaml
@@ -37,9 +37,11 @@ jobs:
           cat hyperfine.md >> results.md
           echo '#### Criterion'
           echo '```' >> results.md
-          if ! mise run bench:compare '' base 2>&1; then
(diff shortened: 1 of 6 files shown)


.devcontainer/post-create.sh
```diff
@@ -21,3 +21,5 @@ mise trust
 # Add pre-commit hooks
 mise install:hooks
 
+#install needed libs for wasm
+sudo apt-get install -y libcups2 libcairo2 libpango-1.0-0
```
RobertJacobsonCDC: 1. Is this change related to this PR?
2. These are odd dependencies for Wasm. What is claiming to need these?

RobertJacobsonCDC (changes requested):
The Wasm test fails, but it looks like it fails on `main`, too. Did we break it when we switched to mise and didn't notice?

RobertJacobsonCDC (approved):

PR #601, opened 2025-11-06
feat: Add get_id to PersonId

k88hudson-cfa:


k88hudson-cfa:
I think as long as people aren't creating PersonIds I don't think there's too much danger here – what are you thinking of? I Getting direct access to the index is turning out to be pretty important for optimizing output-related things / some data structures

RobertJacobsonCDC:
# Can we achieve what we want by providing one or two primitive data structures?

I should have put these comments in this thread:

What if we provide a primitive like this? 
https://github.com/RobertJacobsonCDC/ixa-entities/blob/EntityKeyedMap/src/EntityKeyedMap.rs

Seems to me you'd have two use cases:

1. **When you want `EntityKeyedMap<E: Entity, T>` to be `Vec<T>`-like.**  
   In this use case you could have it deref to a `Vec<T>`.  Technically, that would leak the inner `usize` of `EntityId<E>`, but not in a very useful way,  so it strikes me as a reasonable compromise.  You get and set using `EntityId<E>`, but you can then operate on the `T` values themselves  without regard to their corresponding `EntityId<E>`.

2. **When you want `EntityKeyedMap<E: Entity, T>` to be `HashMap<EntityId<E>, T>`-like.**  
   In this case you'd wrap a `Vec<Option<T>>` instead so that keys could be absent (empty slots).

The [`PropertyValueStore`](https://github.com/RobertJacobsonCDC/ixa-entities/blob/EntityKeyedMap/src/property_value_store.rs) in the in-progress Entities implementation acts like a `HashMap<EntityId<E>, T>`,  but it wraps a [`ValueVec`](https://github.com/RobertJacobsonCDC/ixa-entities/blob/EntityKeyedMap/src/value_vec.rs) and so requires `T: Copy`, which is more restrictive than you want.

PR #602, opened 2025-11-06
feat: Created method get_entity_property

### LINKED ISSUE #596: Support enum properties for define_person_property
```rust

define_person__property_with_default!(person_id, AgeGroup, {Child, Adult}, AgeGroup::Child);

define_person_property!(AgeGroup, {Child, Adult}));

// This creates an enum marker type, that is also the value
enum AgeGroup {
  Child,
  Adult
}
impl PersonProperty for AgeGroup {
    type = AgeGroup
}

// Current
context.get_person_property(person_id, AgeGroup);
```

Since this requires an API change, perhaps we could implement a new method for this: 
```rust 
context.get_entity_property::<_, AgeGroup>(person_id);
let age: AgeGroup = context.get_entity_property(person_id);
```

vinibuzzacaro:
Created three new methods: get_entity_property, get_entity_property_ref and set_entity_property. The methods are basically a copy of get_person_property, get_person_property_ref and set_person_property but i've removed the argument 'propert: T' of those functions.
fix #596

### DIFF
--- a/src/people/context_extension.rs
+++ b/src/people/context_extension.rs
@@ -19,6 +19,7 @@ use crate::{
 /// A trait extension for [`Context`] that exposes the people
 /// functionality.
 pub trait ContextPeopleExt {
+    fn get_entity_property<T: PersonProperty>(&self, person_id: PersonId) -> T::Value;
     /// Returns the current population size
     fn get_current_population(&self) -> usize;
 
@@ -163,6 +164,20 @@ impl ContextPeopleExt for Context {
 
         self.emit_event(PersonCreatedEvent { person_id });
         Ok(person_id)
+    }
+
+    fn get_entity_property<T: PersonProperty>(&self, person_id: PersonId) -> T::Value {
+        let data_container = self.get_data(PeoplePlugin);
+        self.register_property::<T>();
+        if T::is_derived() {
+            return T::compute(self, person_id);
+        }
+        if let Some(value) = *data_container.get_entity_property_ref::<T>(person_id) {
+            return value;
+        }
+        let initialized_value = T::compute(self, person_id);
+        data_container.set_entity_property::<T>(person_id, initialized_value);
+        initialized_value
     }
 
     fn get_person_property<T: PersonProperty>(&self, person_id: PersonId, property: T) -> T::Value {
--- a/src/people/data.rs
+++ b/src/people/data.rs
@@ -262,6 +262,36 @@ impl PeopleData {
         })
     }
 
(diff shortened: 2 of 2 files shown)


RobertJacobsonCDC:
I think the way forward on this feature is via Entities, which is a feature still in progress. Here's the short version of the story:

Right now a "person" is represented by a `PersonId` and has multiple `PersonProperty` values associated with it. Entities generalizes this: An `Entity` is analogous to a table in a relationship database, and the properties of an entity are analogous to the columns that exist on the table. A row in the table is addressable with an `EntityId<Entity>` (implemented as a newtype of a `usize`). In this new paradigm, `Person` is a particular entity, possibly one of several, and `PersonId` is a type alias for `EntityId<Person>`.

One way to implement the API for entity property getters and setters is to have methods on `Context` like this:

```rust
// The `my_entity_id` value is of type `MyEntityId`, which is a type alias for `EntityId<MyEntity>`.
// (The `MyProperty` type knows which entity it belongs to.)
let my_property_value = context.get_property::<MyProperty>(my_entity_id);
// Turbofish-less version of the same call:
let my_property_value: MyProperty = context.get_property(my_entity_id);

// For setters, the property is inferred from the type of the value we are passing in.
context.set_property(my_entity_id, some_property_value);
// ...but if you want to be super-explicit, you could use a turbofish version:
context.set_property::<MyProperty>(my_entity_id, some_property_value);
```

These general accessor methods will be part of the new `entity` module, not part of the (eventually to be deprecated) `people` module.

We also discussed synthesizing a trait extension for `Context` for every new property (using a macro) that provides *named* accessor methods for the property. So if you defined a new property named `Age` for the entity `Person` (defined elsewhere) with the macro invocation `define_property!(Age, Person);`, it would implement a trait extension on `Context` like in this pseudo-Rust:

```rust
pub trait AgeContextExt {
  // Named "get_" + snake_case(property_name)
  fn get_age(&self, person_id: PersonId) -> Option<Age> {
    self.get_property::<Age>(person_id)
  }
  // Named "set_" + snake_case(property_name)
  fn set_age(&self, person_id: PersonId, age: Age) {
    self.set_property::<Age>(person_id, age);
  }
}

// Now client code could make calls like this:
let my_age = context.get_age(person_id);
context.set_age(person_id, Age(32));
```

(We might want to include the entity name, too: `Context::get_person_age()` and `Context::set_person_age()`.) This would save the client code author from having to use the turbofish, which syntax experience has shown can be confusing to new users, at least initially.

These details aren't yet set in stone.

PR #615, opened 2025-11-24
chore: Add periodic counts benchmark for performance testing of add_p…

cdc-as81:
…eriodic_report


ixa-bench/src/reference_sir/bench.rs
```diff
@@ -3,6 +3,7 @@ use crate::hyperfine_group;
 
 fn build_params() -> Parameters {
     ParametersBuilder::default()
+        // when increasing population also increase max_time accordingly.
```
k88hudson-cfa: I don't really understand this comment, do you mean reduce it so the run time is lower?

PR #617, opened 2025-11-24
fix: benchmark action branch checkout

cdc-as81:


k88hudson-cfa (changes requested):
This would expose secrets/PR write access to untrusted code, which is definitely not something you want to do here – if you want to have the action run on untrusted code, you need to isolate the untrusted part / scope the action to read access only

PR #633, opened 2025-12-08
feat: enable negative time

erik-rosenstrom:
This PR enables scheduling of plans at negative times. This is done by setting `context.current_time = AtomicF64(f64::NEG_INFINITY)` upon creation. When `context.get_current_time()` is called `context.current_time` is initialized to the time of the first plan if the time of the plan is less than 0 and 0 otherwise. `context.execute` calls `context.get_current_time()` to initialize `context.current_time`

### DIFF
--- a/ixa-bench/src/bench_utils/hyperfine.rs
+++ b/ixa-bench/src/bench_utils/hyperfine.rs
@@ -4,7 +4,7 @@ use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
 use clap::builder::PossibleValuesParser;
-use clap::{arg, Parser};
+use clap::Parser;
 use ixa_bench::bench_utils::registry::{list_benches, list_groups};
 
 const STYLE_FLAGS: [&str; 2] = ["--show-output", "--style"];


src/context.rs
```diff
@@ -111,7 +114,7 @@ impl Context {
             data_plugins,
             #[cfg(feature = "debugger")]
             breakpoints_scheduled: Queue::new(),
-            current_time: 0.0,
+            current_time: AtomicF64::new(f64::NEG_INFINITY),
```
k88hudson-cfa: Can you add a comment with the initialization semantics, i.e. that this is a placeholder for the initial time, where we use the first time if it's negative or 0 if positive?

RobertJacobsonCDC:
Semantics of the method `Context::get_current_time()` :

- Prior to `Context::execute()`, returns whatever the start time is set to be
  - The `start_time` provided to  `Context::set_start_time()` , or
  - `0.0` otherwise.
  - Returns `0.0` _before_ a call to   `Context::set_start_time()`  and  `start_time` afterwards.
- After `Context::execute()`, whatever the current simulation time is.

Semantics of `Context::set_start_time(start_time: f64)`:

- `start_time` must be finite.
- no already scheduled plan can be scheduled at a time `t < start_time`. (Implementation: just check `start_time >= context.plan_queue.next_time()`.)
- Should we require that `start_time <= 0.0`? (I don't think we need to.)
- `Context::set_start_time` can be called at most once. (Implementation: detect if `context.set_start_time(0.0)` has been called by either using an option or a non-finite sentinel value.)
- `Context::set_start_time`  cannot be called after `Context::execute` _even if it hasn't been called before,_ and even if execution has been stopped for some reason. (The simulation time has already been established at that point.)



1. What constraints should there be on the time at which plans can be scheduled prior to `context.execute()`?
   - If `Context::set_start_time()` has never been called, plans cannot be scheduled prior to time `0.0` (results in a panic).
   - If `Context::set_start_time(start_time)`  has been called, plans cannot be scheduled prior to `start_time` (results in a panic).
2. To what value should `context.current_time` be initialized?
   -  `Context::set_start_time(start_time)`  can be called prior to `Context.execute()` to initialize to a specific `start_time`.
   - `0.0` if  `Context::set_start_time(start_time)`  has not been called.
3. When should the initialization of `context.current_time` occur?
   - Upon `Context` creation, to `0.0`. Client code can call `Context::set_start_time(start_time)`  at most once prior to `Context.execute()`. (Implementation might do something different internally.)

confunguido (changes requested):
I would prefer explicit initialization of the time unless the default is always 0.0.

src/context.rs
```diff
@@ -921,6 +964,293 @@ mod tests {
...
+        context.set_start_time(-5.0);
+        add_plan(&mut context, -5.0, 1);
+        add_plan(&mut context, 10.0, 2);
+        context.execute();
+        assert_eq!(context.get_current_time(), 10.0);
+        assert_eq!(*context.get_data_mut(ComponentA), vec![1, 2]);
+    }
+
+    #[test]
+    fn get_current_time_initializes_to_zero_when_all_positive() {
```
RobertJacobsonCDC: This test doesn't really do what it's name suggests. I suggest replacing it with a test that queues a callback that checks that the current time is zero. It would be testing the fact that by default a new `Context` with no scheduled plans initializes its current time to zero.

src/context.rs
```diff
@@ -921,6 +964,293 @@ mod tests {
...
+        let mut context = Context::new();
+        add_plan(&mut context, 1.0, 1);
+        add_plan(&mut context, 2.0, 2);
+        context.execute();
+        assert_eq!(context.get_current_time(), 2.0);
+    }
+
+    #[test]
+    fn get_current_time_initializes_to_zero_when_empty() {
```
RobertJacobsonCDC: We should have a test that checks the time reported by `context.get_current_time()` before `context.execute()` is ever called.

src/context.rs
```diff
@@ -349,12 +354,46 @@ impl Context {
...
     }
 
+    /// Set the start time for the simulation. Must be finite.
+    ///
+    /// This should be called before `Context.execute()`.
+    /// Can be called only once.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `start_time` is infinite or NaN.
```
RobertJacobsonCDC: Let's expand and reformat this to something like: 

```rust
/// Set the start time for the simulation.
///
/// * This should be called before [`Context::execute`].
/// * `start_time` must be finite (not NaN or infinite).
/// * This can be called only once.
/// * If any plans are already scheduled, `start_time` must be **greater than or equal to**
///   the earliest scheduled plan time.
///
/// # Panics
///
/// Panics if:
/// * `start_time` is NaN or infinite.
/// * the start time has already been set or `Context::execute()` has already been called.
/// * `start_time` is earlier than the earliest scheduled plan time.
```

src/context.rs
```diff
@@ -410,6 +449,17 @@ impl Context {
     /// Execute the simulation until the plan and callback queues are empty
     pub fn execute(&mut self) {
         trace!("entering event loop");
+
+        // Initialize `self.current_time` based on start_time or earliest plan
+        if let Some(start_time) = self.start_time {
+            self.current_time = start_time;
+        } else {
+            self.current_time = match self.plan_queue.next_time() {
+                Some(t) if t < 0.0 => t,
```
RobertJacobsonCDC: This is incorrect as written, because it implicitly sets the start time to the earliest scheduled negative plan time if no start time was specified. What we want is, the _only_ way start time is not 0.0 is if client code calls `context.set_start_time(start_time)` with their own custom `start_time` before they call `context.execute()`. We have a check in `Context::add_plan`, so I'm not sure the `Some(t)` branch is reachable.... But that's another reason to rewrite this code!

src/context.rs
```diff
@@ -241,6 +252,11 @@ impl Context {
     /// continuously repeat the plan at the specified period, stopping
     /// only once there are no other plans scheduled.
     ///
+    /// Notes:
+    /// * The first periodic plan is scheduled at time `0.0`. If `set_start_time` was
+    ///   set to a positive value, this will currently panic because the first plan
+    ///   occurs before the start time (see issue #634 for future behavior).
```
RobertJacobsonCDC: I like this.

src/context.rs
```diff
@@ -208,9 +210,13 @@ impl Context {
...
         phase: ExecutionPhase,
     ) -> PlanId {
+        let current = self.get_current_time();
         assert!(
-            !time.is_nan() && !time.is_infinite() && time >= self.current_time,
-            "Time is invalid"
+            !time.is_nan()
+                && !time.is_infinite()
+                && time >= current,
+            "Time {time} is invalid: must be finite and not less than the current time ({}). Consider calling set_start_time() before scheduling plans.",
```
RobertJacobsonCDC: This is a nitpick, but I'd split this assert into multiple asserts and have specialized error messages for each case.

k88hudson-cfa (approved):
LGTM assuming other reviewers are happy (I looked pretty high level, the changes look great!)

PR #636, opened 2025-12-09
feat: Modified how `Context::current_time` is initialized to allow for negative time.

RobertJacobsonCDC:
We initialize to the sentinel `f64::NEG_INFINITY` upon construction. Then when a concrete value is needed we initialize (in `Context::get_current_time`) to the min of smallest scheduled plan time or 0.0.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -261,8 +261,6 @@ impl Context {
             "Period must be greater than 0"
         );
 
-        //
-
         self.add_plan_with_phase(
             0.0,
             move |context| context.evaluate_periodic_and_schedule_next(period, callback, phase),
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -923,4 +956,161 @@ mod tests {
...
+    fn get_current_time_initializes_to_zero_when_all_positive() {
+        let mut context = Context::new();
+        add_plan(&mut context, 1.0, 1);
+        add_plan(&mut context, 2.0, 2);
+        // Should initialize to 0.0 since all plans are positive
+        assert_eq!(context.get_current_time(), 0.0);
+        context.execute();
+        assert_eq!(context.get_current_time(), 2.0);
+    }
+
```
erik-rosenstrom: I would maybe add a test with a plan schedule at time 0.0 after `get_current_time()` is called to make sure that it executes

src/context.rs
```diff
@@ -113,7 +115,7 @@ impl Context {
             data_plugins,
             #[cfg(feature = "debugger")]
             breakpoints_scheduled: Queue::new(),
-            current_time: 0.0,
+            current_time: AtomicF64::new(f64::NEG_INFINITY),
```
erik-rosenstrom: Just from the perspective of an ixa user, but the `AtomicF64` seems like complicated way to handle `current_time`. It might be useful to provide justification in documentation for context initialization

src/context.rs
```diff
@@ -352,12 +358,41 @@ impl Context {
...
+    /// Get the current time in the simulation.
     ///
-    /// Returns the current time
+    /// Prior to `Context.execute()` or the first call to this method, `self.current_time` is uninitialized. Calling
+    /// this method before `Context.execute()` will initialize the current time to the minimum of
+    /// the time of the earliest scheduled plan (if one exists) and 0.0.
     #[must_use]
     pub fn get_current_time(&self) -> f64 {
-        self.current_time
+        // ToDo(RobertJacobsonCDC):
```
erik-rosenstrom: What is the ToDo? Verify performance?

src/context.rs
cdc-as81: maybe I am missing something here.
But as a user you should assign a f64 to the time, you do not need to use AtomicF64.

erik-rosenstrom (approved):
Looks good to me

PR #642, opened 2025-12-11
fix: wasm build in test action

cdc-as81:


RobertJacobsonCDC (approved):
My issue occurs with the `main` branch, too. Since neither Gil, Kate, or CI can reproduce my issue, let's not block this PR.

I've opened a separate issue #651 and assigned myself, since it appears to be specific to my environment.

PR #644, opened 2025-12-11
fix: Fixed mangled callout formatting. Fixes #641.

### LINKED ISSUE #641: ixa Book callout rendered incorrectly
The callout on this page is rendered incorrectly:

https://ixa.rs/book/topics/random-module.html

Image

RobertJacobsonCDC:
At some point we must have run an autoformatting tool on our markdown in the ixa book that mangled the syntax we use for callouts, and we didn't catch it. This results in callouts being rendered incorrectly. This PR fixes the callout syntax.


cdc-as81 (approved):
LGTM!

PR #663, opened 2026-01-06
chore: add print execution stats fn

cdc-as81:


RobertJacobsonCDC (approved):
This makes a lot of sense to me. I wouldn't bother with the comment:

```rust
// ProfilingContextExt moved to reports.rs
```

But that's just personal preference.

PR #680, opened 2026-01-28
chore: Move entities methods to a context extension trait

k88hudson-cfa:


src/entity/context_extension.rs
```diff
@@ -0,0 +1,489 @@
...
+    }
+    fn query_result_iterator<E: Entity, Q: Query<E>>(&self, query: Q) -> QueryResultIterator<E> {
+        query.new_query_result_iterator(self)
+    }
+}
+
+#[cfg(test)]
+#[allow(clippy::float_cmp)]
+mod tests {
+    #![allow(dead_code)]
```
RobertJacobsonCDC: Both of the `#[allow(...)]` attributes here can be deleted, and doing so happens to silence a clippy lint, so double bonus.

I have the corresponding `#![allow(clippy::float_cmp)]` attribute commented out in `src/context.rs` to test if it was needed. Turns out it's not needed there either, but I forgot to delete the comment! Maybe you could remove that one, too?

RobertJacobsonCDC (approved):
Awesome! Thanks!

Can you run `mise run fix:rust` before merging so it passes CI?

PR #682, opened 2026-01-29
Update reference sir

k88hudson-cfa:
I removed the no queries option and simplified this to legacy/entities.

There seems to be a weird bug here though where the entities count (which uses `self.query_entity_count`) isn't updating after recovering people? I see the test fail on the line after recovering the first person (the count should go from 6 -> 5).
The attack rate is also way too high

### DIFF
--- a/ixa-bench/src/reference_sir/sir_ixa/entities.rs
+++ b/ixa-bench/src/reference_sir/sir_ixa/entities.rs
@@ -61,9 +61,15 @@ impl InfectionLoop for Context {
     }
 
     fn recover_person(&mut self, p: PersonId, _t: f64) {
-        assert_eq!(self.get_property::<_, InfectionStatus>(p), InfectionStatus::Infectious);
+        assert_eq!(
+            self.get_property::<_, InfectionStatus>(p),
+            InfectionStatus::Infectious
+        );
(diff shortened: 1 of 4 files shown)


RobertJacobsonCDC:
It looks like #685 fixes it. I rebased locally, and everything seems to run. All tests pass.

Do you want to rebase and run `mise run fix:rust` to keep the linter happy?

ixa-bench/src/reference_sir/sir_ixa/entities.rs
```diff
@@ -98,20 +58,15 @@ impl InfectionLoop for Context {
             let stats_data = self.get_data_mut(ModelStatsPlugin);
             stats_data.record_infection(current_t);
         }
-
-        // Update the non-query index
-        self.get_data_mut(NonQueryInfectionTracker).insert(p);
     }
 
     fn recover_person(&mut self, p: PersonId, _t: f64) {
+        assert_eq!(self.get_property::<_, InfectionStatus>(p), InfectionStatus::Infectious);
```
RobertJacobsonCDC: I've surprised the compiler can't infer the return type in these comparisons.

RobertJacobsonCDC (approved):

PR #698, opened 2026-02-02
Use IndexSet for indexes

k88hudson-cfa:
Profiling revealed `sample_entities` was responsible for most of the slowdown in the reference benchmark – given that randomly selecting a next contact is a very common operation, having it be O(n) seems like the wrong trade off. 

@RobertJacobsonCDC  pointed out there is some memory overhead here (~x2) so we should consider support for different data structures before too long. 

```
mise run bench:criterion sample_entity
```

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -13,12 +13,18 @@
 //! in scope.
 //!
 
-use std::hash::{Hash, Hasher};
+use std::hash::{BuildHasherDefault, Hash, Hasher};
 
 use bincode::serde::encode_to_vec as serialize_to_vec;
-pub use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
+pub use indexmap::set::Iter as IndexSetIter;
+use indexmap::IndexSet as RawIndexSet;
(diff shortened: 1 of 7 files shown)


k88hudson-cfa:
I'm going to add a more specific benchmark for this case

src/hashing.rs
```diff
@@ -57,6 +57,18 @@ impl<T: Clone> HashSetExt for HashSet<T> {
     }
 }
 
+impl<T: Clone> HashSetExt for indexmap::IndexSet<T> {
```
RobertJacobsonCDC: Let's use our fast hash function with `IndexSet` as we do with `HashMap`. So in `hashing.rs` we'd have something like 

```rust
use indexmap::IndexSet as RawIndexSet;
use rustc_hash::FxHasher;
use std::hash::BuildHasherDefault;

type FxBuildHasher = BuildHasherDefault<FxHasher>;

pub type IndexSet<T> = RawIndexSet<T, FxBuildHasher>;
```

RobertJacobsonCDC (changes requested):
Nice work!

Issue #611, opened 2025-11-18
Identify the overhead of `sir_ixa` compared to `sir_baseline`


RobertJacobsonCDC:
Setting aside the performance of queries, the pure rust `large_sir::baseline` model outperforms the ixa-based `large_sir::ixa_no_queries` and `large_sir::ixa_entities_no_queries` by a factor of 4x-5x. It's not clear why.

- Getting and setting property values should be fast in `large_sir::ixa_entities_no_queries`, so that's unlikely to be the cause.
- ixa-based models make a lot of function calls, but most of them should be inlined for this model.
- ixa-based models have a lot of checks (Do we need to emit an event? Is this parameter in range? Do we need to update an index?) that the pure Rust model can dispense with.

These explanations don't match my intuition. We need to measure.

PR #699, opened 2026-02-02
sample_entity benchmark 

k88hudson-cfa:
Based on investigation in #611, this captures the main culprit for the query slowness

```
mise run bench:criterion sample_entity
```

### DIFF
--- a/ixa-bench/criterion/sample_entity_scaling.rs
+++ b/ixa-bench/criterion/sample_entity_scaling.rs
@@ -1,36 +1,31 @@
+use std::collections::BTreeMap;
 use std::hint::black_box;
+use std::sync::{Arc, Mutex};
+use std::time::Instant;
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use ixa::prelude::*;
 
-// This benchmark checks for regressions that sample_entity is O(1)
-// This is a very common use case in a model, i.e., "get the next person to infect"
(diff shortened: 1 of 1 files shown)


ixa-bench/criterion/sample_entity_scaling.rs
```diff
@@ -0,0 +1,104 @@
+use std::hint::black_box;
+
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
+use ixa::prelude::*;
+
+// This benchmark checks for regressions that sample_entity is O(1)
```
k88hudson-cfa: Right now it is not, but #698 adds a fix which should make it true. I could theoretically have an actual test that fails past a certain threshold

ixa-bench/criterion/sample_entity_scaling.rs
```diff
@@ -0,0 +1,180 @@
...
+    bencher.iter_custom(|iters| {
+        let start = Instant::now();
+        for _ in 0..iters {
+            f();
+            black_box(());
+        }
+        let elapsed = start.elapsed();
+
+        // record ns/sample for this measurement batch
+        ns_per_sample = elapsed.as_secs_f64() * 1e9 / (iters as f64);
```
RobertJacobsonCDC: I might be reading this wrong, but it looks like you are overwriting `ns_per_sample` for every batch and thus are only reporting the value computed for the last batch. Is this intentional?

PR #706, opened 2026-02-03
Proposal: explicit entity id alias

k88hudson-cfa:
I'd like to propose that we have people define an explicit entity id alias instead of auto-generating it with paste. Given that most people won't be declaring entities / it doesn't happen often, I personally think this is:
* more explicit, doesn't require looking up docs to check what the generated type is called
* more searchable:

Change: 
```rs
define_entity!(Mosquito, MosquitoId);
```

Old: 
```rs
define_entity!(Mosquito);
// MosquitoId is generated automatically 
```

Right now in the PR the behaviour is optional, but I'd slightly be in favour of making `define_entity` non-optional (to prevent people forgetting) and leaving `impl_entity` to be optional in case you don't want the alias for some reason. Thoughts?

Still TODO in the PR: update docs.

### DIFF
--- a/examples/basic-infection/src/people.rs
+++ b/examples/basic-infection/src/people.rs
@@ -3,7 +3,7 @@ use ixa::trace;
 
 use crate::POPULATION;
 
-define_entity!(Person);
+define_entity!(Person, PersonId);
 
 // In this model, people only have a single property, their infection status.
 define_property!(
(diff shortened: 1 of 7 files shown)


k88hudson-cfa:
I think we can leave this one in feedback for now, I will re-open if it comes up

PR #708, opened 2026-02-03
Add cli

k88hudson-cfa:
You can test this by pulling the branch and then running `mise run install:cli`

Then, go to some location where you want to create a new ixa project, run

```
ixa new
```

Or try

```
ixa help
```

### DIFF
--- a/src/cli/profile_command.rs
+++ b/src/cli/profile_command.rs
@@ -1,0 +1,44 @@
+use std::fs;
+use std::path::Path;
+
+use super::utils::*;
+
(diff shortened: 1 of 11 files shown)


src/cli/profile_command.rs
```diff
@@ -0,0 +1,44 @@
+use std::fs;
+use std::path::Path;
+
+use super::utils::*;
+
+const PROFILING_PROFILE: &str = r#"
```
k88hudson-cfa: This is like the samply workflow so i was thinking might be good for us too?

PR #713, opened 2026-02-04
feat: Add all! macro to replace tuples

k88hudson-cfa:
This implementation leaves tuples in place (for now – I think if we like it, we might want to consider removing the tuples) but adds a new api via an `all!` macro.

Example:

```js
// Empty
let person = context.add_entity::<Person>(())
let person = context.add_entity(all!(Person))

// With properties
let person = context.add_entity(all!(Person, Age(17), Risk::Low))

// Also for queries
let person = context.sample_entity(SampleRng, all!(Person, InfectionStatus::Infected);

```

### DIFF
--- a/src/macros/all.rs
+++ b/src/macros/all.rs
@@ -1,0 +1,124 @@
+/// Creates a query matching all entities of a given type, optionally filtered by properties.
+///
+/// # Examples
+///
+/// ```ignore
+/// // Add an entity with default properties
+/// let query = all!(Person);
+/// context.add_entity(query)?;
+///
+/// // An inline query matching a single property
+/// let person = context.sample_entity(MyRng, all!(Person, Age(12)))?;
+///
+/// // A query matching multiple properties
+/// let query = all!(Person, Age(12), RiskCategory::High);
+/// let count = context.count_entities(query);
+/// ```
+#[macro_export]
+macro_rules! all {
+    // No properties - generates empty tuple query
+    ($entity:ty) => {
+        $crate::EntityPropertyTuple::<$entity, _>::new(())
+    };
+    // One or more properties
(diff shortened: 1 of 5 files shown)


src/macros/all.rs
```diff
@@ -0,0 +1,101 @@
...
+        $crate::EntityPropertyTuple::<$entity, _>::new(())
+    };
+    // One or more properties
+    ($entity:ty, $($prop:expr),+ $(,)?) => {
+        $crate::EntityPropertyTuple::<$entity, _>::new(($($prop,)+))
+    };
+}
+
+#[cfg(test)]
+mod tests {
```
RobertJacobsonCDC: How about a test to verify `all!` can disambiguate entities when a property belongs to multiple entities.

```rust
// Demonstrate that `all!` can disambiguate entities in otherwise ambiguous cases.
use crate::entity::EntityId;
define_entity!(TestMammal);
define_entity!(TestAvian);
define_property!(struct IsBipedal(bool), TestMammal, default_const = IsBipedal(false));
impl_property!(IsBipedal, TestAvian, default_const = IsBipedal(true));

#[test]
fn all_macro_disambiguates_entities() {

    context.add_entity(all!(TestAvian, IsBipedal(true))).unwrap();
    context.add_entity(all!(TestMammal, IsBipedal(true))).unwrap();

    let result = context.query_result_iterator(all!(TestMammal, IsBipedal(true)));
    assert_eq!(result.count(), 1);

    let sampled_id = context.sample_entity(AllMacroTestRng, all!(TestAvian, IsBipedal(true))).unwrap();
    let expected_id: EntityId<TestAvian> = EntityId::new(0);
    assert_eq!(sampled_id, expected_id);
}
```

PR #715, opened 2026-02-04
chore: Refactored and updated `network` module for entities. Fixes #710.

### LINKED ISSUE #710: The `network` module should use entities.
Right now the `network` module uses `PersonId` everywhere. It needs to be generic over `Entity` type and use `EntityId<E>`.

RobertJacobsonCDC:
I made some structural changes. `EdgeType` now uses the registry pattern, and the network type is accessed according to `Entity::id()` and `EdgeType::id()`.


docs/book/src/migration_guide.md
```diff
@@ -50,8 +50,16 @@ define_property!(
 );
 ```
 
-If you want to use an existing type, or if you want to make the same type a
-property of more than one `Entity`, you can use the `impl_property!` variant:
+The `Person` entity is so common that Ixa defines it for you in `ixa::prelude`:
```
k88hudson-cfa: Maybe remove this for now?

src/network/edge_impl.rs
```diff
@@ -0,0 +1,81 @@
+/*!
+
+Macros for implementing edge types.
+
+*/
+
+#[macro_export]
```
k88hudson-cfa: We will want to move this after when the pr that moves macros lands

k88hudson-cfa (approved):
We'll probably want to revisit this once we start using it more in future ixa work but looks good for now

PR #720, opened 2026-02-04
Fix serde export not working for derives

### LINKED ISSUE #671: Client code needs to have `serde` as a dependency for `Serialize`/`Deserialize` proc macros
For unclear reasons it seems like client code needs to list `serde` as a dependency:

```toml
[dependencies]
csv = "1.3.1"
ixa = { ... }
serde = { version = "1.0.217", features = ["derive"] }
```

It has something to do with how `derive` proc macros are resolved. [This page](https://github.com/serde-rs/serde/issues/1465) seems to suggest there's no supported way to re-export `Serialize`/`Deserialize`, but that doesn't sound right.

Use `docs/book/models/disease_model` for a simple example.

Ideally, client code shouldn't have to list either `csv` or `serde` as a dependency, as they are re-exported by ixa.

k88hudson-cfa:
Because:
- When you use serde with derive in a macro, you have to add a special annotation for the crate with `[serde(crate = ixa)]`
- You can't string concatenate `$crate`

Having usages of the macro in the same crate and also exporting it doesn't work

Since serde is so common, we decided to rip it out and make it a required peer dependency for now. In the future we could consider moving stuff to a ixa-macros crate if we want.

Fixes #671

### DIFF
--- a/src/prelude.rs
+++ b/src/prelude.rs
@@ -13,5 +13,5 @@ pub use crate::{
     define_data_plugin, define_derived_person_property, define_derived_property, define_edge_type,
     define_entity, define_global_property, define_multi_property, define_person_multi_property,
     define_person_property, define_person_property_with_default, define_property, define_report,
-    define_rng, PluginContext,
+    define_rng, impl_entity, impl_property, PluginContext,
 };


RobertJacobsonCDC (approved):
🚀

Issue #746, opened 2026-02-09
EntitySet and EntitySetIterator


RobertJacobsonCDC:
It turns out the `SourceSet` / `SourceSetIterator` and `QueryResultIterator` types are more general in their application that just queries. PR #745 renames `QueryResultIterator` to `EntitySetIterator` in preparation for a general type `EntitySet`.

The `EntitySet` type lifts `SourceSet`s to an algrebra of sets, implementing efficient set operations. These are useful in their own right (e.g. if you want to take the intersection of the results of two queries) but can also be used to implement disjunction (OR), conjunction (AND), and complement (NOT) in queries.

The `SourceSet` type is an internal type only which wraps the lowest level concrete types that can be interpreted as a set (in the mathematical sense) of entities, though perhaps only abstractly. The fundamental requirements for something to be a `SourceSet` are:
1. ability to test membership: `source_set.contains(entity_id)` (virtually always very efficient)
2. ability to iterate over elements: `source_set.iter().collect<Vec<PersonId>>()` (not always efficient)
3. iteration yields unique entity IDs, i.e. a particular entity ID is yielded at most once. (This turns out to be a natural property—we have it at zero cost for source sets.)

The `EntitySet` type lifts source sets to an algebra of sets that maintain these three properties.

An initial implementation can be naïve, but a sophisticated implementation will exploit the fact that both set membership and iteration involve a binary decision procedure on sentences of boolean variables (which represent membership tests on the underlying source sets). The theory for optimal evaluation of these sentences is very well developed. (Roughly speaking, we have a _Reduced Ordered Binary Decision Diagram_, ROBDD.)

It also makes sense for the `EntitySet` type to replace the concrete `IndexSet` type used in APIs that give access to result sets (currently only `with_query_result`).

I have a lot more notes on this that are still evolving.

PR #751, opened 2026-02-10
feat: ValueCountIndex. Fixes #678.

### LINKED ISSUE #678: `ValueCountIndex`: Light weight index for tracking counts of property values
> `query_people_count` is something we need to do *very* often (e.g. to calculate rates etc.) so it's worth making sure this is fast.

Allow clients to create light weight "count" indexes that keep track of the count of entities having a particular property value rather than the full set of all `EntityId`s. (If the property is indexed with a regular index, we already get this for free.)

See also #533.

RobertJacobsonCDC:
This PR implements a new index type, a `ValueCountIndex`, that keeps track of the _count_ of entities having each distinct property value rather than the _set_ of entities having each distinct property value. Despite significant additions to internal indexing infrastructure, there is only one significant addition to the public API: **`ContextEntityExt::index_property_counts<E: Entity, P: Property<E>>(&mut self)`**.

```rust
context.index_property_counts<Person, Age>();
```

- If `context` has no index for `Age`, a new `ValueCountIndex` is created.
- If `context` already has a `ValueCountIndex` for `Age`, this method does nothing.
- If `context` already has a `FullIndex` for `Age`, this method _retains this `Fullndex`_, as a `FullIndex` already provides the functionality of a `ValueCountIndex`.

The function of the existing `ContextEntityExt::index_property<E: Entity, P: Property<E>>(&mut self)` method is augmented in the following way:

```rust
context.index_property<Person, Age>();
```

- If `context` already has a `ValueCountIndex` for `Age`, the `ValueCountIndex` is _replaced_ with a `FullIndex`.

We currently have no mechanism to downgrade or remove an existing index.

# Issues and Questions

## Alternative Public API

Instead of having only `Context::index_property` and `Context::index_property_counts`, we could instead give users complete control over a property's index with a method `Context::set_property_index(&mut self, index_type: PropertyIndexType)`. This would allow client code to set the index to any of `enum PropertyIndexType{Unindexed, FullIndex, ValueCountIndex}` at any time.

### Adding More Index Types

I have opted to keep the type of the index set exposed to the client code as the `IndexSet<EntityId<E>>` type used internally by `FullIndex`. If other alternative index types are to be added, we will need to change the type of the "index set" returned by the internal `get_index_set` methods to be more generic, as presumably the different index types will have different ways of storing sets of entity IDs. A logical choice is to adapt the existing `SourceSet` type for this purpose. This type is already designed to be an interface over the abstract notion of a set of entities that provides an ability to test membership and convert to an iterator or an owned `Vec<EntityId<E>>`. This is only really relevant for `with_query_results`, as this is the only place we give client code direct access to the index set by design. Alternatively, the planned type `EntitySet` probably makes more sense here, as one can do more things with it than with a `SourceSet`, but the `EntitySet` type hasn't been implemented yet. See #746.

### DIFF
--- a/src/entity/context_extension.rs
+++ b/src/entity/context_extension.rs
@@ -2,7 +2,7 @@ use std::any::{Any, TypeId};
 
 use crate::entity::entity_set::EntitySetIterator;
 use crate::entity::events::{EntityCreatedEvent, PartialPropertyChangeEvent};
-use crate::entity::index::{IndexCountResult, IndexSetResult};
+use crate::entity::index::{IndexCountResult, IndexSetResult, PropertyIndexType};
 use crate::entity::property::Property;
 use crate::entity::property_list::PropertyList;
 use crate::entity::query::Query;
@@ -41,6 +41,12 @@ pub trait ContextEntitiesExt {
     /// The actual computation of the index is done lazily as needed upon execution of queries,
     /// not when this method is called.
     fn index_property<E: Entity, P: Property<E>>(&mut self);
+
+    /// Enables value-count indexing of property values for the property `P`.
+    ///
+    /// If the property already has a full index, that index is left unchanged, as it
+    /// already supports value-count queries.
+    fn index_property_counts<E: Entity, P: Property<E>>(&mut self);
 
     /// Checks if a property `P` is indexed.
     ///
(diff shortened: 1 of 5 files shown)


src/entity/context_extension.rs
```diff
@@ -233,21 +250,21 @@ impl ContextEntitiesExt for Context {
...
-                    // there are no entities having this property value.
+                    return;
+                }
+                IndexSetResult::Empty => {
                     let people_set = IndexSet::default();
                     callback(&people_set);
+                    return;
                 }
-                return;
+                IndexSetResult::Unsupported => {}
```
k88hudson-cfa: I wonder if we should put a warn in debug builds if you iterate over some arbitrary number of entities for something like `with_query_results` that indexes aren't supported and to use a different API (like count or sample)

k88hudson-cfa (approved):
Looks good! In terms of exposing this to users, I don't really see an alternative when optimization is important without a lot smarter static analysis / compiler type stuff, and debug tools are going to be even more important as we increase the complexity here.(i'll file something) I wonder if we could be smart about turning on some hints in debug mode?

src/entity/context_extension.rs
RobertJacobsonCDC: I do emit a warning in `with_query_results` if you attempt to use it with an unindex property.

PR #758, opened 2026-02-10
fix: Updated the example in the Ixa Book to use `event.current`

RobertJacobsonCDC:
Pointed out by @swo, we apparently forgot to update the example in the Ixa Book to revert `event.current_value` to `event.current`. Oops!

But the fact that this error was possible raises a couple of questions:

- Apparently we don't build and run this in CI. We should.
- The version of the `ixa` crate we use in the `Cargo.toml` of the example in the Ixa Book should match the release version. ...But what if we're on a prerelease?

For the second issue, we have a few options:

```toml
# Always point to the main branch of the repo
ixa = { git = "https://github.com/CDCgov/ixa", branch = "main" }

# Point to a specific version on crates.io
ixa = "2.0.0-beta1"

# Point to the relative path of the version of ixa the example is from
ixa = { path = "../../../.."}
```

We had been using the last option, but if we want this example to have the property that a new user can just copy and paste it into a new Rust project and have it run—which I personally think is a very valuable property—this isn't a good strategy.

I would prefer specifying a specific version on crates.io, but this breaks for pre-releases. For example, if we update the Ixa Book and its example code to work with a new release that has breaking changes, there just won't be a published release that is compatible yet. So then how do we test it in CI?

Pointing to the main branch of the repo is the least fragile solution, because this is what the Ixa Book is built from, and so it should always be in sync with the example code. But when we update the example code for breaking changes, the example will break in CI until we merge the changes into `main`.

**To do before merging:**

- [x] Choose how to reference the `ixa` dependency for the "first model" example in the ixa book
- [x] Build and run the example in CI

### DIFF
--- a/.github/workflows/check-test.yaml
+++ b/.github/workflows/check-test.yaml
@@ -38,5 +38,8 @@ jobs:
       - name: Run examples
         run: mise run examples
 
+      - name: Build book example
+        run: mise run test:book
+
       - name: Build docs
         run: mise run docs
--- a/docs/book/models/disease_model/Cargo.toml
+++ b/docs/book/models/disease_model/Cargo.toml
@@ -8,6 +8,6 @@ publish = false # Do not publish to the Crates.io registry
 
 [dependencies]
 csv = "1.3.1"
-ixa = { path = "../../../.."}
+ixa = { git = "https://github.com/CDCgov/ixa", branch = "main" }
 rand_distr = "0.5.1"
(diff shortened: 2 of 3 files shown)


k88hudson-cfa:
Hm, good question. I'd probably have it refer to main for now, in the future we could considered versioned books (there's tools for this but probably not worth it at the moment)

k88hudson-cfa (approved):

PR #764, opened 2026-02-11
chore: Faster precommit

k88hudson-cfa:
## TO TEST:

You need to uninstall the precommit hooks and install the new ones:
```
mise run hooks:clear
mise run hooks:install
```
Then git checkout -b test123, add a commit 
```

I tried working on making pre-commit faster, with a few things:

  - Replaced pre-commit with prek, a rust implementation that's a lot faster
  - Replaced markdownlint with rumdl (Rust-based markdown linter)
 - Added rust-toolchain.toml to avoid installs every time... I *think* this should just work 
  - Fixed lint errors for missing links in a few files that came up fwhen adding rumdl
  - Added a mise run `hooks:clear`
  - Added `build` in CI which apparently is better for clippy? 
  
I also removed the justfile since we don't use it anymore

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,7 +1,7 @@
 [tools]
 node = "latest"
 pnpm = "latest"
-prek = "latest"
+"cargo:prek" = "latest"
 "prettier" = "3.6.2"
 "npm:@commitlint/cli" = "19.7.1"
 "npm:@commitlint/config-conventional" = "19.7.1"
@@ -10,7 +10,7 @@ prek = "latest"
 "cargo:mdbook" = "0.4.52"
 "cargo:mdbook-callouts" = "0.2.2"
 "cargo:mdbook-inline-highlighting" = "1.0.0"
-rumdl = "0.1.18"
+"cargo:rumdl" = "0.1.18"
 
 
 [settings]
--- a/prek.toml
+++ b/prek.toml
@@ -1,3 +1,4 @@
+
 [[repos]]
 repo = "builtin"
 

### CODE REVIEW

File: mise.toml [hunk 1]
```diff
@@ -1,21 +1,28 @@
...
 "prettier" = "3.6.2"
-"npm:markdownlint-cli" = "0.45.0"
 "npm:@commitlint/cli" = "19.7.1"
 "npm:@commitlint/config-conventional" = "19.7.1"
 "cargo:wasm-pack" = "latest"
 "cargo:hyperfine" = "latest"
 "cargo:mdbook" = "0.4.52"
 "cargo:mdbook-callouts" = "0.2.2"
 "cargo:mdbook-inline-highlighting" = "1.0.0"
+rumdl = "0.1.18"
```
RobertJacobsonCDC: I got an error until I changed this to:

```toml
"cargo:rumdl" = "0.1.18"
```

mise.toml
```diff
@@ -113,32 +120,33 @@ depends = "build:hyperfine"
...
 [tasks.'check:rust']
+depends = "install-nightly"
 run = [
   # cargo check compiles everything without the final step of code generation
   # --all-targets ensures we include --lib --bins --tests --benches --examples
   "cargo check --all-targets --workspace --all-features",
-  "rustup component add --toolchain nightly-x86_64-unknown-linux-gnu rustfmt",
   "cargo +nightly fmt --check",
-  "cargo clippy --workspace --all-targets --all-features -- -D warnings",
+  "cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings",
```
RobertJacobsonCDC: So we weren't using nightly clippy?

RobertJacobsonCDC (changes requested):

mise.toml
k88hudson-cfa: Maybe the issue is if you had nightly installed anyway you were, but you weren't if you didn't!

mise.toml
```diff
@@ -1,21 +1,28 @@
 [tools]
 node = "latest"
 pnpm = "latest"
-"pre-commit" = "4.3.0"
+prek = "latest"
```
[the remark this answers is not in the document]
k88hudson-cfa: Hm, interesting. I wonder if this needs cargo too?

PR #787, opened 2026-02-19
docs: Tweaks to Your First Model

swo:
- DRY: use `#rustdoc_include` where possible
- Consistency in the population size constant
- Move toward using "entity" over "agent" -- This was not an exhaustive effort; just a target of opportunity thing.
- Make the "this isn't SIR" explanation a bit clearer and more visible
- Clarify the model setup
  - Explain how the model as initially described (each person has an independent time to infection) maps onto the implementation here (there are infection attempts)
  - An alternative would be to pre-schedule all the infections, as per #788, although I suspect that's not great for this first example?
  - Draw inter-infection-attempt times from `Exp::new(FORCE_OF_INFECTION * POPULATION)`. The existing approach definitely works: if $X \sim \mathrm{Exp}(\lambda)$, then $\frac{X}{a} \sim \mathrm{Exp}(a \lambda)$, but this approach is a little more intuitive and it avoids the precision loss thing.
- Gave some hints in the exercises section, and made the 3rd exercise align better with the foodborne illness theme.
- Ignore `incidence.csv` produced by `first_model`

### DIFF
--- a/docs/book/models/disease_model/.gitignore
+++ b/docs/book/models/disease_model/.gitignore
@@ -3,3 +3,5 @@
 
 /target
 Cargo.lock
+
+*.csv
--- a/docs/book/models/disease_model/src/main.rs
+++ b/docs/book/models/disease_model/src/main.rs
@@ -1,14 +1,16 @@
+// ANCHOR: header
 mod incidence_report;
 mod infection_manager;
 mod people;
 mod transmission_manager;
 
-use ixa::{Context, error, info, run_with_args};
+use ixa::{error, info, run_with_args, Context};
 
 static POPULATION: u64 = 100;
 static FORCE_OF_INFECTION: f64 = 0.1;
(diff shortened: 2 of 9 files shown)


RobertJacobsonCDC:
> * Move toward using "entity" over "agent" -- This was not an exhaustive effort; just a target of opportunity thing.

I defer to the experts for the language that is most appropriate to our audience here.

But my understanding is that not all entities are agents, at least as the word entity is used in Ixa. We might have the `Person` entity that serves the role of agent, and then we might have other entities that serve as data sources but that aren't technically agents. A `Setting` entity might hold geographical data, an $\alpha$ influencing infection hazard, etc.; A `Region` entity might hold information about region-specific interventions / policy. Right?

RobertJacobsonCDC (approved):
Excellent! Thank you so much!

PR #833, opened 2026-03-30
chore:  remove tuple queries

### LINKED ISSUE #798: Follow-up: remove tuple queries
We decided as part of  #793 we would remove tuple queries and use `with!` instead

cdc-as81:
## Summary

This PR introduces entity-scoped query/property-bundle construction with `with!(Entity, ...)` and `Entity` while preserving compatibility with the legacy tuple-based query API.

## What Changed

- Added and documented `with!(Entity, ...)` as the preferred public constructor for:
  - entity initialization bundles passed to `add_entity(...)`
  - entity-scoped queries
- Kept `EntityPropertyTuple<E, T>` as the public wrapper for entity-scoped property bundles
- Moved tuple query behavior behind an internal `QueryTuple<E>` trait in `query_impls.rs`
- Added `Query::is_empty_query()` so whole-population queries can be handled explicitly
- Made entity ZSTs implement `Query<E>` and `PropertyList<E>` so callers can use `Person` / `with!(Person)` for whole-population operations
- Updated query/count/sample code paths to use `is_empty_query()` instead of special-casing `()`
- Updated docs and tests to use the entity-scoped forms consistently

## Backward Compatibility

Legacy query forms are still supported:

- Bare query tuples like `(Age(42),)` and `(Age(42), RiskCategory::High)` continue to work
- Empty tuple queries `()` are still supported where type inference is available or the entity type is specified explicitly

This keeps older code working while allowing newer call sites to use the clearer `with!(Entity, ...)` form.

## Why

The previous API relied heavily on raw tuples and `()` for queries, which made entity scope implicit. This change makes the public API more explicit and readable:

- `with!(Person, Age(42))` is clearly a `Person` query
- `with!(Person)` and `Person` clearly represent whole-population queries
- Tuple-specific machinery remains internal, while compatibility is preserved

## Testing

Added and updated tests to cover:

- `with!(Entity)` empty-query behavior
- `with!(Entity, ...)` single- and multi-property queries
- Direct use of `EntityPropertyTuple`
- Continued support for legacy tuple queries
- Continued support for legacy empty tuple whole-population queries


RobertJacobsonCDC (changes requested):
This PR changes all examples and tests to use the `with!` macro consistently for

- `Query` instances (`query_entity_count`, `with_query_results`, `match_entity`, `sample_entity`)
- `PropertyList` instances (`add_entity`)

It also introduces the private trait `QueryTuple`, which mirrors `Query`, though it's not clear what its purpose is. `QueryTuple` is implemented for tuples up to size 20.

What this PR doesn't do is remove tuple queries. `Query` is still implemented for `()` and for tuples of properties up to length 20. I think the intention of the issue was that we'd remove the `Query` impl for all tuples and require client code to use the `with!` macro (and thus `EntityPropertyTuple`). 

Because you already have a blanket `impl<E: Entity, T: QueryTuple<E>> Query<E> for EntityPropertyTuple<E, T>`, you can achieve this by just removing the implementation

```rust
            impl<
                E: Entity,
                #(
                    T~N : Property<E>,
                )*
            > Query<E> for (
                #(
                    T~N,
                )*
            ) {...}
```

from the `impl_query!` macro definition. Then `QueryTuple<E>` would make sense: it allows `EntityPropertyTuple<E, T>` to implement `Query` without the tuple `T` implementing `Query`. 

We have a blanket implementation of `PropertyList` for `EntityPropertyTuple<E, T>` whenever `T: PropertyList`. We could also require users to use the `with!` macro for property lists (adding new entities), in which case we would remove the implementation of `PropertyList` for tuples up to size 20 as well. We could use the same hypothetical mechanism as for tuples: introduce a private `PropertyListTuple` trait that mirrors `PropertyList` and provide a blanket `impl<E: Entity, T: PropertyListTuple<E>> PropertyList<E> for EntityPropertyTuple<E, T>` that delegates to `PropertyListTuple<E>`.

RobertJacobsonCDC:
Closing in favor of #856.

PR #842, opened 2026-04-07
chore: turn off profiling by default

### LINKED ISSUE #818: Turn off profiling by default
Given profiling tools add overhead, I think we should consider having them off by default

cdc-as81:


RobertJacobsonCDC (changes requested):
Double check me on this, but I think the tests in `profiling::reporting` are still relevant if the reporting feature is disabled. I think the API being exercised is still callable in that case.

RobertJacobsonCDC (approved):

PR #844, opened 2026-04-14
feat: Optimize `set_property` if dependent does not need partial change-event processing

RobertJacobsonCDC:
This PR makes `set_property` faster when property-change machinery is not actually needed.

Previously, `set_property` always created `PartialPropertyChangeEvent`s for the property being written and for every derived dependent. These hold the previous value of their respective properties so that property change events can be emitted and indexes and value change counters updated. This PR adds a cheap per-property check and does not create partial change events for properties that have no `PropertyChangeEvent` subscribers, no value-change counters, and no indexes to maintain.

To support that, the change adds:

- a way for `Context` to check whether a given event type has any registered handlers;
- a type-erased property-store hook that answers whether a property needs change processing because it has subscribers, counters, or an index;
- a `set_property` fast path that skips partial-event creation for unaffected properties.

The benchmarks were also expanded with a steady-state mixed benchmark that keeps one `Context` alive, flushes callbacks between measured chunks, and exercises the three relevant cases at once: indexed dependent, counted dependent, and handler-backed dependent. That gives us a way to measure both the new best case and the pessimistic path without distorting results with unbounded callback queue growth.

# Local Benchmarks

Local Criterion benchmarks in the `ixa` repo:

```
set_property/set_property_no_dependents
                        time:   [19.035 ns 19.063 ns 19.096 ns]
                        change: [−42.210% −40.648% −39.368%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe
set_property/set_property_three_dependents
                        time:   [20.509 ns 20.522 ns 20.536 ns]
                        change: [−75.287% −74.613% −74.072%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  6 (6.00%) high severe
set_property/set_property_three_dependents_mixed
                        time:   [45.268 µs 45.355 µs 45.436 µs]
                        change: [−5.4784% −5.0960% −4.7522%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
```

The improvement in the last case is spurious: that's the "control" in which we actually can't avoid doing the work. The fact that it isn't a regression, though, is good, because that says that the extra checking this optimization requires is negligible.

The result of a few runs of `ixa-epi-covid`'s comparison benchmarks (`make bench-compare`) on my local machine:

| **Benchmark**           | **Base (mean)** | **Current (mean)** | **Change**    | **Base Var** | **Base Std Dev** | **Current Var** | **Current Std Dev** |
| ----------------------- | --------------- | ------------------ | ------------- | ------------ | ---------------- | --------------- | ------------------- |
| init/population/10k     | 10.093 ms       | 8.1505 ms          | -19.2% faster | 0.04953 ms²  | 0.2226 ms        | 0.01145 ms²     | 0.1070 ms           |
| init/population/100k    | 107.85 ms       | 89.923 ms          | -16.6% faster | 3.6279 ms²   | 1.9047 ms        | 2.1749 ms²      | 1.4748 ms           |
| execute/population/10k  | 80.725 ms       | 82.416 ms          | +2.1% slower  | 1.0578 ms²   | 1.0285 ms        | 1.4813 ms²      | 1.2171 ms           |
| execute/population/100k | 1.2950 s        | 1.2058 s           | -6.9% faster  | 0.00155 s²   | 0.0394 s         | 0.00526 s²      | 0.0725 s            |

### DIFF
--- a/ixa-bench/criterion/set_property.rs
+++ b/ixa-bench/criterion/set_property.rs
@@ -1,12 +1,19 @@
+use std::cell::RefCell;
 use std::hint::black_box;
 
-use criterion::{criterion_group, criterion_main, Criterion};
+use criterion::{criterion_group, criterion_main, BatchSize, Criterion};
 use ixa::prelude::*;
+use ixa::{impl_derived_property, impl_property};
 
 define_entity!(Person);
 
 define_property!(struct IndependentValue(u64), Person, default_const = IndependentValue(0));
 define_property!(struct BaseValue(u64), Person, default_const = BaseValue(0));
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, serde::Serialize, Hash)]
+struct MixedBaseValue(u64);
+
+impl_property!(MixedBaseValue, Person, default_const = MixedBaseValue(0));
 
 define_derived_property!(
     struct DerivedLowBit(bool),
@@ -41,6 +48,36 @@ define_derived_property!(
     }
 );
 
+define_derived_property!(
+    struct MixedDerivedIndexed(bool),
+    Person,
+    [MixedBaseValue],
+    [],
+    |base| {
+        let base: MixedBaseValue = base;
+        MixedDerivedIndexed(base.0 & 1 == 0)
+    }
+);
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, serde::Serialize, Hash)]
+struct MixedDerivedCounted(u8);
+
+impl_derived_property!(MixedDerivedCounted, Person, [MixedBaseValue], [], |base| {
+    let base: MixedBaseValue = base;
+    MixedDerivedCounted((base.0 % 4) as u8)
+});
+
+define_derived_property!(
+    struct MixedDerivedHandled(bool),
+    Person,
+    [MixedBaseValue],
+    [],
+    |base| {
+        let base: MixedBaseValue = base;
+        MixedDerivedHandled(base.0.trailing_zeros() == 0)
+    }
+);
+
 pub fn criterion_benchmark(c: &mut Criterion) {
     let mut group = c.benchmark_group("set_property");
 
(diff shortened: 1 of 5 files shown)


ixa-bench/criterion/set_property.rs
```diff
@@ -66,6 +103,43 @@ pub fn criterion_benchmark(c: &mut Criterion) {
...
+            || {
+                context.borrow_mut().execute();
+            },
+            |_| {
+                let mut context = context.borrow_mut();
+                for value in 0..256 {
+                    context.set_property(black_box(person), MixedBaseValue(black_box(value)));
+                }
+            },
+            BatchSize::SmallInput,
```
k88hudson-cfa, line 139: What's the reason for `SmallInput`?

k88hudson-cfa (approved):
Nice! Looks good to me. Thanks for adding the benchmark

ixa-bench/criterion/set_property.rs
RobertJacobsonCDC: It's an obscure Criterion detail:

> `SmallInput` indicates that the input to the benchmark routine (the value returned from the setup routine) is small enough that millions of values can be safely held in memory. Always prefer `SmallInput` unless the benchmark is using too much memory.

The `iter_batched` API is designed to let you construct input for the timed iteration without having the construction itself as part of the measured time. I am not using it that way, though. Instead I'm using it to periodically flush the callback queue so it doesn't grow pathologically in an attempt to prevent memory allocation adding noise to the benchmark. So in my case, the "input" is `()` (unit), which is zero sized.

PR #847, opened 2026-04-15
docs: network-hhmodel example: Separate contact from disease progression

swo:
- Don't use `ixa::network`, but instead instantiate edges as entities. This allows us to attach the relative rate of transmission to each edge as a property, rather than encoding the edge types as Rust types.
- `seir.rs` looks for the infected people, then asks `network.rs` who they contact via `get_contacts()`. This is a nicer separation of concerns, so that the infection doesn't need to know about the contact structure.
- Make the period of the periodic plan explicit, so that `get_contacts()` needs a `duration` argument, that says what contacts will happen over what time period.
- Simplify how the relative rate of transmission gets passed around.
  - Don't specify a "reduction" (that divides); instead specify a relative rate (that gets multiplied).
  - Pass this in certain `network::init()` to simplify some test interfaces.
  - Attach the relative rate to each edge as a property.

### DIFF
--- a/examples/network-hhmodel/main.rs
+++ b/examples/network-hhmodel/main.rs
@@ -41,7 +41,7 @@ fn initialize(context: &mut Context) {
         .clone();
 
     // Load network
-    network::init(context, parameters.between_hh_transmission_reduction);
+    network::init(context, parameters.between_hh_transmission_rr);
 
     // Initialize incidence report
     incidence_report::init(context).unwrap();
--- a/examples/network-hhmodel/main.rs
+++ b/examples/network-hhmodel/main.rs
@@ -7,6 +7,8 @@ mod parameters;
 mod seir;
 use std::path::PathBuf;
 
+use parameters::Parameters;
+
 define_entity!(Person);
 define_rng!(MainRng);
 
@@ -27,14 +29,19 @@ fn initialize(context: &mut Context) {
     context.init_random(1);
 
     // Load people from csv and set up some base properties
-    let people = loader::init(context);
+    loader::init(context);
 
     // Load parameters from json
     let file_path = example_dir().join("config.json");
     context.load_global_properties(&file_path).unwrap();
 
+    let parameters = context
+        .get_global_property_value(Parameters)
+        .unwrap()
+        .clone();
+
     // Load network
-    network::init(context, &people);
+    network::init(context, parameters.between_hh_transmission_reduction);
 
     // Initialize incidence report
     incidence_report::init(context).unwrap();
--- a/examples/network-hhmodel/main.rs
+++ b/examples/network-hhmodel/main.rs
@@ -43,6 +43,6 @@ fn initialize(context: &mut Context) {
     let to_infect: Vec<PersonId> = vec![context.sample_entity(MainRng, Person).unwrap()];
 
     #[allow(clippy::vec_init_then_push)]
-    seir::init(context, &to_infect);
+    seir::init(context, &to_infect, 1.0);
     context.execute();
 }
--- a/examples/network-hhmodel/main.rs
+++ b/examples/network-hhmodel/main.rs
@@ -41,7 +41,7 @@ fn initialize(context: &mut Context) {
         .clone();
 
     // Load network
-    network::init(context, parameters.between_hh_transmission_rr);
+    network::init(context, parameters.relative_rate);
 
     // Initialize incidence report
     incidence_report::init(context).unwrap();
@@ -51,5 +51,4 @@ fn initialize(context: &mut Context) {
 
     #[allow(clippy::vec_init_then_push)]
     seir::init(context, &to_infect, 1.0);
-    context.execute();
 }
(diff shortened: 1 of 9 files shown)


RobertJacobsonCDC:
Awesome! I think we've all felt the need for someone with modeling expertise to take a close look at our examples for quite some time now, so it's great that you're doing this. If you want to look over any other of the examples—or even add new ones that you think are conspicuously missing—definitely feel free to! But obviously don't feel obligated.

I'll look into the failing lints and get back to you. Sometimes they feel like a moving target...

RobertJacobsonCDC:
I think it's not failing because of the lints. I think it's failing because the example never terminates: `Cancelled after 365m`.

swo:
@bbbruce -- Robert asked for another pair of eyes on this example. Would you take a look? I could do a more full-blown analysis to check for correctness if we consider the prior state to be a gold standard.

swo:
@bbbruce I reworked a little bit as per our discussion:

- Don't use network module; instead define an `Edge` entity
- I define `Node1` and `Node2` as two separate properties. (My first instinct was to define `Nodes(node1, node2)`, but I couldn't figure out how to `context.query()` for only part of a tuple property, like `.query(with!(Edge, Nodes(_, some_person)))`.)
- I opted for directional edges, which make the queries easier (rather than unidirectional edges, which requires half the entities but makes the queries more complicated).
- I'm having trouble figuring out how to test this. I'd like to say "look for person with data ID 826. They should have 5 contacts." That requires me to (1) query a `Person`, (2) get their `PersonId`, and (3) use that `PersonId` to find the relevant edges. Step 2 has me stumped. Any ideas?

examples/network-hhmodel/network.rs
```diff
@@ -1,111 +1,160 @@
...
+use crate::parameters::Parameters;
 use crate::{Person, PersonId};
 
-define_edge_type!(struct Household, Person);
-define_edge_type!(struct AgeUnder5, Person);
-define_edge_type!(struct Age5to17, Person);
+// ixa properties must implement `Eq` and `Hash`, but `f64`
+// does not. This example manually implements that logic manually.
+#[derive(Copy, Clone, Serialize, Debug)]
+pub struct RR(f64);
```
RobertJacobsonCDC: I'd give it a meaningful name considering this is an example, something like `RelativeTransmissionRate`. Also for the parameters to `network::init` and `network::create_household_networks`, I'd rename `rr` to `relative_transmission_rate`.

examples/network-hhmodel/main.rs
```diff
@@ -43,6 +50,6 @@ fn initialize(context: &mut Context) {
     let to_infect: Vec<PersonId> = vec![context.sample_entity(MainRng, Person).unwrap()];
 
     #[allow(clippy::vec_init_then_push)]
-    seir::init(context, &to_infect);
+    seir::init(context, &to_infect, 1.0);
     context.execute();
```
RobertJacobsonCDC: This `initialize` is only called within the initializer closure passed to `run_with_args`. But `run_with_args` automatically calls `context.execute()` after calling the initializer closure. Calling it here causes it to be run twice.

examples/network-hhmodel/network.rs
```diff
@@ -1,111 +1,160 @@
...
 
-fn load_edge_list<ET: EdgeType<Person>>(context: &mut Context, file_name: &str, inner: ET) {
+// Assert there is only one person with data ID `id`, then get their entity ID
+fn get_entity_id_by_data_id(context: &mut Context, id: u16) -> PersonId {
+    let v: Vec<PersonId> = context.query(with!(Person, Id(id))).into_iter().collect();
+    assert_eq!(v.len(), 1);
+    v[0]
+}
+
+fn load_edge_list(context: &mut Context, file_name: &str, rr: RR) {
```
RobertJacobsonCDC: `rr` &rarr; `relative_transmission_rate`

examples/network-hhmodel/network.rs
```diff
@@ -1,111 +1,160 @@
...
 
-    // Add U5 edges from csv
-    load_edge_list(context, "AgeUnder5Edges.csv", AgeUnder5);
+    contacts
+}
 
-    // Add U18 edges from csv
-    load_edge_list(context, "Age5to17Edges.csv", Age5to17);
+// `rr`: relative rate of transmission between (vs. within) households
+pub fn init(context: &mut Context, rr: f64) {
```
RobertJacobsonCDC: `rr` &rarr; `relative_transmission_rate`

RobertJacobsonCDC (changes requested):

swo:
@RobertJacobsonCDC A few changes this round:

- Call it `relative_rate` and make it clear in the readme what that means
- Clarify some of the language and math around what the relative rate is multiplying
- Simplify function interfaces: mostly pass the relative rate float, rather than the `RelativeRate` struct
- Use the new system for implementing equality for float properties

RobertJacobsonCDC (approved):

Issue #843, opened 2026-04-07
Type aliases break multi-property canonicalization


RobertJacobsonCDC:
When a property used in an indexed multi-property is changed from a distinct newtype to a type alias, composite queries can silently stop matching even though the underlying single-property queries still work. In our case, `SettingCode` was changed from a newtype to type `SettingCode = FIPSCode`, while `Setting` still defines and indexes the multi-property `(SettingCategory, SettingCode)`. After this change, queries by `SettingCode` alone succeed, but queries by `(SettingCode, SettingCategory)` return no results for entities that definitely exist. The likely cause is that ixa’s multi-property canonicalization/index registration uses source-level identifiers from the macro input (e.g. `SettingCode`), while query hashing/lookup uses `Property::name()`/type identity of the concrete property type, which for a type alias resolves to the underlying type (`FIPSCode`). That creates a mismatch between the registered composite property and the runtime composite query key, causing indexed multi-property lookups to return empty instead of matching or falling back to intersecting the component-property indices.

It's not clear how to fix this, or if it's even fixable. 

- The type-level information just isn't available in the macro, because the macro fundamentally acts on syntax, not on fully realized types. We could abandon compile-time canonicalization in favor of some runtime mechanism, but the point of using a compile-time mechanism is that it is performant, and any runtime solution is bound to be slower.
- We could just tell people not to use type aliases for properties (at least not with multi-properties), but I don't see a way to warn people if and when they do the wrong thing.

k88hudson-cfa:
The two options we thought of are:
- getting rid of auto-reordering (forcing people to use the same order)
- type aliases continuing to silently fail with empty results

We don't love either of these

k88hudson-cfa:
Maybe we could implement some kind of `get_name` and check it?

RobertJacobsonCDC:
Resolve by #937.

PR #852, opened 2026-04-20
fix: reject type aliases in define_multi_property! (#843)

k88hudson-cfa:
This isn't a great solution but would give a better error message. Incidentally, it also saves some work on every call to name() so there are some benchmark improvements. Thoughts?

### DIFF
--- a/src/entity/property.rs
+++ b/src/entity/property.rs
@@ -39,15 +39,36 @@ pub enum PropertyInitializationKind {
 pub trait AnyProperty: Copy + Debug + PartialEq + Serialize + 'static {}
 impl<T> AnyProperty for T where T: Copy + Debug + PartialEq + Serialize + 'static {}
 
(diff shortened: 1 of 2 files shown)


src/entity/property.rs
```diff
@@ -39,15 +39,36 @@ pub enum PropertyInitializationKind {
 pub trait AnyProperty: Copy + Debug + PartialEq + Serialize + 'static {}
 impl<T> AnyProperty for T where T: Copy + Debug + PartialEq + Serialize + 'static {}
 
+/// `const fn` string equality — `==` on `&str` isn't `const` on stable.
```
RobertJacobsonCDC, line 42: Ugh, there is so much `const` that needs stabilization!

RobertJacobsonCDC (approved):
Yes! This might not be ideal, but it is much better than the status quo.

PR #854, opened 2026-04-23
feat: `SourceSet::PopulationRange` subsumes `Empty`, `Entity`, and `WholePopulation` variants

### LINKED ISSUE #824: Add a `SourceSet` variant for a range of `EntityId`s
We have these `SourceSet` variants:

```rust
/// Represents the set of `EntityId<E>`s for which a particular `Property` has a particular value.
pub(crate) enum SourceSet<'a, E: Entity> {
    Empty,
    Population(usize),
    Entity(EntityId<E>),
    IndexSet(&'a IndexSet<EntityId<E>>),
    PropertySet(BxPropertySource<'a, E>),
}
```

We should have a `SourceSet::PopulationRange(Range<usize>)` to represent any contiguous range of `EntityId`s. This single variant can also replace all uses of `Empty`, `Population`, and `Entity` (the singleton set). `SourceSet` thus simplifies to:

```rust
/// Represents the set of `EntityId<E>`s for which a particular `Property` has a particular value.
pub(crate) enum SourceSet<'a, E: Entity> {
    PopulationRange(Range<usize>),
    IndexSet(&'a IndexSet<EntityId<E>>),
    PropertySet(BxPropertySource<'a, E>),
}
```

RobertJacobsonCDC:
# Added `SourceSet::PopulationRange` Variant

This PR consolidates contiguous entity-ID sources around `PopulationRange` and carries that representation through the `entity_set` stack. `SourceSet` no longer has separate empty, whole-population, or singleton variants; those cases are now expressed canonically as ranges, with helper constructors for empty, singleton, and general population ranges. The iterator side was simplified in the same direction: `SourceIterator` now relies on the range-backed path for empty and whole-population iteration, and `EntitySetIteratorInner` no longer has its own dedicated empty variant.

On top of that representation cleanup, `EntitySet` now has basic range-aware set algebra. Unions of overlapping or adjacent ranges collapse to a single range, intersections of ranges collapse to their overlap, and range differences simplify when the result is still empty or one contiguous range; split differences continue to use the generic expression path. The work also removed the older helper-driven empty/universal/singleton reductions that depended on the deleted variants and replaced them with a single `as_range()` helper used only for the new range optimizations.

The branch also updates downstream query construction and tests to match the new canonical form. Empty queries and full-population query results now construct range-backed sources, docs/comments in `entity_set` were refreshed to describe the new model, and the `entity_set`/iterator test coverage was rewritten and expanded to cover canonical empty/singleton/full-population ranges, optimized range operations, and the remaining non-optimized split-difference behavior.

This PR replaces #826.

### DIFF
--- a/src/entity/context_extension.rs
+++ b/src/entity/context_extension.rs
@@ -451,8 +451,8 @@ impl ContextEntitiesExt for Context {
         // Special case the empty query, which creates a set containing the entire population.
         if query.type_id() == TypeId::of::<()>() {
             warn!("Called Context::with_query_results() with an empty query. Prefer Context::get_entity_iterator::<E>() for working with the entire population.");
-            callback(EntitySet::from_source(SourceSet::Population(
-                self.get_entity_count::<E>(),
+            callback(EntitySet::from_source(SourceSet::PopulationRange(
+                0..self.get_entity_count::<E>(),
             )));
             return;
         }
--- a/src/entity/entity_set/entity_set.rs
+++ b/src/entity/entity_set/entity_set.rs
@@ -5,11 +5,12 @@
 //! is constructed eagerly but evaluated lazily: membership tests ([`contains`]) walk
 //! the tree on demand, and iteration is deferred to [`EntitySetIterator`].
 //!
-//! Construction methods apply algebraic simplifications (e.g. `A ∪ ∅ = A`,
-//! `A ∩ A = A`) and reorder operands by estimated size to improve short-circuit
-//! performance.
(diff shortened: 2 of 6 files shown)


RobertJacobsonCDC:
Trivial optimizations are implemented when both sets are `SourceSet::PopulationRange`, but it makes sense to include trivial optimizations for, say, an arbitrary set `A` and the empty set, for example.

cdc-as81 (approved):
LGTM

PR #848, opened 2026-04-17
feat: implement `Eq` and `Hash` for `Property`

RobertJacobsonCDC:
This PR completes the transition of property indexing and query lookup from serialized surrogate hashes to standard Rust `Eq`/`Hash` semantics. Property values and canonical values are now first-class map keys, the property macros support both normal derived implementations and manual float-style equality/hash behavior, and shared multi-property indexes use allocation-free canonical-value reconstruction instead of byte serialization.

The PR is split into commits corresponding to the following phases:

**Phase 0: Added baseline benchmarks for property hashing, equality, and value count indexes**

- Added `ixa-bench/criterion/property_semantics.rs`.
- No code changes in `ixa` proper.
- `mise run bench:create '' main` on this commit to establish a baseline for comparison.

**Phase 1: Make Properties `Eq + Hash`**

- Tightened the property contract so both property values and `Property::CanonicalValue` implement `Eq` and `Hash`. 
- Extended `define_property!` with `impl_eq_hash = Eq | Hash | both | neither`.
- Kept the default case simple: ordinary properties still derive `Eq` and `Hash` automatically.
- Added macro-generated `Eq`/`Hash` support for types that cannot use ordinary derives.
- Implemented generated `Hash` via `rkyv` streaming into a hasher with `HasherWriter`. Does not allocate.
- Implemented generated equality via `rkyv` using fixed-size archived-byte comparison with `EqualityBufferWriter`. Does not allocate.
- Re-exported `rkyv` for macro-generated code and updated property/macro docs so the required trait list now includes `Eq` and `Hash`.

**Phase 2: Switch Hashing To Use `Hash`**

- Rebased property hashing on the crate’s deterministic `Hash` path instead of `bincode`/`serde`.
- Centralized deterministic `u128` hashing in `hashing::one_shot_128`.
- Removed the old serialized-hash implementation and the `bincode` re-export/dependency from the property/indexing path.
- Updated multi-property lookup so it no longer serializes components into `Vec<u8>` just to compute a lookup hash.
- Added `ixa-derive` support for reconstructing canonical multi-property values from sorted query parts with `canonical_from_sorted_query_parts_closure`.

**Phase 3: Remove Hash-Only Plumbing**
- Converted `FullIndex` to `HashMap<P::CanonicalValue, IndexSet<EntityId<E>>>`.
- Converted `ValueCountIndex` to `HashMap<P::CanonicalValue, usize>`.
[...]

PR #856, opened 2026-04-25
chore: migrate all queries and `add_entity` calls to use `with!` syntax.

RobertJacobsonCDC:
This PR removes support for passing naked tuples of property values to the public query and `add_entity` APIs. Users now express property bundles with `with!(Entity, ...)`, and use the entity type itself, such as `Person`, when they want the whole population or default initialization.

The query implementation keeps the tuple machinery internally, but the public API is narrowed through a new `Query` marker trait. The old `Query` trait was renamed to `QueryInternal`, its unnecessary `Copy` bound was removed, and public query methods now accept only the public `Query` surface. `EntityPropertyTuple<E, Q>` and entity ZSTs implement that marker; naked tuples do not.

Entity initialization follows the same public-facing rule. `add_entity` now accepts `PropertyInitializationList<E>`, which is implemented for `with!(Entity, ...)` bundles and for the entity type itself. The existing `PropertyList<E>` tuple machinery remains available for internal initialization work and type-level property lists used by value-change counting.

The branch also migrates examples, tests, benches, and docs away from naked tuple values. User-facing documentation now consistently shows `with!(Person, Age(25))` for property-filtered queries or initialization, and `Person` for whole-population queries or default initialization.

Commits on this PR:

1. Migrate existing code to the new with!(Entity, ...) style before changing trait bounds, so call-site churn is isolated from API narrowing.
2. Rename the old Query trait to QueryInternal and remove its unnecessary Copy bound.
3. Introduce the new public Query marker trait and narrow public query APIs so with!(Entity, ...) and Entity work, but naked tuples do not.
4. Introduce PropertyInitializationList and narrow add_entity so initialization follows the same public rule: with!(Entity, ...) for property values, Entity for defaults.
5. Update user-facing docs to explain the new with! and entity-type syntax without exposing the internal trait split.

This PR is stacked on top of #848 and supersedes #833.

### DIFF
--- a/ixa-bench/src/reference_sir/sir_ixa.rs
+++ b/ixa-bench/src/reference_sir/sir_ixa.rs
@@ -52,13 +52,13 @@ impl InfectionLoop for Context {
         self.get_data(ModelStatsPlugin)
     }
     fn infected_people(&self) -> usize {
-        self.query_entity_count::<Person, _>((InfectionStatus::Infectious,))
+        self.query_entity_count::<Person, _>(with!(Person, InfectionStatus::Infectious))
     }
     fn random_person(&mut self) -> Option<PersonId> {
-        self.sample_entity(NextPersonRng, ())
+        self.sample_entity(NextPersonRng, Person)
     }
     fn random_infected_person(&mut self) -> Option<PersonId> {
-        self.sample_entity(NextPersonRng, (InfectionStatus::Infectious,))
+        self.sample_entity(NextPersonRng, with!(Person, InfectionStatus::Infectious))
     }
     fn infect_person(&mut self, p: PersonId, t: Option<f64>) {
         if self.get_property::<_, InfectionStatus>(p) != InfectionStatus::Susceptible {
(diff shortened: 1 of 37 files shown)


ixa-bench/src/reference_sir/sir_ixa.rs
```diff
@@ -52,13 +52,13 @@ impl InfectionLoop for Context {
         self.get_data(ModelStatsPlugin)
     }
     fn infected_people(&self) -> usize {
-        self.query_entity_count::<Person, _>((InfectionStatus::Infectious,))
+        self.query_entity_count::<Person, _>(with!(Person, InfectionStatus::Infectious))
```
k88hudson-cfa: Can't you remove the turbofish here now?

ixa-bench/src/reference_sir/sir_ixa.rs
```diff
@@ -149,13 +149,13 @@ impl InfectionLoop for Context {
 
         // Set up population
         for _ in 0..population {
-            self.add_entity::<Person, _>(()).unwrap();
+            self.add_entity::<Person, _>(Person).unwrap();
```
k88hudson-cfa: Here too

k88hudson-cfa (approved):
Looks good to me!

PR #858, opened 2026-04-28
feat: add `Indexable` trait and `IndexableMap`

k88hudson-cfa:
Adds a `Indexable` trait/ `IndexableMap` data structure and a usage example in our reference benchmarks. 

Things that could be improved:
* Better name for `IndexableMap`?
 * I could implement this as a sealed trait so as not to expose the `index()`, but I can't think of a way that people could implement it otherwise on other things. Do we think (a) making it implementable on other things is important for the public api and (b) is there any other clever way to do that that I haven't thought about?


RobertJacobsonCDC:
## General Editorial Comments

Since you still need to `impl Index<KeyType> for ContainerType` (or some getter method on `ContainerType`), the advantage of a key-side trait is that this `impl` can be a blanket implementation: `impl<K: KeyTrait> Index<K> for ContainerType`. If you have $m$ concrete `KeyType`s and $n$ concrete `ContainerTypes`, you'd need:

- $m\cdot n$ explicit `impl Index<KeyType> for ContainerType`, but only
- $m$ explicit `impl KeyTrait for KeyType` and $n$ explicit `impl<K: KeyTrait> Index<K> for ContainerType` blanket implementations, or $m + n$ total `impl`s.

For $m, n \in \{1, 2\}$ or for $m=1$ and $n\in \mathbb{N}$, it's actually more efficient to do the explicit `impl`s for the concrete types. In our case, we can get away with one blanket impl: `impl<E: Entity> KeyTrait for EntityId<E>`. So $m = 1$. We also have just one container type, at least for now.

If we decide we need a key-side trait, I think I'd do it this way:

- Name it `AsIndex`. The container is the thing _being_ indexed, so the container is *indexable*. The key is the *index*, or maybe the *indexer*, the thing doing the indexing. I don't know. Naming is hard.
- Probably make the trait unimplementable by client code, but this is probably not important so long as they can't provide implementations for types that we provide implementations for, which they can't without getting a compiler error anyway.
- Provide a blanket `impl<K: Into<usize>> Indexable for K` instead of individual implementations. (Things like `EntityId<E>` do not already implement `Into<usize>`.) This gets rid of the macro. I personally wouldn't bother implementing it for other numeric types. After all, the standard library doesn't go even this far. In fact, I might even just implement it for `usize` itself and not bother with the blanket impl.
- Provide a blanket `impl<E: Entity> Indexable for EntityId<E>`.
- The `ContainerType` should only be indexable by one `KeyType`. (You already have this.)

## Specific Comments About This Implementation

The semantics of `iter` are odd to me. I get that you're wanting a `HashMap`-like iter API, but `IndexableMap<I, V>` isn't indexable by `usize` in general (unless `I = usize`), so I don't see the utility in yielding the internal slot index. As it is, the internal slot index is an internal implementation detail. It makes more sense to me to have a mechanism to construct an `I` from a `usize` and yield `(I, &V)` pairs. (I would also mirror the `HashMap<K, V>` API and replace `indices` with `keys`.) 

Alternatively, if you are committed to not ever constructing a value of `I`—which is actually reasonable—why not also have a `Vec`-like API? You don't have to worry about producing invalid `I` values, so you could allow `push` and `extend` methods to shove as many values into the container as the client wants. 

Apart from the above, the current implementation constructs `(index, value)` pairs and then immediately discards the index. The compiler might optimize that away, but my inclination is to just do something like:

```rust
pub fn values(&self) -> impl Iterator<Item = &V> {
    self.entries.iter().filter_map(Option::as_ref)
}

pub fn values_mut(&mut self) -> impl Iterator<Item = &mut V> {
    self.entries.iter_mut().filter_map(Option::as_mut)
}
```

RobertJacobsonCDC:
Are all of the changes under `ixa-bench` related?

ANSWER: Yes, she added basic itinerary support in the baseline SIR model so we have a basis for comparison.

PR #861, opened 2026-04-28
chore: Replace `ValueVec` with `Vec` in `PropertyValueStoreCore` implementation

### LINKED ISSUE #860: Replace `ValueVec` with `Vec` in `PropertyValueStoreCore`
Previous refactoring work has removed the need to use `ValueVec` for the backing storage of properties. Previously, it was convenient to be able to write to the property storage in contexts in which we only have immutable access to the property storage. Those instances have been refactored away, and existing usages are all cases in which we actually have a mutable reference to the owning objects. Moving to a `Vec` simplifies the implementation.

We might want to retain `ValueVec` in the codebase as a utility until we have confidence we no longer want it.

RobertJacobsonCDC:
This PR replaces `ValueVec` with `Vec` in the implementation of `PropertyValueStoreCore` (the backing storage for property values). See #860.

Instead of removing `ValueVec` altogether, I relocated it to a `data_structures` submodule. That submodule can also be the home of other planned generic data structures like `IndexableMap` / `DenseEntityMap` / `EntityVec` whenever they are ready. Nothing is currently using `ValueVec`, though, so we might slate it for removal once we have confidence we don't want it anymore. (Or maybe we remove it in this PR?)


k88hudson-cfa (approved):
Looks good, let's keep `value_vec` to take a look at if we need it for Global Properties

PR #862, opened 2026-04-28
docs: Suggestions: network model

swo:
Builds off #847


swo:
Suggestions incorporated into #847 

cc @bbbruce

PR #863, opened 2026-04-28
feat: Introduce `DenseEntityMap`, `EntityVec`

### LINKED ISSUE #857: Use EntityId as index
Experiment with using `EntityId` as an index into data structures, either by implementing traits on `EntityId` or providing user access to data structures or some combination of the two.

This subsumes the `Indexable` trait idea.

RobertJacobsonCDC:
This PR adds two typed, vector-backed data structures for working with `EntityId<E>` keys:

- `EntityVec<E, V>`: a thin `Vec<V>` wrapper that can be indexed by `EntityId<E>`.
- `DenseEntityMap<E, V>`: a dense map from `EntityId<E>` to `V`, backed by `Vec<Option<V>>`.

The motivation here is that we want to maintain the invariant for `EntityId<E>` that only _valid_ entity ID values can ever exist, which requires preventing client code from constructing them, but we _also_ want efficient data structures in which we associate values to `EntityId<E>` efficiently. `HashMap<EntityId<E>, V>` is already available to client code, so we're really talking about using `EntityId<E>` to index into a vector-like structure.

# Issues and Open Questions

## In General

- There might be other opaque ID types like `PlanId` that we want to use with this. Right now our only use case seems to be `EntityId<E>`, but if there are other types in the future, we can change these data structures to be generic over a trait type (`AsIndex` / `Indexable`).

## `EntityVec<E, V>`

- Should an `EntityVec` *also* be indexable with a `usize`? Or only with `EntityId<E>`?

## `DenseEntityMap<E, V>`

- We could also have methods that expose the underlying `Vec<Option<V>>` so long as we don't allow unconstrained vector-growing operations on it (`push`, `extend`). The use case I am imagining is iterating over `(EntityId<E>, &mut Option<V>)`, where you are given the entity ID even when no `V` is associated with it yet.

### DIFF
--- a/src/data_structures/dense_entity_map.rs
+++ b/src/data_structures/dense_entity_map.rs
@@ -1,0 +1,610 @@
+/*!
+
+A `DenseEntityMap<E, V>` is a map from `EntityId<E>` to values of type `V` with a hash-map-like API
+optimized for densely populated maps.
+
+A `DenseEntityMap` is "dense" in the sense that it uses a vector `Vec<Option<V>>` internally for
+storage, using the `EntityId<E>` to index into the vector. Use a `DenseEntityMap` in any of the
+following cases:
+
+ - The number of entities is small
+ - Most entities are expected to be used as a key
+ - Access or creation speed is important
+ - You want access to the `EntityId<E>` key a value was stored with
+
+A `DenseEntityMap` allocates lazily by default, only allocating enough memory to access the item for
+the largest entity ID stored in the map. If you know beforehand how many entities you expect to
+store, you can use the `DenseEntityMap::with_capacity` constructor or `DenseEntityMap::reserve` to
+preallocate the map, which avoids reallocations.
+
+Because every value added to a `DenseEntityMap` is accompanied by a valid `EntityId<E>`,
+`DenseEntityMap` is guaranteed to only "store" valid entity IDs. You can therefore use it as a
+replacement for `EntityVec<E, V>` (or just `Vec<V>`) for cases where you need to recover the original
+entity ID that a value was stored with, for example, by iterating over the (entity ID, value) pairs
+returned by `DenseEntityMap::iter`. The only cost you pay for this is the extra memory needed to
+store `Option<V>` values instead of `V` values, which in some cases is nothing. The `EntityId<E>`
+itself is not stored.
+
+A `DenseEntityMap<E, V>` can be cheaply converted to an `EntityVec<E, Option<V>>` using the
+`DenseEntityMap::into_entity_vec` method.
+
+## Example
+
+Imagine you have `Person` and `Setting` entities, and you want an efficient way to store for each
+`SettingId` a `Vec<PersonId>` representing all the people that can be found in the setting. It is
+possible to use a `HashMap<SettingId, Vec<PersonId>>` to store this information, but a
+`DenseEntityMap<SettingId, Vec<PersonId>>` would be more efficient.
+
+```rust,ignore
+use ixa::data_structures::DenseEntityMap;
+
+let mut setting_membership = DenseEntityMap::<SettingId, Vec<PersonId>>::new();
+
+// During population initialization you might initialize the map with data in, say, a `PersonRecord`
+// struct that has a `home_id` field of type `SettingId`.
+let person_id = context.add_entity(with!(Person, person_record.age));
+let setting_members = setting_membership.get_or_insert(person_record.home_id, Vec::new);
+setting_members.push(person_id);
+
+// Look-ups are extremely efficient.
+if let Some(setting_members) = setting_membership.get(setting_id){
+    // Do something with the setting members.
+}
+
+// You can also iterate over the (entity ID, value) pairs.
+for (setting_id, setting_members) in setting_membership.iter() {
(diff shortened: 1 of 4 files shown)


src/data_structures/dense_entity_map.rs
```diff
@@ -0,0 +1,580 @@
...
+
+ - The number of entities is small
+ - Most entities are expected to be used as a key
+ - Access or creation speed is important
+ - You want access to the `EntityId<E>` key a value was stored with
+
+A `DenseEntityMap` allocates lazily by default, only allocating enough memory to access the item for
+the largest entity ID stored in the map. If you know beforehand how many entities you expect to
+store, you can use the `DenseEntityMap::with_capacity` constructor or `DenseEntityMap::reserve` to
+preallocate the map, which avoids reallocations.
```
k88hudson-cfa: I would make this a little more usage-oriented:

```suggestion
If you know beforehand how many entities you expect to store, you should use the  `DenseEntityMap::with_capacity` constructor or `DenseEntityMap::reserve` since it will be more efficient (it avoids reallocations every time you add an item)
```

RobertJacobsonCDC:
## Design Notes

`EntityVec` deliberately does not expose APIs that return `EntityId<E>` values. Although it can be indexed by an existing `EntityId<E>`, it does not know that every vector position corresponds to a valid entity. Its length can be changed with ordinary vector-like operations such as `push`, `extend`, `resize`, `truncate`, or `From<Vec<V>>`, and those operations are not tied to the entity store’s population. Returning `(EntityId<E>, &V)` pairs from `EntityVec` would require manufacturing IDs from raw indices, which could expose _invalid_ `EntityId` values that were never created by the entity system, so `EntityVec` is limited from doing so for this reason.

`DenseEntityMap` can safely return `EntityId<E>` values because every occupied entry is created by inserting with an existing `EntityId<E>`. Internally it may reconstruct the key from the occupied slot index during iteration, but that slot only became occupied because valid client code previously supplied that `EntityId<E>`. Empty slots are represented with `None` and skipped.

The two types are therefore similar in storage strategy but different in invariants:

- `EntityVec<E, V>` has vector semantics: every in-bounds index has a value, `len()` is the number of stored values, iteration yields values only, and growth appends anonymous positions.
- `DenseEntityMap<E, V>` has map semantics: keys may be sparse, `len()` is the number of present key-value pairs, missing entries are represented by empty slots, and iteration yields `(EntityId<E>, &V)` pairs.
- `EntityVec` stores `V` directly; `DenseEntityMap` stores `Option<V>`, which may add memory overhead depending on `V`.
- `EntityVec` is best when the entity ID is only needed for lookup. `DenseEntityMap` is best when the key must be recovered later.

## Use Cases

Use `EntityVec<E, V>` when values are naturally stored in entity-id order and the caller already has the `EntityId<E>` needed for lookup. It is appropriate for compact per-entity side data where the collection behaves like a vector and missing interior entries are not part of the model.

Use `DenseEntityMap<E, V>` when the collection behaves like a map but the key space is expected to be dense enough that vector indexing is preferable to hashing. It supports sparse keys, insertion, removal, presence checks, lazy growth to the largest inserted key, and iteration over stored `(EntityId<E>, value)` pairs.

Client code can already use `HashMap<EntityId<E>, V>` for fully general keyed storage. `DenseEntityMap<E, V>` is preferable when entity IDs are small or densely populated, when lookup and insertion speed matter, when deterministic entity-id iteration order is useful, or when avoiding hash-table overhead is worth the vector-backed representation. A `HashMap` remains the better choice for very sparse key spaces where allocating up to the largest entity ID would waste memory.

## Deferred Work

- The benchmarks from #858 are deferred to a follow-up PR.
- Setting membership implementation motivates this, but implementation of more specific settings functionality is out of scope for this PR.

src/data_structures/entity_map.rs
```diff
@@ -0,0 +1,610 @@
...
+            self.len += 1;
+        }
+
+        slot.as_mut().unwrap()
+    }
+
+    /// Removes and returns the value for `entity_id`, if present.
+    pub fn remove(&mut self, entity_id: EntityId<E>) -> Option<V> {
+        let removed = self.data.get_mut(entity_id.0).and_then(Option::take);
+        if removed.is_some() {
```
k88hudson-cfa, line 186: Do you want to add a test for this case?

k88hudson-cfa (approved):
I'm not convinced this should be `Dense` only if it's faster than `HashMap` for sparse data also. So I'd probably just call it an `EntityMap`. One thing that could potentially be problematic in some specific cases is always resizing on removal, but I think you can leave it as is.

RobertJacobsonCDC:
> One thing that could potentially be problematic in some specific cases is always resizing on removal, but I think you can leave it as is.

Yeah, I'm not convinced of that either. At best it saves you from scanning the empty tail when iterating. It isn't trimming the allocation. So maybe we shouldn't bother trimming the tail.

RobertJacobsonCDC:
> I'm not convinced this should be `Dense` only if it's faster than `HashMap` for sparse data also.

I don't especially like any of the names, so I'm happy to change it to `EntityMap`—by which I mean, I am equally unhappy using that name.

PR #866, opened 2026-04-30
docs: Properties chapter of the Ixa Book

### LINKED ISSUE #853: Ixa Book chapter on properties
Properties are sophisticated enough at this point that they probably justify their own chapter in the Ixa Book. We have also learned a few things working on `ixa-epi-covid` that might be worth including there.

- Use of `f64` data in a property and the `impl_eq_hash` parameter more generally. (#848)
  - When to use `impl_eq_hash = both` (generic generated imples) vs. `impl_eq_hash = neither` (manual impls) vs. `OrderedFloat<f64>`
- Separating data into different properties vs. storing multiple data in a single property
  - Indexing and/or querying needed? Separate fields.
  - Always accessed together? Same field gives locality, potential performance gain.
- custom display function for serialization to reports
- Types that don't fit the form expected by `define_property!` / `define_derived_property!`
- Types that are properties of multiple entities
- A warning about using type aliases with properties (#843)

Unclear if in scope, but might be nice to have _somewhere_:

- Patterns for properties that want to be vectors (or non-`Copy` types more generally)
  - M-N relationships using entities and properties
  - using an index vs. using a data plugin (custom indexing strategy)
- Using a property to store the "active" value of something (latest vaccine, active itinerary, etc.)
- Using a trait extension of `ContextBase` / a trait on `Context` to define an API that hides details of property implementation.

RobertJacobsonCDC:
This PR adds a chapter all about properties to the Ixa Book, including a section on implementing `Eq` and `Hash`. The source file is `docs/book/src/topics/properties.md`. (The GitHub rendered markdown file is [here](https://github.com/CDCgov/ixa/blob/RobertJacobsonCDC_853_properties_docs/docs/book/src/topics/properties.md).)

To render the book locally:

```bash
mise run docs
```

Then open `website/book/topics/properties.html`.

### DIFF
--- a/docs/book/src/SUMMARY.md
+++ b/docs/book/src/SUMMARY.md
@@ -11,6 +11,7 @@
   - [The Incidence Report](first_model/reporter.md)
   - [Next Steps](first_model/next-steps.md)
 - [Topics](topics/topics.md)
+  - [Properties](topics/properties.md)
   - [Indexing](topics/indexing.md)
   - [Burn-in Periods and Negative Time](topics/burn-in-and-negative-time.md)
   - [Handling Errors](topics/handling-errors.md)
--- a/docs/book/src/migration_guide.md
+++ b/docs/book/src/migration_guide.md
@@ -2,7 +2,10 @@
 
 ## Properties in the new Entity system
 
-In the new Entity system, properties work a little differently.
+In the new Entity system, properties work a little differently. We now have a
+[chapter devoted to properties in the ixa book](topics/properties.md). This section
+of the migration guide describes the changes in the new system relative to the
+old system.
 
 ### The Old Way to Define a Property
 
--- a/docs/book/src/topics/properties.md
+++ b/docs/book/src/topics/properties.md
@@ -1,10 +1,10 @@
 # Properties
 
 Properties are the data attached to an entity. For example, a `Person` entity might have the properties `Age` and
-`InfectionStatus`. In ixa, the property's value type is also the property type: the implementation of the trait
+`InfectionStatus`. In Ixa, the property's value type is also the property type: the implementation of the trait
 `Property\<Person>` by the concrete property type is what ties that Rust type to the `Person` entity.
 
-Never implement `Property\<T>` for a type `T` directly. Instead, use one of the provided macros:
+Never implement `Property\<E>` for an `Entity` type `E` directly. Instead, use one of the provided macros:
 
 | Macro                      | Use case                                      |
 |:---------------------------|:----------------------------------------------|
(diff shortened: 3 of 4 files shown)


swo (commented):
Overall I think this is a very helpful chapter. I learned a lot while reading it!

As I read, I fiddled with the text; see my [suggestions](https://github.com/CDCgov/ixa/compare/RobertJacobsonCDC_853_properties_docs...RobertJacobsonCDC_853_properties_docs-swo). It's mostly minor reorderings or rewordings, in ways that made things easier for me to understand.

A few things I wanted to call out:

- All properties take on values immediately upon creation of an entity. Is it idiomatic ixa to make properties that are Options for things you haven't implemented yet? Or is that straying into data plugin territory?
- You set up the explicit vs. constant vs. derived divide, but I think it should really just be derived vs. not.
  - I think it's very intuitive that a property can have a default value upon instantiation. This just feels like a nice convenience, not something deep about classes of properties.
  - If you do keep the three-way division, I'd recommend against calling them "constant," which is confusing, because you can set them to whatever.
  - While I'm here I'll also recommend replacing "non-derived" with "base" or "source," or something that clarifies what it actually is.
- I would re-organize a little bit, to start with non-derived properties and `define_property!` and how to use them, then go one step deeper and say that `define_property!` is just a nice wrapper around `impl_property!`, and then go into derived properties only after that. I can take a go at that if you think it would be helpful.
- Does Beau's thing, about wanting to use an enum for edge types, to avoid duplicating a value associated with that edge, match with the motivation for canonical values?
- I think it's useful to know that the same property type (eg `Age`) can be defined for multiple entities, and it's nice to know that that happens via implementing eg `Property<Person>`, but it's a little confusing because no one should be every writing `impl Property<Person> for Age` themsevles.

swo:
@RobertJacobsonCDC I recommend you merge this; I was pointing an analyst to it, and I think it's way better than nothing!

k88hudson-cfa (approved):
Looks good with follow-up comments from scott

RobertJacobsonCDC:
Thanks for all this feedback, @swo !



Some comments:



> Is it idiomatic ixa to make properties that are Options for things you haven't implemented yet?

Yes, and this is worth adding to this chapter. I'll add a section to make this explicit. (It's mentioned in the section on custom display functions.)



> You set up the explicit vs. constant vs. derived divide, but I think it should really just be derived vs. not.

Well, not exactly. I make this distinction *for initialization modes only*. But if you find it confusing, others will to. Let me tweak some language to make this more clear.



> I'd recommend against calling them "constant," which is confusing

This language comes from our need to distinguish between a "default" that is dynamically computed versus a default that is provided as a static constant in the property definition. So the adjective refers to the value being assigned to the property rather than the property itself.

We no longer support dynamically computing default values, so the distinction is vestigial. It probably makes sense to just call it "default value". But that's an API change out of scope of this PR. I'll open a separate ticket for it.



> I would re-organize a little bit...

I think I did a bit of reorganizing since your review that makes it closer to your suggestion. I'm doing another pass. Let me push my changes incorporating your feedback. 

But actually, if you're feeling motivated to work on Ixa documentation, we probably have a greater need for user guide material for topics that aren't covered at all. I made [a `docs` tag for issues related to documentation](https://github.com/CDCgov/ixa/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20label%3Adocs). It'd be really useful to know which topics you and your team run into that don't yet have any user-guide coverage but that most new users will want to learn about. We've sort of adopted a just-in-time documentation strategy of waiting until we identify a need before spending the effort, just as a pragmatic way of prioritizing work. 

Of course, any and all feedback and contributions are gold. 



> Does Beau's thing, about wanting to use an enum for edge types, to avoid duplicating a value associated with that edge, match with the motivation for canonical values?

I don't know. I'm not familiar with Beau's design for edge types.



> but it's a little confusing because no one should be every writing `impl Property<Person> for Age` themsevles.

Right. I try to put this front and center when I introduce the list of macros. I also use the example of using the same type as a property for multiple entity types under the "When to use `impl_property!`" section. 

For some of these things, I wonder if it would be helpful to have a whole chapter devoted to answering questions of the form, "How do I do X?" Or maybe a chapter of common "modeling patterns"? (Or both?)

PR #873, opened 2026-05-04
chore: Refactor `global_properties` to use the registry pattern

### LINKED ISSUE #865: Refactor global properties to use the "registry pattern"
Use the same pattern for global properties as we use for data plugins.

RobertJacobsonCDC:
This PR refactors global property storage out of the data plugin system and into `Context` directly. The original hash map based storage is replaced with the "registry pattern" we use for data plugins, entities, and other things elsewhere: `set_global_property_value` and `get_global_property_value` now index directly into `Context::global_properties` using `T::id()`. 

I also removed the unused string-based getter mechanism. 

Unfortunately, the name-based registration and loading machinery is otherwise left intact, so `load_global_properties`, `add_global_property`, and the serialized lookup path continue to work through the existing accessors. The reason is that the `load_global_properties`  uses a string name-based lookup, and it's not clear how to refactor this while also retaining source compatibility with existing code.

On the other hand, if we are willing to break source compatibility, we might as well wait until we have a better idea of the alternative design for global parameters in order to do so, so we don't give client code whiplash.

Client code should remain source compatible for this PR.

### DIFF
--- a/src/global_properties.rs
+++ b/src/global_properties.rs
@@ -3,7 +3,7 @@
 //! Global properties are not mutable and represent variables that are
 //! required in a global scope during the simulation, such as
 //! simulation parameters.
-//! A global property can be of any type, and is is just a value
+//! A global property can be of any type and is just a value
 //! stored in the context. Global properties are defined by the
 //! [`crate::define_global_property!()`] macro and can then be
 //! set in one of two ways:
@@ -15,21 +15,21 @@
 //! will result in an error.
 //!
 //! Global properties can be read with [`Context::get_global_property_value()`]
-use std::any::{Any, TypeId};
+use std::any::Any;
 use std::cell::RefCell;
-use std::collections::hash_map::Entry;
 use std::error::Error;
 use std::fmt::Debug;
 use std::fs;
 use std::io::BufReader;
 use std::path::Path;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, LazyLock, Mutex};
 
 use serde::de::DeserializeOwned;
 
 use crate::context::Context;
 use crate::error::IxaError;
-use crate::{define_data_plugin, trace, ContextBase, HashMap, HashMapExt};
+use crate::{trace, ContextBase, HashMap, HashMapExt};
 
 type PropertySetterFn =
     dyn Fn(&mut Context, &str, serde_json::Value) -> Result<(), IxaError> + Send + Sync;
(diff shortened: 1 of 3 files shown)


src/global_properties.rs
```diff
@@ -175,37 +150,14 @@ pub trait ContextGlobalPropertiesExt: ContextBase {
...
-            name: T::name().to_string(),
-            source,
-        })?;
-        let data_container = self.get_data_mut(GlobalPropertiesPlugin);
-        data_container.set_global_property_value(&property, value)
-    }
+    ) -> Result<(), IxaError>;
 
     /// Return value of global property T
     #[allow(unused_variables)]
```
k88hudson-cfa: I think you can remove this lint exception

src/macros/define_global_property.rs
k88hudson-cfa, line 41: I think this can be removed now right?

k88hudson-cfa (approved):
This looks good with a couple of small cleanup things. For the record I don't believe  the which/that change is justified, but I wouldn't block on it

src/macros/define_global_property.rs
RobertJacobsonCDC: It's used in error messages—but they are errors that only occur if there is a bug in ixa internals. It can also be useful for debugging during development. We also have this on properties and entities.

What I am thinking is, I'd like to refactor global properties to be more in line with how entity properties work. To satisfy the principle of least surprise, since `Property` has a `name` method, `GlobalProperty` also has a name method. Of course, at least for `Property` (but not for `Entity`), the name itself is used for the canonical value mechanism for multi-properties, so there's actually a separate motivation for its existence in that type... 

But I'm not married to this idea at all. Maybe none of these types really need a name method at the end of the day. And for that matter, the error checking that we do that currently uses `name()` might also be unnecessary—maybe we just call `unwrap()` or used some `*_unchecked` method instead—because if the implementation of ixa is correct then those errors never occur.

Anyway, this is probably a pretty low stakes decision, and I don't have strong feelings either way, but I'd say let's just leave it in for now. It's not exactly hurting anything being there. Thoughts?

RobertJacobsonCDC:
> For the record I don't believe the which/that change is justified, but I wouldn't block on it

Alright, alright! Reverted!

src/macros/define_global_property.rs
k88hudson-cfa: Ah thanks helpful. Yes let's just leave it

PR #875, opened 2026-05-05
feat: Add benchmark chart toggles for y-axis starting at 0.0, outlier clipping

RobertJacobsonCDC:
Added two quality of life toggles to the benchmark history page:

1. **start the y-axis at 0.0**, making absolute benchmark scale easier to compare across runs. Turning it off preserves Chart.js’s existing automatic axis behavior.
2. **MAD clipping** heuristic for extreme high outliers in time-series charts.

Both are toggled on by default.

### DIFF
--- a/scripts/index.html
+++ b/scripts/index.html
@@ -220,6 +220,11 @@
           y-axis starts at 0.0
         </label>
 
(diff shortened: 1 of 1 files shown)


cdc-as81 (approved):
LGTM

PR #877, opened 2026-05-05
feat: add `sample_single_excluding` to `random` module

k88hudson-cfa:
This adds a `sample_single_excluding` method for sampling a random peer member (determined via `PartialEq`). I guess we could also rename it to `sample_peer`.

```rs
let peers: Vec<PersonId> = /* ... */;
let me: PersonId = /* ... */;

let peer = context
      .sample(PeerRng, |rng| sample_single_excluding(rng, &peers, me))
      .expect("at least one other peer");
 let peer = context.sample(PeerRng, |rng| sample_single_excluding(rng, &peers, &me));
```
It uses iteration at small n and rejection at larger n (>=4). Benchmarks suggest the advantages for iteration are clear at `2` and within noise for `3`. To run benchmarks:

```
cargo bench --bench sample_single_excluding -p ixa-bench
```

Results (3 runs, `sample_size = 200`, `--warm-up-time 2 --measurement-time 5`):
                                                                                                            
  | n | run 1 rej / iter / auto | run 2 rej / iter / auto | run 3 rej / iter / auto | median best | gap |   
  |---:|---|---|---|---|---|                                                                                
  | 2 | 26.0 / 6.2 / 7.1 | 20.0 / 5.8 / 6.9 | 17.9 / 6.8 / 6.8 | iteration ~6.3 | iteration ~3x faster than rejection |                                                                                               
  | 3 | 13.1 / 13.7 / 19.0 | 13.2 / 11.7 / 12.3 | 13.2 / 12.8 / 14.0 | tied ~13 | within noise |          
  | 4 | 10.4 / 15.1 / 11.0 | 11.2 / 21.1 / 9.0 | 12.1 / 15.3 / 11.0 | rejection ~11 | rejection ~30% faster than iteration |

### DIFF
--- a/ixa-bench/Cargo.toml
+++ b/ixa-bench/Cargo.toml
@@ -108,3 +108,8 @@ harness = false
 name = "property_semantics"
 path = "criterion/property_semantics.rs"
 harness = false
+
+[[bench]]
+name = "sample_single_excluding"
+path = "criterion/sample_single_excluding.rs"
+harness = false
--- a/ixa-bench/criterion/sample_single_excluding.rs
+++ b/ixa-bench/criterion/sample_single_excluding.rs
@@ -16,7 +16,7 @@ use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use ixa::rand::rngs::StdRng;
 use ixa::rand::SeedableRng;
 use ixa::random::{
-    sample_excluding_iteration, sample_excluding_rejection, sample_single_excluding,
+    sample_single_excluding, sample_single_excluding_iteration, sample_single_excluding_rejection,
 };
 
 const SEED: u64 = 42;
@@ -65,9 +65,12 @@ fn bench_strategies(c: &mut Criterion, results: &Results) {
         group.bench_with_input(BenchmarkId::new("rejection", &id_param), &(), |b, _| {
             let mut rng = StdRng::seed_from_u64(SEED);
             let ns = bench_ns_per_sample(b, || {
-                let v =
-                    sample_excluding_rejection(black_box(&mut rng), black_box(&slice), excluded)
-                        .unwrap();
+                let v = sample_single_excluding_rejection(
+                    black_box(&mut rng),
+                    black_box(&slice),
+                    excluded,
+                )
+                .unwrap();
                 black_box(v);
             });
             record(results, "rejection", n, ns);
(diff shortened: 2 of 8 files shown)


RobertJacobsonCDC:
I would make the naming uniform. We have:

- `sample_single_excluding`
- `sample_excluding_iteration` → I'd call it `sample_single_excluding_iteration`
- `sample_excluding_rejection` → I'd call it `sample_single_excluding_rejection`



# Questions

## Assumptions?

Can you say more about the assumptions of the data being sampled from? My interpretation of your implementations is that our assumptions are these:

- We are *not* assuming that elements are *distinct*. In particular, there may be several copies of the excluded value in the data.
- We are not assuming any ordering.
- We are not assuming that the data actually even contains a non-excluded element.



I ask because we can do a lot better if we know that the excluded value, if it exists in the set, is unique. In fact, correct me if I am wrong here, but the semantics we almost always want is, *sample uniformly from the set of values*, not, *sample uniformly from the set of non-excluded indices*. So should we instead implement a function `sample_single_excluded_from_distinct`?

## Container types?

Also, you have implemented this only for slices. Don't we need it for `EntitySet`s? In which case I think we actually have three container categories:

1. Known length random access containers (slices, `Vec<T>`). We can sample an index from `0..container.len()` and then index into it in constant time, `container[idx]` (equivalently, the iterator's `nth` implementation is an efficient lookup).
2. Known length linear access containers (`HashMap`, `BinaryHead`). We can still sample an index from `0..container.len()`, but we cannot index into the container (equivalently, the iterator's `nth` implementation is a linear scan).
3. Unknown length linear access containers (streams, sets computed on demand). We cannot sample an index, and accessing the `nth` element is a linear scan.

Rejection sampling has different asymptotics between category 1 and 2, and I think you need to use reservoir sampling for category 3 no matter what. 

I have some ideas about your implementation that I'll put in a separate comment. I have a write-up, but I want to benchmark a real implementation of my suggestions first.

RobertJacobsonCDC:
Reviewing this was really fascinating.

My first reaction was to try an alternative algorithm with better runtime complexity. Your `sample_excluding_iteration` algorithm takes $3*n/2$ steps on average. My alternative takes $(n+1)/2 + m$ steps on average, where $m$ is the count of excluded values. So my algorithm should be faster in every case, approaching yours as $m \to n$. Except in practice it's not actually faster, at least not for the range of inputs we care about, namely when $m$ is some significant percentage of $n$ and $n$ less than 1000 or so.

The $n = 2$ case is also bananas. The following is _slower_ than `sample_excluding_iteration`:

```rust
let idx = rng.random_range(0..2u32);
let excluded = excluded.borrow();

if slice[idx as usize] != *excluded {
    return Some(&slice[idx as usize]);
}
if slice[(1 - idx) as usize] != *excluded {
    return Some(&slice[(1 - idx) as usize]);
}
None
```

What?!

Anyway, it looks like your implementation is as good as it gets for the inputs we care about. (Under the given assumptions, of course.)

k88hudson-cfa:
I refactored this to move all the sampling methods to EntitySet because `sample_entity_excluding` needs two independent random-access lookups at arbitrary indices, which is awkward in the iterator version where `nth(p)` has already consumed everything before `p`. Once that method lives on `EntitySet`, I thought it made sense to move the other methods too. 

I think the existing methods could use some additional test coverage but I figured it would be better to do in a follow-up since this patch is already getting large

RobertJacobsonCDC (approved):

PR #879, opened 2026-05-05
chore: Remove progress bar feature

RobertJacobsonCDC:
Removed the progress bar feature. This deletes the progress API/module, removes the timeline progress update from `Context`, drops the `--timeline-progress-max` runner option, and simplifies logging by removing the progress-bar-specific encoder wrapper.

The feature/dependency wiring and runner integration test were removed, and the CLI docs were regenerated so the removed flag no longer appears. The performance docs no longer describe progress bar APIs.


k88hudson-cfa (approved):
LGTM

PR #880, opened 2026-05-05
chore: Update several dependencies

### LINKED ISSUE #876: Bump non-`rand`-related dependencies
Several dependencies are out of date and need bumping:

- `ctor@0.5.1` -> `ctor@1.0.1`: This one might require usage changes. Investigate whether the `wasm-bindgen` workaround in `src/lib.rs` is still required after updating.
- `criterion@0.7.0` -> `criterion@0.8.2`
- `strum@0.27.1` -> `strum@0.28.0`
- `sysinfo@0.37.0` -> `sysinfo@0.38.4`

Remove no longer used:

- `hashbrown`

In all cases, check compatibility with `ixa-epi-covid`—in particular, can we bump the corresponding `ixa-epi-covid` dependencies?

**Out of scope:** Anything `rand`-adjacent. These need to be bumped all at once, and only when the ecosystem has sufficiently adopted the new `rand` version.

~This issue might be convenient to combine with #834 .~

RobertJacobsonCDC:
This PR updates the following dependencies:

- `ctor@0.5.1` -> `ctor@1.0.1`
- `criterion@0.7.0` -> `criterion@0.8.2`
- `strum@0.27.1` -> `strum@0.28.0`
- `sysinfo@0.37.0` -> `sysinfo@0.38.4`

Also removed the unused `hashbrown` crate.

The only one requiring any source changes is `ctor`. The new version allows us to get rid of the `#[cfg(target_family = "wasm")]` stuff we have littering `lib.rs`. I separated it into its own commit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ serde_json = "^1.0.135"
 ixa-derive = { version = "2.0.0-beta.2", path = "ixa-derive" }
 seq-macro = "^0.3.5"
 paste = "^1.0.15"
-ctor = "^0.5.0"
+ctor = "^1.0.1"
 clap = { version = "^4.5.26", features = ["derive"] }
 clap-markdown = "0.1.5"
 log = "^0.4.22"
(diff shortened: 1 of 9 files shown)


k88hudson-cfa:
It looks to me like `link-section`, which is used by `ctor`'s `priority` feature, relies on nightly-only features in the wasm path, which seems less than ideal/maybe worth reporting upstream? We could work around this with `default-features = false, features = ["std", "proc_macro"]` for now since we don't use that feature

k88hudson-cfa (approved):
Nice!

PR #883, opened 2026-05-08
feat: Compare `sample_entity_whole_population` relative to "sentinel" CPU-bound benchmark

### LINKED ISSUE #874: Investigate odd variance of whole population sampling
The plots for whole population sampling on `ixa/scripts/index.html` show that odd behavior was introduced at or before PR #844. With the exception of the benchmark for #835, the behavior didn't exist prior to that PR.  Sometimes the operation is 3x as slow.

RobertJacobsonCDC:
Motivated by #874, this PR adds a CPU-bound "sentinel" benchmark and compares the _relative_ performance of `sample_entity_whole_population` in `sample_entity_scaling`. The idea is, even on different CPUs this relative performance will be stable.

Updated benchmark result processing to record `relative_to_sentinel` metadata for `sample_entity_whole_population/{1000,10000,100000}` when the sentinel is present. The dashboard now overlays that relative series on the same chart as absolute timing, with a right-side `x sentinel` axis, while hiding the sentinel as a standalone benchmark. PR benchmark comments also filter out the internal sentinel row.

Added test coverage for the result parser and PR comment formatter, including sentinel ratio recording and sentinel-row filtering.

### DIFF
--- a/ixa-bench/criterion/sample_entity_scaling.rs
+++ b/ixa-bench/criterion/sample_entity_scaling.rs
@@ -5,6 +5,7 @@ use std::time::Instant;
 
 use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion};
 use ixa::prelude::*;
+use ixa::rand::Rng;
 
 define_rng!(SampleScalingRng);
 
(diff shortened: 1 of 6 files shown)


RobertJacobsonCDC:
~Because this PR technically adds a benchmark, the benchmark is expected to fail in CI.~ 

Edit: We fixed this behavior in a recently merged PR.

k88hudson-cfa (approved):

PR #884, opened 2026-05-09
feat: Added "internals" MDBook

### LINKED ISSUE #570: Implementation documentation
It would be really nice to have documentation about internal implementation beyond what is in the code itself. A few diagrams have been checked in to the repo some time ago, but they are (or soon will be) outdated. Recently we added some material to [the wiki](https://github.com/CDCgov/ixa/wiki). For now, the wiki looks like the best home for this material.

Related to ongoing work with property storage, indexing, and querying:

- #541 
- #564 
- #533 
- #536

RobertJacobsonCDC:
The PR adds an in-repo home for developer-oriented notes and documentation about `ixa` internals.

- I moved some content from the GitHub repo wiki and deleted obsolete wiki content.
- Organized as an MDBook
- Added a `docs:internals` mise task but left existing `mise run docs` behavior the same.


RobertJacobsonCDC:
This is also a natural home for diagrams that aren't (yet) used in the Ixa Book itself.

k88hudson-cfa (commented):
Generally speaking, I think this kind of design documentation is difficult to keep up to date as code evolves but is useful to understand the original intent of the authors if somebody wants to go back in history and figure out why things were implemented a certain way. I'd recommend treating some of this as a series of snapshots (like an Architectural Decision Record or similar), which we could make a section for. 

If you intend to keep this up to date, it's not particularly well-suited to being maintained in a collaborative way: breaking it up into sections and having it live closer to the code would be an improvement IMO.

RobertJacobsonCDC:
Based on discussion, I'm going to rework this to have a more "Architectural Decision Record" (ADR) structure. Example: https://github.com/mozilla/fxa/blob/main/docs/adr/0028-evaluate-playwright.md. The idea is, architectural descriptions are presented in a more "snapshot" format than as a living document that needs to be kept in sync. This also gives me a natural place for some of the more detailed notes I have historically been shoving in a PR description.

Some editorial thoughts.
- The idea isn't to be completely comprehensive, only to be _useful_. 
- These notes shouldn't be an additional development burden. Rather, they should be a way to capture intellectual effort that has already been spent.

RobertJacobsonCDC:
Closing this PR for now. I'll take another stab at this another time. The parent issue #570 will remain open until then.

PR #890, opened 2026-05-12
fix: execution stats availability messaging

### LINKED ISSUE #851: "Memory and CPU statistics are not available on your platform." on WASM
We have had some semantic drift in our profiling / execution stats system. Consequently, running our examples results in the message

> Memory and CPU statistics are not available on your platform.

being printed on the console.

This is a meta-issue in the sense that there are probably multiple independent things that need fixing:

- Should these examples try to print execution stats? (I would argue yes, probably.) If so, can we enable stats for them specifically? Or maybe they should respect a command line flag if they use the runner.
- If memory and CPU stats are not available for the platform (which is the case for WASM builds specifically), execution stats consists of wall-time. So there is still something to display—we don't need to turn off printing of the stats. Is the message being displayed via a log message? It should be. Is it a warning? If so, maybe it should be demoted to `INFO` or something?
- That particular message should only be displayed if it really is the case that 
  1. the execution stats features are enabled but 
  2. those features are unavailable on the user's platform. 
  
  As far as I know, they are only unavailable for WASM. So it appears we are printing that message for condition (1) instead of condition (1) AND (2).

cdc-as81:
## Summary

- Distinguishes disabled execution resource stats from platform-unavailable resource stats.
- Stops printing `Memory and CPU statistics are not available on your platform.` when `profiling` is not enabled.
- Removes the runner’s unconditional WASM stats warning and keeps `--no-stats` behavior unchanged.


src/runner.rs
```diff
@@ -300,14 +300,7 @@ where
 
     context.init_random(args.random_seed);
 
-    if args.no_stats {
-        context.print_execution_statistics = false;
-    } else {
-        if cfg!(target_family = "wasm") {
-            warn!("the print-stats option is enabled; some statistics are not supported for the wasm target family");
```
RobertJacobsonCDC, line 307: I kind of like the idea of keeping this warning specifically for Wasm builds. We're saying, "You explicitly requested we collect runtime statistics, but for your platform most of those statistics are unavailable." That seems to me to be a reasonable thing to inform the user. My only question is, should the log message be category `warning` or `info`?

src/execution_stats.rs
```diff
@@ -41,6 +41,33 @@ pub struct ExecutionStatistics {
...
+fn resource_statistics_status(stats: &ExecutionStatistics) -> ResourceStatisticsStatus {
+    #[cfg(not(feature = "profiling"))]
+    {
+        let _ = stats;
+        ResourceStatisticsStatus::Disabled
+    }
+
+    #[cfg(feature = "profiling")]
+    {
+        if stats.max_memory_usage == 0 {
```
RobertJacobsonCDC: No, don't give special semantics to a zero value. Do an actual platform check.

RobertJacobsonCDC (changes requested):
This is really overwrought. I don't see any real advantage to introducing the `ResourceStatisticsStatus` enum and `resource_statistics_status`. You only have two call sites:

1. `print_execution_statistics`
2. `log_execution_statistics`

Since you're multiplying  `#[cfg(...)]` attributes at every call site _and_ in the definitions of `ResourceStatisticsStatus` /  `resource_statistics_status`, you've actually make the implementation more complicated by spreading implementation support checks across three different places (on top of introducing a new enum and free function).

Instead, inline the platform support detection at each call site. The logic is something like:

```rust
if cfg!(feature = "profiling") {
  if cfg!(target_family = "wasm") {
    // Or maybe `info!`
    warn!("Memory and CPU statistics are not available on your platform.");
  }
  // Do whatever the function does when profiling is enabled.
} else {
  // Do whatever the function does when profiling is disabled.
}
// Do whatever the function does unconditionally.
```

src/execution_stats.rs
```diff
@@ -238,11 +238,11 @@ impl ExecutionProfilingCollector {
...
     println!("━━━━ Execution Summary ━━━━");
-    if summary.max_memory_usage == 0 {
-        println!("Memory and CPU statistics are not available on your platform.");
-    } else {
-        #[cfg(feature = "profiling")]
-        {
+    #[cfg(feature = "profiling")]
+    {
+        if cfg!(target_family = "wasm") {
+            println!("Memory and CPU statistics are not available on your platform.");
```
RobertJacobsonCDC, line 244: What does `println!` do on a Wasm target? Use `info!` instead so it is symmetric to `log_execution_statistics`.

RobertJacobsonCDC (changes requested):

src/execution_stats.rs
cdc-as81: since in WASM println! does not show anywhere, users should call log_execution_statistics and not print_execution_statistics

RobertJacobsonCDC: There's a logic to that. 👍

RobertJacobsonCDC (approved):

PR #891, opened 2026-05-12
chore: remove double validate on add global prop

### LINKED ISSUE #819: Validation of inputs is performed twice in `define_global_property!`
While debugging the `validate_inputs` function in ixa-epi-covid, I found that the function runs twice when the model is run, even though `define_global_property!(GlobalParams, Params, validate_inputs);` is the only place in ixa-epi-covid where the function is called.

From some limited investigation, it appears that:

The `setter` in `add_global_property` itself calls `validate` at https://github.com/CDCgov/ixa/blob/45a10907391173ba67c724367cc3fa9b8f81851e/src/global_properties.rs#L70
 
and then calls it again via `set_global_property_value` at https://github.com/CDCgov/ixa/blob/45a10907391173ba67c724367cc3fa9b8f81851e/src/global_properties.rs#L179

cdc-as81:


RobertJacobsonCDC (approved):
LGTM

PR #894, opened 2026-05-12
chore: handle new Criterion benchmarks in CI

### LINKED ISSUE #892: CI should not fail when benchmarks are added
Right now the failure happens because criterion can't compare the new benchmark against anything (it doesn't exist)

cdc-as81:
## Summary

- Use Criterion's `--baseline-lenient` mode so newly added benchmarks do not fail PR benchmark CI.
- Report benchmarks without a saved baseline in a new `Not Compared` section.
- Keep real benchmark, parser, and regression failures failing CI.
- pinned node and pnpm in mise.toml, since we do not want every update to break our tests.

### DIFF
--- a/.github/workflows/bench-compare.yaml
+++ b/.github/workflows/bench-compare.yaml
@@ -49,25 +49,35 @@ jobs:
           target="${{ matrix.criterion_target }}"
 
           base_status=0
+          base_missing_target=0
           compare_status=0
           regress_status=0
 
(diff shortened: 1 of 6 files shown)


RobertJacobsonCDC (approved):
Can you say why we are pinning the versions in this PR? I don't recall my we originally used "latest" here, so I don't have any context for this.

Otherwise, this seems to look ok to me.

PR #895, opened 2026-05-13
chore: reference sir with entity map

### LINKED ISSUE #893: Implement usage of EntityMap/sampling for settings in reference sir
This will serve as an example for how to use the code

erik-rosenstrom:
This PR updates the reference Ixa SIR to have an example that uses `EntityMap`s to implement settings. It is compared against the baseline SIR which uses vectors index by household. 

This implementation mirrors the implementation of #858


ixa-bench/src/reference_sir/sir_ixa.rs
```diff
@@ -32,12 +38,27 @@ define_property!(
     default_const = InfectionStatus::Susceptible
 );
 
-trait InfectionLoop {
+define_property!(
```
k88hudson-cfa: It doesn't look like this is used for anything?

ixa-bench/src/reference_sir/sir_ixa.rs
```diff
@@ -147,13 +195,43 @@ impl InfectionLoop for Context {
...
+                assignment.push(h);
+            }
+        }
+        while assignment.len() < population {
+            let h = self.sample_range(HouseholdRng, 0..num_households);
+            assignment.push(h);
+        }
+
+        // Add people in assignment order and record their household.
+        for &h_idx in &assignment {
```
k88hudson-cfa, line 221: You could pre-allocate before this with `reserve`:

```rs
households_data.members.reserve(num_households);
households_data.by_person.reserve(population);
```

k88hudson-cfa (changes requested):
Looks good!

ixa-bench/src/reference_sir/sir_ixa.rs
erik-rosenstrom: I was having some trouble with the `add_entity` method, this was left over code from trying to resolve that issue. I will remove it.

k88hudson-cfa (approved):
Thanks! Looks good

PR #902, opened 2026-05-18
chore:  add timeout for playwright browser install

### LINKED ISSUE #897: Unpin node/pnpm, fix wasm tests
In https://github.com/CDCgov/ixa/pull/894 we had an issue with wasm tests timing out; we want to unpin and fix those

Also, apparently our timeout is super long?

cdc-as81:


mise.toml
```diff
@@ -1,6 +1,6 @@
 [tools]
-node = "22.20.0"
-pnpm = "10.33.0"
+node = "latest"
+pnpm = "latest"
```
RobertJacobsonCDC, line 3: This needs an explanation, since we only just pinned them.

RobertJacobsonCDC (changes requested):
Why aren't we limiting the runtime of every GitHub action? Why just this one in particular? It seems like there should be a more general solution that applies to everything.

mise.toml
cdc-as81: the timeout we are adding here is for the install (actually the download...)
the issue we have seen earlier when downloads were hanging was due to pnpm source moving to a different site where it now requires crt for the download.

RobertJacobsonCDC (approved):
The fact that we now set the CA certs means that we can access the latest `node` and `pnpm` sources.

PR #905, opened 2026-05-18
chore: Make rand usage deterministic and disable default features

### LINKED ISSUE #900: Remove getrandom dependency
The `getrandom` dependency is annoying to deal with downstream for wasm projects; it is only used by os_rng + thread_rng, which we don't actually need. We should consider changing our dependency on `rand` to `default-features = false`

cdc-as81:
## Summary

- Disable `rand` default features and explicitly enable only `std`, `small_rng`, and `std_rng`
- Replace remaining `thread_rng` usage with seeded RNGs in tests and benchmarks
- Use a deterministic default seed for benchmark population generation
- Remove the direct wasm `getrandom` dependency entry from the main crate manifest

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,11 @@ authors = ["The Ixa Developers <cfa@cdc.gov>"]
 
 [workspace.dependencies]
 approx = "^0.5.1"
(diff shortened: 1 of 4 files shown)


k88hudson-cfa (approved):
Looks good, thanks!

PR #915, opened 2026-05-20
feat: support primitive types in define_property!

k88hudson-cfa:
This is a proposal for a possible alternative that supports primitive values. The overall idea is that `get_property` returns the inner primitive, but you can still use the wrapper type for `set_property`/`with!`

```rs
let a: u8 = ctx.get_property::<Person, MyAge>(p);
ctx.set_property(p, MyAge(25));
ctx.add_entity(with!(Person, MyAge(25)));
```

### Before (legacy form, still works)
```rust
define_property!(struct Age(u8), Person);
define_property!(struct Score(i32), Person, default_const = Score(0));
define_derived_property!(struct IsAdult(bool), Person, [Age], \|age\| IsAdult(age.0 >= 18));`
```

### After

```rust
define_property!(Age: u8, Person);
define_property!(Score: i32, Person, default_const = 0);`
define_derived_property!(IsAdult: bool, Person, [Age], \|age\| IsAdult(age >= 18));
```
Note: the macro generates `pub struct Age(pub u8)` plus `From<u8>`/`From<Age>` impls

## `Property` trait — new associated type + methods

**Before:**

```rust
pub trait Property<E: Entity>: AnyProperty {
    type CanonicalValue;
    // ...
}
```

**After:**

```rust
pub trait Property<E: Entity>: AnyProperty {
    type Value: AnyProperty;
    fn into_value(self) -> Self::Value;
    fn from_value(value: Self::Value) -> Self;
    type CanonicalValue;
    // ...
}
```

### DIFF
--- a/docs/book/models/disease_model/src/transmission_manager.rs
+++ b/docs/book/models/disease_model/src/transmission_manager.rs
@@ -13,7 +13,7 @@ define_rng!(TransmissionRng);
 fn attempt_infection(context: &mut Context) {
     trace!("Attempting infection");
     let person_to_infect = context.sample_entity(TransmissionRng, Person).unwrap();
-    let person_status: InfectionStatus = context.get_property(person_to_infect);
+    let person_status: InfectionStatus = context.get_property::<Person, InfectionStatus>(person_to_infect);
 
     if person_status == InfectionStatus::S {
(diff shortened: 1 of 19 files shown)


k88hudson-cfa:
Here's an overview of all the options I can think of:

image

PR #917, opened 2026-05-22
feat: `track_periodic_value_change_counts!` macro

RobertJacobsonCDC:
Introduces a macro to simplify the syntax for calling `context.track_periodic_value_change_counts(...)`.

Some example invocations:

```rust
track_periodic_value_change_counts!(
    context,
    Person,
    InfectionStatus,
    1.0,
    handle_incidence_tracking
);

track_periodic_value_change_counts!(
    context,
    Person,
    InfectionStatus,
    [Age],
    1.0,
    move |_context, counter| {
        let _ = counter;
    }
);

track_periodic_value_change_counts!(
    context,
    Person,
    InfectionStatus,
    [Age, HighRisk],
    1.0,
    handle_incidence_tracking
);

```

PR #921, opened 2026-05-25
feat: `schedule_relative!` macro

### LINKED ISSUE #914: `schedule!`?
I've already written this pattern many times:

```rust
context.add_plan(context.get_current_time() + my_delay, move |context| {
    my_action(context, arg1, arg2, arg3)
});
```

and I'm wishing I could say:

```rust
schedule!(context, my_delay, my_action, arg1, arg2, arg3);
```

This would obviate the `schedule_my_action()` functions that seem common and redundant.

RobertJacobsonCDC:
Implemented a convenience macro:

```
schedule_relative!(context, my_delay, my_action, arg1, arg2, arg3);
```

expands to:

```rust
context.add_plan(context.get_current_time() + my_delay, move |context| {
    my_action(context, arg1, arg2, arg3)
});
```

The point of the macro is:

1. The convenience of scheduling at a relative future time instead of an absolute future time.
1. Capturing the concrete parameter values passed to the handler without having to manually write the closure syntax.

# Open Questions / Issues

- A similar macro `track_periodic_value_change_counts` (#917) mirrors the name of the method, so maybe we want to name this `add_plan_relative!` instead?
- The "relative" bit is just because the macro adds the `my_delay` parameter to `context.get_current_time()` for you. Maybe we should have "relative" versions of the methods for convenience as suggested by #836? Likewise, should there be an _absolute time_ version of this `schedule_relative!` macro, maybe called `schedule!` or just `add_plan!`?

### DIFF
--- a/src/macros/schedule_relative.rs
+++ b/src/macros/schedule_relative.rs
@@ -1,0 +1,32 @@
+/// Schedules an action after a delay relative to the context's current time.
+///
+/// The scheduled action is called with the executing `context: &mut Context` as
+/// its first argument, followed by each remaining macro argument in the same
+/// order.
+///
+/// For example:
+///
+/// ```ignore
+/// schedule_relative!(context, my_delay, my_handler, arg1, arg2, arg3);
+/// ```
+///
+/// expands to code like:
+///
+/// ```ignore
+/// {
+///     let time = context.get_current_time() + my_delay;
+///     context.add_plan(time, move |context| {
+///         my_handler(context, arg1, arg2, arg3)
+///     })
+/// }
+/// ```
+#[macro_export]
+macro_rules! schedule_relative {
+    ($context:expr, $delay:expr, $action:expr $(, $arg:expr)* $(,)?) => {{
+        let context = ($context);
+        let time = context.get_current_time() + $delay;
+        context.add_plan(time, move |context| {
+            ($action)(context $(, $arg)*)
+        })
+    }};
+}
--- a/integration-tests/ixa-runner-tests/tests/macros.rs
+++ b/integration-tests/ixa-runner-tests/tests/macros.rs
@@ -161,6 +161,26 @@ mod tests {
 
     fn macro_named_value_change_handler<C>(_context: &mut Context, _counter: &mut C) {}
 
+    fn record_sum(
+        context: &mut Context,
+        records: Rc<RefCell<Vec<(f64, u32)>>>,
+        arg1: u32,
+        arg2: u32,
+        arg3: u32,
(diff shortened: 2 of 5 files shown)


src/macros/schedule_relative.rs
```diff
@@ -0,0 +1,34 @@
...
+///     })
+/// }
+/// ```
+///
+/// The `context` may be an expression and is evaluated twice, once to read the current time and
+/// once to add the plan.
+#[macro_export]
+macro_rules! schedule_relative {
+    ($context:expr, $delay:expr, $action:expr $(, $arg:expr)* $(,)?) => {{
+        let time = ($context).get_current_time() + $delay;
```
k88hudson-cfa: why not just assign `let ctx = $context` here to avoid the doube eval?

k88hudson-cfa (approved):
One suggestion but otherwise this looks good

src/macros/schedule_relative.rs
RobertJacobsonCDC: I should also introduce a local scope for the temporaries.

RobertJacobsonCDC: > I should also introduce a local scope for the temporaries.

Oh, I do! 😄

swo:
I think the nicest thing for me is to avoid writing `context` four times, to avoid writing the `|move|` and the braces, etc. The single idea "at X time in the future, do Y" will be a single line of code. This is great!

re: terminology. You can see my bias from #836. My expectation is that most plans in our models will be from event-driven loops, and so use relative time. I'm not sure what that means for method/macro names. I can see why, if there's a single `add_plan` method, it must refer to absolute time, and so an `add_plan_relative!` macro should be named explicitly. But maybe that means we want an `add_plan_absolute!` or something, to be explicit both ways.

I also see how `add_plan_relative` is good, because then you don't introduce the new term "schedule." On the other hand, a lot of code seems to have functions called `schedule_something()`, so I wonder if there's garden path terminology here.

RobertJacobsonCDC:
> My expectation is that most plans in our models will be from event-driven loops, and so use relative time.

Yeah, I think we figured it's easy enough for client code to do the absolute &rarr; relative conversion, and so we favored a smaller set of orthogonal APIs. I do like the word _schedule_ better than _add_, personally, but... now we have an inconsistency.

Personally I think the fact that this macro lets the user avoid a manual construction of awkward closure syntax is its most convincing justification. If you haven't encountered closures / lambdas in other languages, they're kind of weird.

The more model code we write the better we'll understand what all the little paper cuts are.

PR #922, opened 2026-05-26
chore: update mdbook (0.5.3) versions

### LINKED ISSUE #763: Update the MDBook toolchain
We pin the versions of our MDBook toolchain in `mise.toml`:

```toml
[tools]
# ...
"cargo:mdbook" = "0.4.52"
"cargo:mdbook-callouts" = "0.2.2"
"cargo:mdbook-inline-highlighting" = "1.0.0"
```

MDBook has been on v5 for a while, and the two plugins have recently been updated to be compatible with that series. We can update these versions all at once. Keep them pinned to known working versions, though, so CI doesn't randomly break when one of them has a new release.

cdc-as81:
## Summary

Fix mdBook HTML checker warnings caused by inline code spans containing angle-bracket placeholders or Rust generics, and update the mdBook-related tooling versions in `mise.toml`.

## Changes

- Updated mdBook tooling:
  - `mdbook`: `0.4.52` -> `0.5.3`
  - `mdbook-callouts`: `0.2.2` -> `0.3.0`
  - `mdbook-inline-highlighting`: `1.0.0` -> `2.0.0`
- Escaped inline generic examples in:
  - `docs/book/src/topics/indexing.md`
  - `docs/book/src/topics/handling-errors.md`
  - `docs/book/src/migration_guide.md`

### DIFF
--- a/docs/book/src/cli-usage.md
+++ b/docs/book/src/cli-usage.md
@@ -4,7 +4,7 @@ This document contains the help content for the `ixa` command-line program.
 
 ## `ixa`
 
-Default cli arguments for ixa runner
+Default cli arguments for Ixa runner
 
 **Usage:** `ixa [OPTIONS]`
 
--- a/docs/book/src/migration_guide.md
+++ b/docs/book/src/migration_guide.md
@@ -67,7 +67,7 @@ pub enum InfectionStatus {
     Recovered,
 }
 
-// Implements `Property\<Person>` for an existing type.
+// Implements `Property<Person>` for an existing type.
 impl_property!(
     InfectionStatus,
     Person,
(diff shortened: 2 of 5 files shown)


RobertJacobsonCDC (changes requested):
The correct fix is to escape the opening `<` character like we do elsewhere:

```markdown
...have functions that return `Result\<U, V>` to indicate...
```

Manually constructing a `<code>` block is fragile, hard to read, and breaks highlighting of inline code.

docs/book/src/cli-usage.md
RobertJacobsonCDC, line 16: This `<CONFIG>` is still eaten by the converter and appears to be the cause of the remaining warning. This file is generated, so it's not a matter of simply escaping the `<` in this case. I personally don't think it's a big deal, and considering the fact that it's not clear what the fix is in this case, maybe we consider it out of scope for this PR. Just thought I'd point it out.

RobertJacobsonCDC (approved):

docs/book/src/cli-usage.md
RobertJacobsonCDC: I also don't see a way to exclude `cli-usage.md` from preprocessing by `mdbook-inline-highlighting`. 🤷

PR #923, opened 2026-05-26
chore: pin rust toolchain and print mise rust versions

### LINKED ISSUE #725: Use rust-toolchain + dependabot / upgrade script
It's annoying to break main because of rust changes, i think we should probably have a cannnical version at least for ci and upgrade it monthly maybe

cdc-as81:
## Summary

- Pin the repo Rust toolchain to `1.95.0` with `rust-toolchain.toml`
- Install/export the pinned Rust toolchain in CI and release workflows
- Add monthly Dependabot updates for the Rust toolchain
- Add mise version printing before Rust/Cargo-related tasks

### DIFF
--- a/.github/workflows/check-test.yaml
+++ b/.github/workflows/check-test.yaml
@@ -40,9 +40,6 @@ jobs:
       - name: Setup environment
         uses: ./.github/actions/setup-env
 
-      - name: Install wasm target
-        run: rustup target add wasm32-unknown-unknown
-
       - name: WASM tests
         run: mise run test:wasm
         env:
--- a/.github/actions/setup-env/action.yml
+++ b/.github/actions/setup-env/action.yml
@@ -20,8 +20,6 @@ runs:
       shell: bash
       run: |
         rustup toolchain install
-        active_toolchain="$(rustup show active-toolchain | awk '{print $1}')"
-        echo "RUSTUP_TOOLCHAIN=${active_toolchain}" >> "$GITHUB_ENV"
 
     - name: Install cargo-binstall
       uses: cargo-bins/cargo-binstall@v1.15.6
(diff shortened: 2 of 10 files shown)


.github/workflows/check-test.yaml
```diff
@@ -40,9 +40,6 @@ jobs:
       - name: Setup environment
         uses: ./.github/actions/setup-env
 
-      - name: Install wasm target
-        run: rustup target add wasm32-unknown-unknown
```
RobertJacobsonCDC, line 44: We don't need this here because it's now specified in `rust-toolchain.toml` and installed by Cargo automatically when needed.

On the other hand, when Cargo invokes `rustup` to install the Wasm target, will it get saved to the GitHub cache?

RobertJacobsonCDC: I guess a similar thing happens in the `fix:rust` task that invokes `cargo +nightly`: the nightly toolchain is installed if it doesn't exist yet. Is it saved to the cache?

RobertJacobsonCDC (approved):
We should probably pin to 1.96.0, which was just released.

.github/workflows/check-test.yaml
cdc-as81: Yes, it should be cached. The setup action caches `~/.rustup/toolchains/`, and the cache key includes `rust-toolchain.toml` and `mise.toml`. The Wasm target from `rust-toolchain.toml` is installed by `rustup toolchain install` during setup, and installed targets live under the toolchain directory, so they are included in that cache.

Same for nightly: `setup-env` runs `mise run install-nightly`, which installs the nightly toolchain/components under `~/.rustup/toolchains/`, so those are cached too. An exact cache hit will not rewrite the cache, but changes to `rust-toolchain.toml` or `mise.toml` produce a new primary key, so target/toolchain changes get saved on the next successful run.

PR #924, opened 2026-05-27
chore: add mise command to generate test coverage

### LINKED ISSUE #925: Add test coverage

cdc-as81:
## Summary

- Add `cargo-llvm-cov` as a mise-managed tool.
- Add a `mise run coverage` task for local Rust coverage generation across the workspace and all targets.
- Generate both summary and HTML coverage reports, with HTML output available under `target/llvm-cov/html/`.
- Ignore `*.profraw` coverage artifacts so raw profiling files do not show up as untracked files.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -24,6 +24,12 @@ run = [
   "rustup component add --toolchain nightly-x86_64-unknown-linux-gnu rustfmt clippy",
 ]
 
+[tasks.'install-coverage']
+description = "Install Rust coverage tools"
+run = [
+  "rustup component add llvm-tools-preview",
+]
+
 [tasks.build]
 description = "Build default targets. Run with --release for a release build."
 run = "cargo build --workspace"
@@ -86,8 +92,8 @@ run = "cargo test --no-default-features"
 
 [tasks.coverage]
 description = "Generate local Rust test coverage reports"
+depends = ["install-coverage"]
 run = [
-  "rustup component add llvm-tools-preview",
   "cargo llvm-cov clean --workspace",
   "cargo llvm-cov --workspace --all-targets --no-report",
   "cargo llvm-cov report --summary-only",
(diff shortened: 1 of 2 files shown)


mise.toml
```diff
@@ -83,6 +84,18 @@ run = "pnpm exec playwright test"
...
+description = "Generate local Rust test coverage reports"
+run = [
+  "rustup component add llvm-tools-preview",
+  "cargo llvm-cov clean --workspace",
+  "cargo llvm-cov --workspace --all-targets --no-report",
+  "cargo llvm-cov report --summary-only",
+  "cargo llvm-cov report --html",
+  "cargo clean -p ixa-integration-tests",
+  "find . -name '*.profraw' ! -path './target/*' -delete",
+]
```
RobertJacobsonCDC, line 101: Can you explain what all of these do? For example:

- Why is tool installation incorporated into this command? Don't we use the mechanism built into mise for this? Should it be a separate command?
- Similarly, it looks like this command also performs its own `clean` step. Why? I don't think we do this with other `mise` commands. Are they just intermediate artifacts that cannot be reused or what?
- Do we need both `cargo llvm-cov report --summary-only` and `cargo llvm-cov report --html`?
- Why is `cargo clean -p ixa-integration-tests` a part of this? It seems unrelated.

RobertJacobsonCDC (changes requested):
Please clarify what this PR does.

mise.toml
cdc-as81: I separated the install command.

The coverage flow is:

- `cargo llvm-cov clean --workspace`: removes stale coverage profiles/artifacts so a new report is not polluted by old `.profraw` data.
- `cargo llvm-cov --workspace --all-targets --no-report`: runs the workspace tests with coverage instrumentation, but defers report generation.
- `cargo llvm-cov report --summary-only`: prints the terminal summary table.
- `cargo llvm-cov report --html`: writes the browsable HTML report under `target/llvm-cov/html`.

We do not strictly need both report commands. They do not rerun tests, but they do produce different outputs: console summary vs. HTML. we can keep one of them.

RobertJacobsonCDC (approved):
This will be interesting to play with.

PR #927, opened 2026-05-27
chore: Factored out `impl_property_eq_hash!` as user-facing macro

### LINKED ISSUE #911: Split out generating `Eq` & `Hash` for property type into a separate user-facing macro
Split out the macro code that generates `PartialEq` / `Eq` and `Hash` implementations for a property type from `define_*_property`. That way if client code needs to bring their own pre-defined type we can still generate these implementations for them.

RobertJacobsonCDC:
This PR introduces new user-facing macros:

| Macro                    | Purpose                                      |
|:-------------------------|:---------------------------------------------|
| `impl_property_eq!`      | Implements `PartialEq` and `Eq`              |
| `impl_property_hash!`    | Implements `Hash`                            |
| `impl_property_eq_hash!` | Convenience macro that invokes both macros   |

I also added language to the Properties chapter of the Ixa Book explaining when to use them.

The point of these is, client code can now use them to generate implementations for these traits when they cannot be derived in the case they define the property type themselves instead of using `define_*_property!` macros. 

These macros previously existed as "internal subcommands" of the `define_property!` macro, so from that point of view this isn't exactly new functionality. It's just that it's now intended to be exposed to the end user.


src/macros/property_impl.rs
```diff
@@ -47,6 +47,13 @@ of the `#[derive(..)]` boilerplate, so you have to remember to derive or impleme
 required by [`AnyProperty`][crate::entity::property::AnyProperty] for your type: `Copy`, `Clone`,
 `Debug`, `PartialEq`, `Eq`, and `Hash`.
 
+If the type cannot derive `PartialEq` / `Eq` or `Hash`, for example because it contains `f32` or
+`f64`, use [`impl_property_eq!`][macro@crate::impl_property_eq],
+[`impl_property_hash!`][macro@crate::impl_property_hash], or
+[`impl_property_eq_hash!`][macro@crate::impl_property_eq_hash] to generate those implementations
+for the manually declared type. These macros require `rkyv::Archive` and `rkyv::Serialize` derives
```
k88hudson-cfa: Do you want `ixa::rkyv` and `ixa::rkyv::Serialize` here maybe?

k88hudson-cfa (approved):
LGTM

PR #932, opened 2026-06-02
chore: add gated MSRV verification

### LINKED ISSUE #930: Can we add a CI check for min rust version
need to run in CI, if dependencies changed
need to be add to mise prepush
use  cargo msrv

cdc-as81:
## Summary

- Add `cargo-msrv` to the mise toolchain and define `mise run check:msrv` to verify the declared `rust-version`.
- Add a CI job that detects `Cargo.toml` changes and runs the MSRV check only when package metadata changes.
- Wire the MSRV job into `all-green`, while allowing it to be skipped when no `Cargo.toml` files changed.
- Keep `check-all`/precommit fast by excluding MSRV there, but include MSRV in `prepush`.
- Update display assertions in `property_impl.rs` to use `.to_string()`.

### DIFF
--- a/.github/workflows/check-test.yaml
+++ b/.github/workflows/check-test.yaml
@@ -89,61 +89,6 @@ jobs:
       - name: Build docs
         run: mise run docs
 
-  cargo-toml-changes:
-    runs-on: ubuntu-latest
-    outputs:
-      changed: ${{ steps.changed.outputs.changed }}
-
-    steps:
-      - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0
-
-      - name: Check for Cargo.toml changes
-        id: changed
-        shell: bash
-        run: |
-          set -euo pipefail
-
-          case "${{ github.event_name }}" in
-            pull_request)
-              base="${{ github.event.pull_request.base.sha }}"
-              head="${{ github.sha }}"
-              ;;
-            push)
-              base="${{ github.event.before }}"
-              head="${{ github.sha }}"
-              ;;
-            *)
-              echo "changed=true" >> "${GITHUB_OUTPUT}"
-              exit 0
-              ;;
-          esac
-
-          if [[ "${base}" == "0000000000000000000000000000000000000000" ]]; then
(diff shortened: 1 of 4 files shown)


mise.toml
```diff
@@ -168,7 +169,7 @@ depends = "build:hyperfine"
 run = './target/release/hyperfine'
 
 [tasks.check-all]
-run = ["mise run check:*"]
+run = ["mise run check:rust", "mise run check:markdown"]
```
k88hudson-cfa: if you don't want `msrv` to run in the check group, it would be better to call it `tasks.msrv` or `tasks.check-msrv` rather than breaking the group semantics

.github/workflows/check-test.yaml
```diff
@@ -89,6 +89,61 @@ jobs:
...
+    runs-on: ubuntu-latest
+    outputs:
+      changed: ${{ steps.changed.outputs.changed }}
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Check for Cargo.toml changes
```
k88hudson-cfa: You could consider putting this in its own workflow and using:

```yaml
pull_request:
      paths: ['**/Cargo.toml']
```

mise.toml
```diff
@@ -185,6 +186,10 @@ run = [
   "cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings",
 ]
 
+[tasks.'check:msrv']
+description = "Verify the declared minimum supported Rust version"
+run = 'XDG_CACHE_HOME="$PWD/target/msrv-cache" cargo msrv --no-log verify --all-features'
```
k88hudson-cfa: What's the purpose of setting`XDG_CACHE_HOME`? I think it's a bit weird to have something platform-specific in a mise command / silently do nothing otherwise, if you need it specifically for CI I would set it there

k88hudson-cfa (changes requested):

k88hudson-cfa (approved):

PR #934, opened 2026-06-02
fix: Use schedule_relative!

swo:
- Use `schedule_relative!` in docs
- Incorporate @RobertJacobsonCDC's fix of `schedule_relative!` macro from #954

I set out to use the fruits of #921. Overall, I think it's a nice improvement. I find it easier to avoid the `get_current_time()`, for example.

When there was a well-established handler, things look really slick, like in `examples/births-deaths/src/population_manager.rs`. (It also clarifies now that you're "scheduling" a function to "do" something.)

Perhaps not surprisingly, there are a number of places where the `add_plan` closure had a `context.do_something` form, so `schedule_relative!` also needs a closure (eg, [this line](https://github.com/CDCgov/ixa/compare/swo_schedule_relative?expand=1#diff-55f1d1e67b694b7f9385b5ec5bae2eb44f7d746efa684b9e939922f6c479204cR22)).

I think this is OK, since it encourages a practice where you have `schedule_relative!` call the handler, which is a `do_something(context, ...)` function. And that's a nice place to put the traces.

I'm calling this draft because there are some compilation errors that made me scratch my head. There's some problem with using `schedule_relative!` twice in a row, maybe?

### DIFF
--- a/docs/book/models/disease_model/src/infection_manager.rs
+++ b/docs/book/models/disease_model/src/infection_manager.rs
@@ -1,8 +1,8 @@
 use ixa::prelude::*;
 use rand_distr::Exp;
 
-use crate::INFECTION_DURATION;
 use crate::people::{InfectionStatus, Person, PersonId};
+use crate::INFECTION_DURATION;
 
 // ANCHOR: infection_status_event
 pub type InfectionStatusEvent = PropertyChangeEvent<Person, InfectionStatus>;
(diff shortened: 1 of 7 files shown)


RobertJacobsonCDC:
The `schedule_relative!` macro still has an outstanding issue for the case of a closure handler: it hits a weird edge case in the Rust compiler that prevents the compiler from being any to infer the parameter type(s) of the closure (the type of `context`), and so client code is required to provide type annotations. 

We decided it's better to merge the "moved borrow" fix and the examples / docs usages in this PR and then address the closure parameter type inference separately as issue #956. The workaround, namely providing the type annotations, is really annoying but not catastrophic, and the `schedule_relative!` macro is new, so it's not a breaking change for v2.0.

RobertJacobsonCDC (approved):
Approved, but see last comment.

PR #937, opened 2026-06-03
feat: Replace universal `Property: Eq + Hash` constraint with `IndexableProperty`

### LINKED ISSUE #909: Split Property/IndexableProperty
If we split Property from IndexableProperty in the code we could derive IndexableProperty for all Property + Eq + Hash. This would solve the issue of implementing properties that have floats in them

RobertJacobsonCDC:
This PR relaxes the previously universally required `Property: Eq + Hash` constraint in favor of an `IndexableProperty` constraint that applies only when a property is actually indexed.

It is organized into three main commits:

1. Remove sharing indexes between properties. Previously, "equivalent" properties (multi-properties with the same components in different orders) would all share a single index. This commit removes this ability to share indexes between properties, warns during multi-property registration if another equivalent multi-property is already registered, and panics if you try to index a property for which an equivalent property exists (because the index could be inaccessible).

2. Remove `Property::CanonicalValue` and supporting code. The associated type and the conversion methods to and from `CanonicalValue` require a procedural macro to be generated. Without the shared index use case, we don't need all this infrastructure. 

   The only remaining use case is that of resolving a query to an indexed multi-property. The only tricky part is converting the query to the multi-property type in order to look up the value in the underlying `HashMap<P, IndexSet>`. Instead of statically generated conversion code, we use a slice of `&Any` values, sort them, then downcast into the right tuple type. This code can be generated with a `macro_rules` macro.

3. Remove the `Eq + Hash` requirement for all `Property` types, introduce `IndexableProperty`, and only require `IndexableProperty` for `Context::index_property` execution path.

# Questions and Issues

- From the perspective of client code, the only thing that has changed (I think) is that a property doesn't need to implement `Eq + Hash` when it doesn't need it. The behavior of the  `define_*_property` methods hasn't changed: by default they still include `Eq` and `Hash` in the list of derives unless the `impl_eq_hash = ...` parameter is provided. 
- A fortunate side effect of this refactor is that type aliases no longer break the query / indexing system (#843). The reason is, the names of the types are no longer used to do query &rarr; multi-property index resolution. (We use `TypeId` instead.)

---

**To Do before merging:**

- [x] Run performance benchmarks to check for regressions
- [x] Update some of the language in the Properties chapter of the book
- [x] Rebase


src/entity/query/query_impls.rs
```diff
@@ -4,6 +4,7 @@ use seq_macro::seq;
 
 use crate::entity::entity_set::{EntitySet, EntitySetIterator, SourceSet};
 use crate::entity::index::IndexSetResult;
+use crate::entity::multi_property::multi_property_id_for_property_type_id;
```
k88hudson-cfa: Looks like this is unused

src/entity/entity_set/source_set.rs
```diff
@@ -317,7 +332,9 @@ impl<'a, E: Entity> PartialEq for SourceSet<'a, E> {
                 left == right || (left.is_empty() && right.is_empty())
             }
             (Self::IndexSet(left), Self::IndexSet(right)) => std::ptr::eq(&**left, &**right),
-            (Self::PropertySet(left), Self::PropertySet(right)) => left.id() == right.id(),
+            (Self::PropertySet(left), Self::PropertySet(right)) => {
+                left.same_logical_query(&**right)
```
k88hudson-cfa, line 336: Maybe we could add a test that unions a property query with itself and checks it dedups to one source? A second case with equivalent multi-properties in different orders would also cover the cross-type stuff

k88hudson-cfa (changes requested):

k88hudson-cfa:
Let's make sure we have a plan for breaking changes with canonical @RobertJacobsonCDC

k88hudson-cfa:
This just needs a rebase looks like

k88hudson-cfa (approved):

PR #938, opened 2026-06-03
chore: Derive Copy and Clone for IxaEvent types

### LINKED ISSUE #399: Can `#[derive(IxaEvent)]` also derive `Copy`?
All the cases where a method can take a type that implements `IxaEvent` also require the type to implement `Copy`. Can `#[derive(IxaEvent)]` also derive `Copy` for easier ergonomics?

These methods also require `IxaEvent` to be `'static`, so just like `PersonProperty` was made to require the type be `'static`, could a similar thing be done here with the `IxaEvent` trait?

cdc-as81:
## Summary

- Make `IxaEvent` require `Copy + 'static` directly.
- Update `#[derive(IxaEvent)]` to generate `Copy` and `Clone` impls.
- Remove manual `Copy`/`Clone` impls from entity event types.
- Update event tests and plugin-context test fixtures to rely on `#[derive(IxaEvent)]` alone.
- Add coverage for generic event types where the generic parameter is not `Copy`.

### DIFF
--- a/ixa-derive/src/lib.rs
+++ b/ixa-derive/src/lib.rs
@@ -5,7 +5,7 @@ mod utilities;
 
 use proc_macro::TokenStream;
 use quote::quote;
-use syn::{parse_macro_input, DeriveInput};
+use syn::{parse_macro_input, parse_quote, DeriveInput};
 
 #[proc_macro_derive(IxaEvent)]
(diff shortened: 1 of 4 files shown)


RobertJacobsonCDC (approved):
Works for me.

PR #943, opened 2026-06-05
fix: bench_filter_indexed_entity not empty

### LINKED ISSUE #941: Fix "bench_filter_indexed_entity" so the result set isn't empty
The "bench_filter_indexed_entity" benchmark in `ixa-bench/criterion/large_dataset.rs` is intended to measure the performance of `filter_entities` for the case of an indexed query (jointly indexed via multi-property)

```rust
        let entity_ids: Vec<EntityId<Person>> = context.get_entity_iterator().collect();
        criterion.bench_function("bench_filter_indexed_entity", |bencher| {
            bencher.iter_batched(
                || entity_ids.clone(),
                |mut entities| {
                    context.filter_entities(
                        &mut entities,
                        with!(Person, Age(30u8), SchoolId(1u32), WorkplaceId(1u32)),
                    );
                },
                BatchSize::SmallInput,
            );
        });
```

It turns out that when `age.0 == 30` we always have `SchoolId(0)`. Therefore, the index bucket is always empty, and we always hit the empty bucket optimized fast path for this. Now, there's nothing wrong with measuring this optimized fast path itself, but that isn't the intention of this benchmark.

Either change age to be in the range `5..=18` or change the `SchoolId` to `SchoolId(0)`. Whichever you change, make sure the result set is nonempty. 

This change will invalidate historical timings for this benchmark. I see three alternatives:

1. Leave the old values. We'll just have to know that the performance regression isn't real.
2. Create a new benchmark that retains the old behavior of hitting the "empty" fast path, and copy the historical timings for this benchmark to be timings for the "new" benchmark. (Said another way, rename this benchmark, and add a new benchmark for the nonempty case under the old name. Historical data belong with the old renamed benchmark.)
3. Just delete historical data.

I kind of like the idea of keeping a benchmark for the empty fast path, but it's not exactly a very important benchmark, so I don't mind if we decide to just delete it.

cdc-as81:
## Summary

Fixes `bench_filter_indexed_entity` so it measures a nonempty jointly indexed multi-property query instead of the empty-bucket fast path.

The benchmark now filters on:

```rust
with!(Person, Age(5u8), SchoolId(1u32), WorkplaceId(0u32))

### DIFF
--- a/ixa-bench/criterion/large_dataset.rs
+++ b/ixa-bench/criterion/large_dataset.rs
@@ -126,13 +126,23 @@ pub fn criterion_benchmark(criterion: &mut Criterion) {
 
     {
         let entity_ids: Vec<EntityId<Person>> = context.get_entity_iterator().collect();
+        let mut matching_entities = entity_ids.clone();
+        context.filter_entities(
+            &mut matching_entities,
+            with!(Person, Age(5u8), SchoolId(1u32), WorkplaceId(0u32)),
+        );
+        assert!(
+            !matching_entities.is_empty(),
+            "bench_filter_indexed_entity query must have a nonempty result set"
+        );
+
         criterion.bench_function("bench_filter_indexed_entity", |bencher| {
             bencher.iter_batched(
                 || entity_ids.clone(),
                 |mut entities| {
                     context.filter_entities(
                         &mut entities,
-                        with!(Person, Age(30u8), SchoolId(1u32), WorkplaceId(1u32)),
+                        with!(Person, Age(5u8), SchoolId(1u32), WorkplaceId(0u32)),
                     );
                 },
                 BatchSize::SmallInput,

### CODE REVIEW

File: ixa-bench/criterion/large_dataset.rs [hunk 1]
```diff
@@ -126,13 +126,23 @@ pub fn criterion_benchmark(criterion: &mut Criterion) {
 
     {
         let entity_ids: Vec<EntityId<Person>> = context.get_entity_iterator().collect();
+        let mut matching_entities = entity_ids.clone();
+        context.filter_entities(
+            &mut matching_entities,
+            with!(Person, Age(5u8), SchoolId(1u32), WorkplaceId(0u32)),
+        );
+        assert!(
+            !matching_entities.is_empty(),
```
RobertJacobsonCDC, line 135: Won't this fail probabilistically? Maybe there's a better way to choose the query values.

RobertJacobsonCDC (changes requested):
What are we going to do with the historical benchmark values as discussed in [the parent issue](https://github.com/CDCgov/ixa/issues/941)? If we are going to move them to a new benchmark, it makes sense to me to add that benchmark on this PR.

RobertJacobsonCDC:
Just a note to point out that the 10x "performance regression" for this benchmark is expected and not a problem. We should, however, decide what to do with the historical benchmark data. We could delete it or move it to a benchmark for the empty fast path or...?

ixa-bench/criterion/large_dataset.rs
cdc-as81: The assertion could start failing if the seed, population size, generator logic, rand behavior/version, or benchmark parameters change.
but that is the same with other benchmarks.

RobertJacobsonCDC (approved):

PR #946, opened 2026-06-08
chore: deploy docs from release tags

### LINKED ISSUE #266: Render docs and book from the release branch instead of main.
We render the docs and the book in CI with CDCgov/ixa/.github/workflows/docs.yml whenever we push to main. Once we have #236 we should instead render them when we push to release. That way the online docs and book will reflect the current release.

cdc-as81:
## Summary

- Deploy GitHub Pages docs only from `ixa-v*` release tag pushes instead of `main`
- Add manual workflow dispatch with a required release tag input
- Ensure manual runs checkout `refs/tags/<tag>` and reject non-`ixa-v*` inputs

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -3,12 +3,18 @@
 name: Deploy doc to Pages
 
 on:
-  # Runs on pushes targeting the default branch
+  # Runs on pushes of release tags.
   push:
-    branches: main
+    tags:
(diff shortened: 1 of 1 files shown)


k88hudson-cfa (approved):
I think this is reasonable for now; we could switch to a multiple version publishing mechanism at some point maybe to also see changes on main (since this will be gated on updating docs in between releases)

PR #948, opened 2026-06-09
chore: require entity-first multi-property definitions

### LINKED ISSUE #942: Prohibit degenerate multi-properties
A multi-property is a tuple of properties that implements `Property` using `define_multi_property!`. Up until now we have allowed multi-properties that are just singleton tuples of some other property, that is, multi-properties of the form `(P, )`, even though it doesn't make sense to do this, because we thought it didn't matter. But it turns out there is an edge case for singleton queries where we assume the single property is not itself a multi-property. Prohibiting degenerate multi-properties that have a single component 

- would make that assumption true 
- prevents client code from creating confusing multi-properties

Related, we should probably prevent the syntax error of accidentally writing an incorrect `define_multi_property!` invocation as

```rust
define_multi_property!(Person, (Age, InfectionStatus));
```

that should be

```rust
define_multi_property!(Person, Age, InfectionStatus);
```

(Or make both invocations expand to the same thing.)

cdc-as81:
## Summary

- require `define_multi_property!(Entity, PropertyA, PropertyB, ...)`
- reject empty, singleton, tuple-first, and nested tuple multi-property definitions
- update repo call sites, docs, benchmarks, and tests to the new syntax


k88hudson-cfa:
This is a breaking change, let's leave out of 2.0 and discuss for 3.0

docs/book/src/topics/indexing.md
```diff
@@ -11,7 +11,7 @@ context.index_property::<Person, Age>();
 
 // For multi-indexes
 // Where properties are defined:
-define_multi_property!((Name, Age, Weight), Person);
+define_multi_property!(Person, Name, Age, Weight);
```
k88hudson-cfa: We discussed as a group and we figure we wanted to go with:
```rs
define_multi_property!(Person, (Name, Age, Weight));
```

k88hudson-cfa (commented):
Let's merge this post 2.0

RobertJacobsonCDC (approved):
This looks a lot simpler than I expected it to be! Nice.

Issue #705, opened 2026-02-03
Add query tracking to profiler


k88hudson-cfa:
It might be helpful to add something like this for queries to profiler output. Ideally I think these would be great as markers in `samply` if that gets added, but for now maybe something like this (+ maybe % of runtime) 
```
Query          Count         Total           Mean           Min           Max
--------------------------------------------------------------------------------
(Age, County)     150       12.34ms        82.27us       10.12us      250.45us
(Age)            1000        5.67ms         5.67us        1.23us       45.67us
All                50        1.23ms  
```

The point would be to debug which queries are taking the most time and inform if changes to which custom data structures are needed etc. 

Having some kind of display value implemented for queries will be helpful for this

Issue #280, opened 2025-03-17
Optimistically create indexes or warn


k88hudson-cfa:
Right now all index creation is manual. We could create indexes if you run a query a certain number of times, or warn people maybe for a certain log level?

PR #951, opened 2026-06-10
chore: Index multi-properties by default

### LINKED ISSUE #672: Multi-properties should be indexed by default.
Currently the only use-cases of multi-properties is 
- to index several properties jointly. (Also, to _count_ combinations of different values, which is an index function.)
- to subscribe to property change events for a set of properties as if it were a single property. 

Thus, it probably makes sense to make them indexed automatically.

Right now, you can _define_ a multi-property:

```rust
define_multi_property!((Name, Age, Weight), Person);
```

And then, separately, you set the multi-property as indexed in a specific `Context` instance:

```rust
context.index_property::<Person, (Name, Age, Weight)>();
```

Why make the user do this?

It's not clear how to implement this. Right now we have no mechanism for a property to know if it's a multi-property, or more generally if it should be indexed by default. If we built this into the type somehow (e.g. `fn Property<E>::default_indexed() -> bool`), we could do a check in the `PropertyValueStoreCore<E: Entity, P: Property<E>>`. Alternatively, we could have a mechanism for creating different constructors for `PropertyValueStoreCore<E: Entity, P: Property<E>>`, one that indexes and another that doesn't. [Entities Design Notes](https://github.com/CDCgov/ixa/wiki/Entities-Design-Notes#multi-properties) favors a method `Property<E>::index_by_default()`.

cdc-as81:
## Summary

  - Make `define_multi_property!` create a `FullIndex` by default via
    `Property::default_index_type`.
  - Allow a multi-property definition to opt into `ValueCountIndex` or opt out
    with `PropertyIndexType::Unindexed`.
  - Preserve #937’s no-shared-storage design: equivalent multi-properties remain
    distinct, while the first registered representative handles query routing and
    is the only one eligible for a default index.
  - Keep explicit `Context::index_property` calls compatible with populated
    multi-properties, including default-indexed and unindexed definitions.
  - Update documentation and regression coverage for defaults, overrides,
    equivalent property ordering, and explicit indexing.


RobertJacobsonCDC:
I think this needs to be blocked by #937. Also, it's not clear which kind of index should be created. We should discuss how this interacts with these issues, too:

- #280

- #283

- #705

cdc-as81:
lets add an option to the macro to set the index type.
property_store.set_property_indexed::<P>(PropertyIndexType::FullIndex);

docs/book/src/topics/properties.md
```diff
@@ -679,14 +679,25 @@ define_property!(
...
+    (Age, InfectionStatus),
+    ixa::entity::PropertyIndexType::Unindexed
+);
 ```
 
-For a deeper discussion of when to create multi-property indexes, see [Indexing](indexing.md). Because the components
-of a multi-property are already required to be properties, multi-properties usually "just work". If a multi-property
-will be indexed, each component value must also support `Eq` and `Hash`; this mainly matters for components containing
-plain `f32` or `f64`.
+Use the underlying property names in `define_multi_property!`, not type aliases. For a deeper discussion of when to
```
RobertJacobsonCDC: > Use the underlying property names in `define_multi_property!`, not type aliases.

This issue was resolved by #937.

> Each component value must also support `Eq` and `Hash`; this mainly matters for components containing plain `f32` or `f64`.

Unless you opt out of indexing by passing `PropertyIndexType::Unindexed` as a third parameter to `define_multi_property!`.

docs/book/src/topics/indexing.md
```diff
@@ -191,23 +191,42 @@ might look like this:
 | `(30, recovered)`             | `\[4, 8, 35, 36]`                |
 
 Ixa hides the boilerplate required for creating a multi-index with the macro
-`define_multi_property!`:
+`define_multi_property!`. Multi-properties create a full index automatically:
```
RobertJacobsonCDC: > Multi-properties create a full index automatically:

I would add the phrase "by default" everywhere we say this. This clues the reader into the fact that this default can be over-ridden.

src/entity/query/mod.rs
```diff
@@ -889,8 +889,6 @@ mod tests {
         let mut context = Context::new();
         let mut people = Vec::new();
 
-        context.index_property::<Person, (Age, County)>();
```
RobertJacobsonCDC, line 892: I wonder if we should have a test that _also_ explicitly indexes a multi-property even though multi-properties are indexed by default. The reason is, we don't want this to break existing code, and (as far as I know) pretty much all existing code will have an explicit call to `context.index_property` for every multi-property precisely because we formerly required indexing to be done explicitly.

src/entity/context_extension.rs
```diff
@@ -1134,13 +1134,7 @@ mod tests {
...
                 .unwrap();
         }
-        context.index_property::<Person, InfectionStatusVaccinated>();
-        // Force an index build by running a query.
-        let _ = context.query_result_iterator(with!(
-            Person,
-            InfectionStatus::Susceptible,
-            Vaccinated(true)
-        ));
+        assert!(context.is_property_indexed::<Person, InfectionStatusVaccinated>());
```
RobertJacobsonCDC: This line is _not_ equivalent to the code it is replacing. So you are changing what this test is testing. Is this intentional?

src/entity/property_store.rs
```diff
@@ -122,6 +122,22 @@ pub(super) fn get_property_dependents_static<E: Entity>(property_index: usize) -
 /// data/metadata is associated with the [`crate::entity::property::Property`] if it doesn't already exist. In
 /// our use case, this method is called in the `ctor` function of each `Property<E>` type.
 pub fn add_to_property_registry<E: Entity, P: Property<E>>() {
+    add_to_property_registry_with_constructor::<E, P>(PropertyValueStoreCore::<E, P>::new_boxed);
```
RobertJacobsonCDC: What is the point of having two separate functions `add_to_property_registry` and `add_to_property_registry_with_constructor`?

And a related question: Can't we just have `PropertyValueStoreCore::<E, P>::new_boxed` work the way `PropertyValueStoreCore::<E, P>::new_boxed_with_default_index` works? Why do we need two distinct constructors? (I'm not saying we don't, I'm just saying if we do the reason isn't clear to me.)

src/entity/property_value_store_core.rs
```diff
@@ -182,6 +182,28 @@ impl<E: Entity, P: Property<E>> PropertyValueStoreCore<E, P> {
     }
 }
 
+impl<E: Entity, P: IndexableProperty<E>> PropertyValueStoreCore<E, P> {
+    fn default_index() -> Option<Box<dyn PropertyIndex<E, P>>> {
```
RobertJacobsonCDC: This function has a single usage. Inline it.

src/entity/property_value_store_core.rs
```diff
@@ -182,6 +182,28 @@ impl<E: Entity, P: Property<E>> PropertyValueStoreCore<E, P> {
...
+        if P::index_id() != P::id() {
```
RobertJacobsonCDC: I don't understand this check. Non multi-properties use the default trait implementation for `P::default_index_type` (I think), so this looks superfluous to me. And it precludes using `default_index_type` for non multi-properties in the future, which seems like a premature design decision at this point.

RobertJacobsonCDC (changes requested):
PR #937 removed sharing indexes between properties. This PR appears to reintroduce shared properties. I don't think we want to do this. Instead, let's retain the warning and error behavior explicitly introduced in \#937. I think this will simplify this PR quite a bit.

There's also the philosophical question of whether we want to expose `Property::default_index_type` as public API. So for example, we could add a parameter to `define_*_property!` / `impl_*_property` that lets client code set the return value of `Property::default_index_type`. But let's defer that question to another issue / PR. (I don't think we have a use case for it, so let's not pay the cost of added code complexity.)

docs/book/src/topics/properties.md
cdc-as81: Fixed. `define_multi_property!` now accepts component type aliases; its rustdoc shows supported usage, and a regression test verifies an alias-defined multi-property is selected for an indexed query.

cdc-as81:
Fixed. Equivalent multi-properties now use distinct storage; only the registered representative receives the default index.

docs/book/src/topics/indexing.md
cdc-as81: Fixed. The indexing and properties documentation now describes these indexes as defaults and explains the representative-only behavior.

src/entity/query/mod.rs
cdc-as81: Fixed. Added a regression test that explicitly indexes a populated default-indexed multi-property and verifies the full index and matching results.

src/entity/context_extension.rs
cdc-as81: Fixed. Expanded the unindexed-default multi-property test to populate entities before explicit indexing, then verify the completed full index returns exactly the matching entity IDs.

src/entity/property_store.rs
cdc-as81: No change. `new_boxed` must support every `Property`, including properties without `Eq + Hash`, while default-index construction requires `IndexableProperty`. The shared registry helper already contains all registration logic; the two entry points select the constructor compatible with those bounds.

src/entity/property_value_store_core.rs
cdc-as81: Fixed. Inlined the default-index match in its sole caller.

src/entity/property_value_store_core.rs
cdc-as81: Fixed as part of issue 1. Removed `index_id()` routing and the representative-ID guard; default index creation now depends only on `P::default_index_type()`.

RobertJacobsonCDC:
This PR is a lot more invasive than I expected, and I can't tell if it is because I don't fully understand some edge cases or if the complexity is genuinely unnecessary.

The PR adds:

- The single method `add_to_property_registry` is now expanded to `add_indexed_to_property_registry` and `add_to_property_registry_with_constructor`.
- `PropertyValueStoreCore` gets a new constructor `new_boxed_with_default_index`.
- `PropertyMetadata` gets a new field `index_new_entity_fn`



The `index_new_entity_fn` field is the most suspicious to me. When a new property is registered (before `main` executes), it stores a constructor in `value_store_constructor` that can construct a new property value store for the property in a type erased way. This constructor is used in the constructor of a `PropertyStore<E>` (the thing that holds every property value store for every property of an entity) so that the `PropertyStore<E>` doesn't need to know all the concrete types of all of its properties.

What I expected is that the function stored in `value_store_constructor` would already construct the property value storage in a way that respects the `Property::default_index_type()` value—or more precisely, I would expect `PropertyValueStoreCore::<E, P>::new_boxed` itself to respect the `Property::default_index_type()` value. The property registry doesn't need to be involved at all, as far as I can tell.

Am I missing something?

cdc-as81:
> This PR is a lot more invasive than I expected, and I can't tell if it is because I don't fully understand some edge cases or if the complexity is genuinely unnecessary.
> 
> The PR adds:
> 
> * The single method `add_to_property_registry` is now expanded to `add_indexed_to_property_registry` and `add_to_property_registry_with_constructor`.
> * `PropertyValueStoreCore` gets a new constructor `new_boxed_with_default_index`.
> * `PropertyMetadata` gets a new field `index_new_entity_fn`
> 
> The `index_new_entity_fn` field is the most suspicious to me. When a new property is registered (before `main` executes), it stores a constructor in `value_store_constructor` that can construct a new property value store for the property in a type erased way. This constructor is used in the constructor of a `PropertyStore<E>` (the thing that holds every property value store for every property of an entity) so that the `PropertyStore<E>` doesn't need to know all the concrete types of all of its properties.
> 
> What I expected is that the function stored in `value_store_constructor` would already construct the property value storage in a way that respects the `Property::default_index_type()` value—or more precisely, I would expect `PropertyValueStoreCore::<E, P>::new_boxed` itself to respect the `Property::default_index_type()` value. The property registry doesn't need to be involved at all, as far as I can tell.
> 
> Am I missing something?

Thanks—there is one constraint here that was not obvious from the previous
implementation.

I agree that default-index construction belongs with property-value-store
construction. I refactored the code accordingly:

- `PropertyMetadata` no longer has a separate `index_new_entity_fn` field. It
  now stores only the
  [constructor](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L110-L125).
- The constructor returns the property value store and its optional new-entity
  dispatcher as one value, so they cannot get out of sync
  ([source](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L52-L61)).
- `new_boxed_with_default_index` has been removed. The indexable constructor
  evaluates `P::default_index_type()` once and passes the selected index into
  `new_boxed_with_index`
  ([source](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L199-L212),
  [store constructor](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_value_store_core.rs#L41-L58)).

The reason `PropertyValueStoreCore::<E, P>::new_boxed()` cannot directly honor
`P::default_index_type()` is the trait bound introduced by constructing an
index.

`new_boxed()` must work for every property:

```rust
impl<E: Entity, P: Property<E>> PropertyValueStoreCore<E, P> {
    fn new_boxed() -> Box<dyn PropertyValueStore<E>> {
        // ...
    }
}
```

However, constructing either index requires:

```rust
P: IndexableProperty<E> // Property<E> + Eq + Hash
```

That requirement is visible in
[`PropertyIndexType::new_property_index`](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/index/mod.rs#L38-L49)
and the
[`IndexableProperty` definition](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property.rs#L176-L187).

Rust cannot acquire the `Eq + Hash` bound conditionally after inspecting the
runtime `PropertyIndexType`. Putting this directly in the general constructor
would therefore require either:

1. Restoring `Eq + Hash` as a requirement for every `Property`, undoing the
   change from [#937](https://github.com/CDCgov/ixa/pull/937), or
2. Adding another construction hook to the public `Property` trait and every
   property macro.

The two registration entry points are the narrow compile-time bridge between
those bounds:

- Ordinary `P: Property<E>` values register an unindexed constructor.
- Multi-properties, which satisfy `P: IndexableProperty<E>`, register the
  constructor that honors `default_index_type()`.

The relevant
[registration split](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L172-L213)
is implemented in `property_store.rs`.
The registry still contains only type-erased construction behavior; it does not
contain or share an index instance.

The dispatcher is required because constructing an empty index is not
sufficient to keep it current. After a new entity's explicit values have been
assigned, the dispatcher computes the derived or multi-property value and
inserts the entity into the index
([entity creation](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/context_extension.rs#L340-L372),
[dispatcher](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L63-L79)).

There is regression coverage showing that:

- Every constructed indexed store has its matching dispatcher
  ([test](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/entity/property_store.rs#L649-L665)).
- Different `Context` instances receive independent indexes
  ([test](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/macros/property_impl.rs#L1697-L1727)).
- Explicitly upgrading an unindexed property indexes existing entities and
  continues indexing subsequently added entities
  ([test](https://github.com/CDCgov/ixa/blob/e3c49eb449bf097f388580f01a39aa40c1a5754c/src/macros/property_impl.rs#L1880-L1955)).

So your expectation is correct in principle: the registered constructor should
produce a store that respects the configured default. The remaining
specialized registration path exists only to retain the narrower `Property`
bound established by #937 while constructing indexes for types known to
implement `IndexableProperty`.

PR #959, opened 2026-06-15
chore: add must use

### LINKED ISSUE #659: Add `#[must_use]` where it makes sense.
As a matter of code hygiene, we should really have `#[must_use]` in more places. A great example:

```rust
fn sample_people<R: RngId + 'static, Q: Query>(
    &self,
    rng_id: R,
    query: Q,
    n: usize,
) -> Vec<PersonId>
```

This method does have a subtle side effect of mutating the internal state of the random number source, but it's nonsensical for client code to call this method without using the result. If client code calls the method without using the result, it's almost certainly a mistake, potentially with significant runtime cost.

Other obvious cases: `get_population` / `get_entity_count`, `get_person_property`/`get_property`, ...

Adding a `#[must_use]` annotation
- Catches logic bugs at compile time
- Documents API intent (“this value matters”)
- Improves correctness without runtime cost

I wouldn't object if a PR also improves exist or adds missing doc comments to the affected functions.

cdc-as81:
## Summary

Adds `#[must_use]` annotations across value-returning APIs where ignoring the result is likely a bug or wasted work.

This covers the main entity/query/property/random sampling surfaces, collection helpers, registry/store/index accessors, profiling/execution helpers, network getters, and relevant `ixa-bench` support code.

## Details

- Added `#[must_use]` to user-facing APIs such as:
  - `ContextEntitiesExt::get_property`
  - `query_entity_count`
  - `sample_entity`
  - `count_and_sample_entity`
  - `sample_entities`
  - `get_entity_count`
  - `query` / `query_result_iterator`
- Added `#[must_use]` to random sampling helpers, including:
  - `ContextRandomExt::sample`
  - `sample_distr`
  - `sample_range`
  - `sample_bool`
  - `sample_weighted`
  - lower-level sampling algorithms
- Added annotations to pure/value-returning helpers in entity stores, property stores, indexes, collections, hashing, profiling, execution stats, and network internals.

### DIFF
--- a/integration-tests/ixa-runner-tests/tests/macros.rs
+++ b/integration-tests/ixa-runner-tests/tests/macros.rs
@@ -315,7 +315,7 @@ mod tests {
         let e = ctx.get_edge::<Person, TestEdge>(p1, p2).unwrap();
         assert_eq!(e.weight, 1.0);
         // Now remove the edge and ensure it's gone
-        ctx.remove_edge::<Person, TestEdge>(p1, p2);
+        assert!(ctx.remove_edge::<Person, TestEdge>(p1, p2).is_some());
         assert!(ctx.get_edge::<Person, TestEdge>(p1, p2).is_none());
 
         // Data plugin access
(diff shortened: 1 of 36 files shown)


RobertJacobsonCDC (approved):
Thanks!

PR #962, opened 2026-06-16
feat: Triggers

### LINKED ISSUE #944: Threshold triggers?
In userland, we have some code that triggers an event (shutdown, or perhaps a policy) at a certain threshold, that is, when a quantity reaches a certain level. 

This is accomplishable today with custom events, but there are some footguns/optimizations that we could potentially address with a dedicated API or recommended pattern. Some observations: 

- Threshold conditions tend to be either an incrementing count (e.g., deaths, cases), or a time;
- Sometimes a threshold is triggered by simultaneous events and you want to maintain idempotency / only trigger it once;
- If you run a query to access a count, there is sometimes a perf trade-off with maintaining the index 

Questions
- Are there other types of thresholds we have to consider (e.g. entity counts?) 
- Do we need to support other kinds of operators (e.g., OR) / different match conditions (<, <=)
- Should thresholds be re-triggerable, or always fire only once? 
- How are time thresholds the same/different from periodic plans or "passive" plans? 
- How much should we optimize these to be maximally fast vs use pre-existing primitives (i.e., the event system)? 

Sketches

```rs
context.add_property_threshold(Person, InfectionStatus::Infectious, 100, |context, count| {
    println!("{count} people are now infectious");
    context.schedule_shutdown();
});

context.add_time_threshold(50, |context| {
    context.schedule_shutdown();
});

// Or a fancy builder
context.when_once(
    count(Person, InfectionStatus::Infectious).at_least(100) 
    | count(Person, Alive(false)).at_least(100)
    | time().at_least(200.0),
    |ctx| ctx.schedule_shutdown()
);

### PR DESCRIPTION
**Pull request by RobertJacobsonCDC**
This branch adds a new `triggers` module for registering higher-level simulation conditions that emit normal user-defined Ixa events.

It introduces a small trigger API around `Context::register_trigger` and a catalog of built-in concrete trigger types with builder-pattern constructors and helper methods (`emit_with`, `emit_value`, `emit_default`) to bind a particular trigger criterion to the emission of a user-prescribed `IxaEvent`. 

```rust
#[derive(IxaEvent)]
struct FirstDeath {
    person: EntityId<Person>,
}

context.register_trigger(
    // The `to` constructor matches whenever a property changes to the given value.
    PropertyChangeTrigger::to(Alive(false))
  			// The `once` option specifies that we only match on the first observation
  			// and then permanently deactivate. Default is to emit on every match.
        .once()
  			// The `emit_with` method specifies how to construct an event instance from
  			// an observed criterion match.
        .emit_with(|event| FirstDeath {
            person: event.entity_id,
        }),
);

// Subscribe to the event.
context.subscribe_to_event(|_context, _event: FirstDeath| {
    // perform cleanup tasks
	context.shutdown();
});
```

Built-in trigger criterion:

- `TimeTrigger`: simulation reaches a configured time and execution phase.
- `EntityCountTrigger`: creation of an entity brings that entity type’s total count to a threshold.
- `PropertyChangeTrigger`: a property write changes a property value from/to a particular value.
- `PropertyValueCountTrigger`: the count of entities having a given property value reaches a threshold.

The special `TogglingTrigger`  composes activation and deactivation criteria (any of the above) into a simple stateful trigger. This supports thermostat-style hysteresis, such as turning an intervention on at one threshold and off at another, while emitting separate activation and deactivation events (potentially of different types).


RobertJacobsonCDC:
From my conversation with @erik-rosenstrom: 

- We are missing a periodic time trigger.
- We can imagine other variations on the property value count triggers, especially:
  - One that is tied to a property but not to a particular single value of that property (or to some set of values). The use case here might be, e.g., value counts _stratified_ by geography. 
  - Incidence-oriented value count ("hospitalizations in the last week") instead of prevalence-oriented value count (instantaneous total hospitalized)

Let's add the periodic time trigger to this PR, since it feels like an obviously useful one.

src/triggers/property_value_count.rs
```diff
@@ -0,0 +1,389 @@
...
+                });
+
+                context.subscribe_to_event({
+                    let count = Rc::clone(&count);
+                    let on_match = Rc::clone(&on_match);
+                    move |context, event: PropertyChangeEvent<E, P>| {
+                        let previous_count = count.get();
+                        let current_count =
+                            match (event.previous == self.value, event.current == self.value) {
+                                (false, true) => previous_count + 1,
```
cdc-as81, line 347: Property value count triggers can underflow if an entity is created with the tracked value and then changed away before callbacks drain. The queued creation handler reads the entity’s later property value and skips incrementing, then the queued property-change handler decrements from zero at `src/triggers/property_value_count.rs:348`, causing a debug panic and corrupting the count in release. Add regression coverage for `add_entity(with!(..., tracked)); set_property(..., untracked); execute();` and make creation accounting independent of deferred property reads.

k88hudson-cfa (approved):
Looks good to me - I think there are some doctests that are failing to compile? I think this might be a rebase issue or something

src/triggers/property_value_count.rs
RobertJacobsonCDC: I don't think this is possible. Entity creation queues the handler for the `EntityCreatedEvent`, and an immediate property change queues the handler for a `PropertyChangeEvent`, and these callbacks are processed in queued order by the event loop.

PR #963, opened 2026-06-16
chore: refactor tests

### LINKED ISSUE #952: Move tests in the `tests` folder to `integration-tests`
We have tests spread across three sources:

1. Unit tests live in submodules directly in the codebase
2. The `tests` directory
3. The `integration-tests` directory

I propose we eliminate the `tests` directory, moving the tests in that directory to a new workspace member under `integration-tests`.

# The difference between `tests` and `integration-tests`

Tests in the `tests` directory are treated as integration tests specifically for the `ixa` crate in particular. Cargo treats this directory as special. Each Rust file directly under this directory is compiled as its own test crate, rather than as code belonging to the `ixa` crate itself. They use `ixa`’s dev-dependencies, feature selection, target directory, and package context. But from Rust visibility/module perspective, they are external crates using `ixa` through its public API.

The `integration-tests` directory is not treated in a special way by Cargo. It is repo-specific convention. It contains distinct crates that we have defined as part of the workspace. Each crate under `integration-tests/` has its own `Cargo.toml`, so it can define its own dependencies, features, binaries, crate type, build/test tooling, and package-level configuration while still participating in the workspace.

# Proposed new structure

```
integration-tests/
  ixa-runner-tests/
  ixa-wasm-tests/
  ixa-repo-tests/
```

We might think about whether `ixa-runner-tests` itself should be reorganized. It has grown beyond just testing the `runner` functionality.

cdc-as81:
## Summary

- Split integration tests into focused crates for runner, macros, profiling, and repo-level checks
- Move shared global-property fixtures under `integration-tests/fixtures`
- Keep profiling coverage in a dedicated package with the `profiling` feature enabled
- Update runner tests and fixture paths after the package rename


RobertJacobsonCDC (approved):
This looks like a reasonable organization to me.

PR #967, opened 2026-06-18
chore: re-export IxaEvent derive from ixa

### LINKED ISSUE #964: Re-export `ixa_derive::IxaEvent` from `ixa`
Currently, normal client code needs both imports:

```rust
use ixa::IxaEvent;        // the trait
use ixa_derive::IxaEvent; // the derive macro

#[derive(IxaEvent)]
struct MyEvent;
```

The trait is exported from `ixa`, but the derive macro is defined in `ixa-derive`. The macro expansion emits an unqualified `impl IxaEvent for ...`, so the trait must also be in scope at the derive site. Consequently:

- `use ixa::IxaEvent;` alone is not enough, because `ixa` does not re-export the derive macro at crate root.
- `use ixa_derive::IxaEvent;` alone is not enough, because the generated impl needs the `ixa::IxaEvent` trait in scope.

A more ergonomic API would be to re-export the derive macro from `ixa` itself, serde-style, so users could write only `use ixa::IxaEvent;`. (The two identical import names are allowed because Rust keeps derive macros and traits in separate namespaces.) It turns out that `ixa-derive` is already a dependency of `ixa`, so we are not saving client code from depending on `ixa-derive` if they don't need it—they are paying for it anyway.

cdc-as81:
## Summary

- Re-export `ixa_derive::IxaEvent` from the `ixa` crate root
- Document the new ergonomic usage: `use ixa::IxaEvent; #[derive(IxaEvent)]`
- Update internal derives to exercise the crate-root re-export
- Remove direct `ixa-derive` dependencies from example crates that no longer need them

### DIFF
--- a/examples/basic-infection/Cargo.toml
+++ b/examples/basic-infection/Cargo.toml
@@ -11,7 +11,6 @@ publish = false
 
 [dependencies]
 ixa = { path = "../../" }
-ixa-derive = { path = "../../ixa-derive" }
 
 serde.workspace = true
 rand_distr.workspace = true
--- a/examples/births-deaths/Cargo.toml
+++ b/examples/births-deaths/Cargo.toml
@@ -11,7 +11,6 @@ publish = false
 
 [dependencies]
 ixa = { path = "../../" }
-ixa-derive = { path = "../../ixa-derive" }
 
 serde.workspace = true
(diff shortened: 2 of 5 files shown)


RobertJacobsonCDC (approved):
LGTM

PR #970, opened 2026-06-22
chore: remove getrandom

cdc-as81:
removed un used getrandom.
also updated cargo-bins/cargo-binstall to latest.


RobertJacobsonCDC (approved):
Did you mean to bump the version of `cargo-binstall` in this PR? It's probably fine, but....

PR #973, opened 2026-06-22
chore: Audit obsolete Clippy allow attributes

### LINKED ISSUE #936: Audit `allow(clippy)`
Given that we switched from pedantic to not pedantic at some point, it might be worth auditing the code base for `#[allow(clippy`... that aren't actually needed anymore. 

You can use `mise run check:rust` or `cargo +nightly clippy --workspace --all-targets --all-features -- -D warnings` after removing the allow line to make sure it's not needed

cdc-as81:
## Summary

- Removed obsolete `#[allow(clippy::...)]` attributes that are no longer needed after disabling Clippy pedantic lints.
- Kept only the local Clippy allowances still required by the current workspace lint configuration.
- Narrowed one `approx_constant` allowance from module scope to the specific test that needs it.

### DIFF
--- a/examples/basic-infection/src/transmission_manager.rs
+++ b/examples/basic-infection/src/transmission_manager.rs
@@ -26,7 +26,6 @@ fn attempt_infection(context: &mut Context) {
     // An alternative implementation calculates each person's time to infection
     // at the beginning of the simulation and schedules their infection at that time.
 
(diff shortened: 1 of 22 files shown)


RobertJacobsonCDC (approved):
LGTM

PR #975, opened 2026-06-23
chore: Add unsubscribe support for event listeners

### LINKED ISSUE #972: RemoveListener Once or both for handlers
Let's start by adding a `removeListener` and see how it goes

cdc-as81:
## Summary

Adds token-based event listener unsubscription to the event system.

`subscribe_to_event` now returns an `EventListenerId<E>`, which can be passed to `unsubscribe_from_event` to stop receiving future emissions for that event type.

## Changes

- Added `EventListenerId<E>` as an opaque typed listener token.
- Updated `subscribe_to_event` to return the listener token.
- Added `unsubscribe_from_event` on `Context` and `ContextBase`.
- Stores event handlers with listener ids so individual listeners can be unsubscribed.
- Added per-context listener identity so tokens from one `Context` cannot unsubscribe listeners from another.
- Updated plugin context coverage for the new API.
- Updated `has_event_handlers` to reflect active listeners after unsubscribe.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -8,7 +8,6 @@ use std::collections::VecDeque;
 use std::fmt::{Display, Formatter};
 use std::marker::PhantomData;
 use std::rc::Rc;
-use std::sync::atomic::{AtomicUsize, Ordering};
 
 use crate::data_plugin::DataPlugin;
 use crate::entity::entity_store::EntityStore;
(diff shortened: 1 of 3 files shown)


src/context.rs
```diff
@@ -141,18 +181,68 @@ impl Context {
...
+        let handler_vec: &mut Vec<EventHandlerRegistration<E>> =
+            handler_vec.downcast_mut().unwrap();
+        let Some(index) = handler_vec.iter().position(|entry| {
+            entry.listener_id.context_id == listener_id.context_id
+                && entry.listener_id.id == listener_id.id
+        }) else {
+            return false;
+        };
+
+        handler_vec.remove(index);
```
RobertJacobsonCDC: Use `swap_remove` here.

src/context.rs
```diff
@@ -141,18 +181,68 @@ impl Context {
...
+        let empty = handler_vec.is_empty();
+        if empty {
+            self.event_handlers.remove(&event_type);
```
RobertJacobsonCDC: If a single handler is repeatedly added and removed, then we are repeatedly allocating and dropping a vector. It makes more sense to me to keep the allocated vector around, because we probably have only a handful of event types anyway, which means we are only retaining a handful of vectors.

RobertJacobsonCDC (changes requested):
There's no need for a context ID or to tie the `EventListenerId<E>` to the context ID. We don't do that for `EntityId` or `PlanId`. That removes the context ID bookkeeping and simplifies `EventListenerId<E>`.

src/context.rs
cdc-as81: but that would replace the removed id with the last id, not preserving the order. I thought we wanted to keep the order?

cdc-as81: our internal docs:
`    /// Handlers will be called upon event emission in order of subscription as
    /// queued `Callback`s with the appropriate event.
    pub fn subscribe_to_event<E: IxaEvent>(&mut self, handler: impl Fn(&mut Context, E) + 'static) {`
suggest we want to maintain order

RobertJacobsonCDC (approved):
Looks good to me!

PR #976, opened 2026-06-24
feat: Shutdown semantics and shutdown-time plans

### LINKED ISSUE #926: Create pre-shutdown event and refine semantics
Sometimes things like reports need to be finalized before a shutdown event, so probably it would be a good idea to emit some kind of event before shutdown happens

RobertJacobsonCDC:
This PR introduces a distinguished "shutdown time" at which plans can be scheduled (with optional `ExecutionPhase`) through two new methods, `Context::add_shutdown_plan` and `Context::add_shutdown_plan_with_phase`. The plans scheduled at this distinguished time execute during a "normal" shutdown cycle (described below). These methods return a `PlanId`, and plans scheduled through them can be cancelled using the usual `Context::cancel_plan` method, just like any other plan.

In addition, we introduce two distinct shutdown modes:

1. `Context::shutdown()`: Triggers the "normal" shutdown procedure. The event loop continues executing all jobs and plans that are scheduled at `context.current_time` (all phases). Once all jobs and all plans scheduled for `context.current_time` are exhausted, the shutdown-time plans (and any newly queued callbacks) are executed. The current simulation time is never advanced.
2. `Context::abort()`: Immediately exits the event loop without executing any callbacks, plans, or the new shutdown-time plans. (This preserves the semantics of the previous implementation of `Context::shutdown()`.)

This PR includes a refactor of the plan queue behind the scenes, removing vestigial generic parameters, renaming it to `PlanQueue`, and introducing functionality to support the distinguished shutdown-time plans. 

Distinguishing between active and passive plans is _out of scope_ for this PR.

---

## Rationale for new shutdown semantics

- We retain the "abort" ability we previously had as the semantics for `context.shutdown()` via `Context::abort()`.
- The "normal" shutdown semantics preserve the notion that models shouldn't depend on the arbitrary execution order of plans/jobs scheduled "simultaneously."
- "Extra" work is likely to be typically nothing or very small anyway for `Normal` shutdown request.
- Given that we have shutdown-time plan scheduling:
  - We should probably at least exhaust the job queue, because the plans scheduled to run at shutdown time might themselves generate jobs they expect to run (changing property values, emitting other events).
  - Exhausting all plans for the current time in all phases before executing the "shutdown time" plans makes scheduled plan execution easier to reason about and avoids awkward interaction between execution phases of regular plans scheduled at current time and execution phases of plans scheduled at "shutdown time." (It would be awkward if a plan scheduled at current time with phase `ExecutionPhase::First` didn't run but a plan scheduled at shutdown time with phase `ExecutionPhase::Last` did run.)

## Rationale for distinguished shutdown-time plan scheduling

Notice that "shutdown time" plan scheduling can be thought of as strictly more powerful than a `ShutdownPending` event: anything that would subscribe to the `ShutdownPending` event could instead be "scheduled" at the shutdown time, whereas shutdown time additionally has `ExecutionPhase`s. If there is a weird use case in which events and event subscribers are more convenient for shutdown-time execution, the shutdown semantics described above allow client code to implement this themselves by scheduling a plan at shutdown time that emits their own `ShutdownPending` `IxaEvent`.


src/lib.rs
```diff
@@ -69,7 +69,14 @@ pub use network::{ContextNetworkExt, Edge, EdgeType};
 
 pub mod macros;
 
-pub mod plan;
+pub mod plan_queue;
+/// Compatibility re-exports for plan-related public API.
+///
+/// The plan queue implementation lives in [`crate::plan_queue`]. This module is
+/// retained so existing `ixa::plan::PlanId` imports continue to work.
+pub mod plan {
```
cdc-as81, line 77: This looks like an unintended public API break: `ixa::plan` used to expose `Queue`, `PlanSchedule`, `Plan`, and `Queue::list_schedules`, but the new compatibility shim only re-exports `PlanId`. If that removal is intentional, it should be treated/documented as a breaking change; otherwise, can we preserve the old exports or add a compatibility layer?

RobertJacobsonCDC: None of those other types were ever intended to be public, only `PlanId`. I consider it a bug fix rather than a breaking change.

(We avoid breaking in the case of `PlanId` by using the shim.)

src/plan_queue.rs
```diff
@@ -0,0 +1,506 @@
...
+                return Some(Plan {
+                    time: entry.time,
+                    data,
+                });
+            }
+        }
+    }
+
+    #[doc(hidden)]
+    pub(crate) fn remaining_plan_count(&self) -> usize {
```
k88hudson-cfa, line 205: It's maybe worth commenting this intentionally omits the shutdown queue

k88hudson-cfa (approved):
This looks good to me one comment inline

src/plan_queue.rs
RobertJacobsonCDC: This method will be removed in the passive plans PR anyway.

PR #978, opened 2026-06-25
chore: Add RNG Debug State

### LINKED ISSUE #620: Make internal RNG state visible for debugging purposes
Add a method to get (a hash of) the internal state of an RNG that can be used to compare across runs for debugging purposes.

# Implementation

It's not clear what is the best way to implement this feature.

If `RngType` implements `Serialize`:

```rust
pub fn debug_rng_state<R: RngId + 'static>(&self, _rng_type: R) -> u64 
where
    R::RngType: serde::Serialize,
{
    let rng = get_rng::<R>(self); // &mut R::RngType

    // 1. Serialize internal RNG state to a stable byte representation
    let mut state_bytes = Vec::new();
    bincode::serialize_into(&mut state_bytes, rng)
        .expect("RNG must be serializable for debug_rng_state");

    // 2. Hash the serialized bytes
    let mut hasher = DefaultHasher::new();
    state_bytes.hash(&mut hasher);
    hasher.finish()
}
```

If `RngType` implements `Clone`:

```rust
pub fn debug_rng_state<R: RngId + 'static>(&self, _rng_type: R) -> u64 {
    let rng = get_rng::<R>(self);

    // Copy internal state non-destructively; requires Clone on RNG type
    let sample1 = clone.next_u64();
    let sample2 = clone.next_u64();
    let sample3 = clone.next_u64();
    let sample4 = clone.next_u64();

    let mut hasher = std::collections::hash_map::DefaultHasher::new();
    sample1.hash(&mut hasher);
    sample2.hash(&mut hasher);
    sample3.hash(&mut hasher);
    sample4.hash(&mut hasher);
    hasher.finish()
}
```

# Background / Alternatives

We discussed the challenge of debugging random number related code logic and threw out some ideas:

- We could check that the internal state of the RNG is consistent (across runs) at points which we believe are consistent w.r.t. our program logic, e.g. by providing a partial hash of the internal state.
- We could trigger some kind of hard reset/reseed of RNGs so that at the trigger point we can guarantee consistent alignment of random state.
- There's an alternative model in which instead of the traditional RNG, the random number is generated as a hash of the relevant state involved in the request for the random number, "hyper reproducible" random numbers.

cdc-as81:
## Summary

Adds `ContextRandomExt::debug_rng_state(...) -> u64`, a deterministic debug fingerprint for the current position of a context RNG.

The method clones the selected RNG, draws from the clone, and hashes those draws with the crate’s deterministic hasher, so callers can compare RNG checkpoints across runs without advancing the live RNG.

## Changes

- Added `debug_rng_state` to `ContextRandomExt`.
- Kept the new `Clone + RngCore` bounds local to this method.
- Used `DeterministicHasher` for stable fingerprinting within the same code/dependency set.
- Documented that the value is for debugging comparison only, not serialization or cross-version reproducibility.
- Added tests for:
  - Same seed and same progress matching.
  - Fingerprint changing after RNG progress.
  - Repeated calls not changing the fingerprint.
  - Debug calls not affecting the next sample.
  - Resetting with the same seed restoring the fingerprint.
  - Independent fingerprints across RNG IDs.

### DIFF
--- a/src/random/context_ext.rs
+++ b/src/random/context_ext.rs
@@ -1,13 +1,14 @@
 use std::any::TypeId;
 use std::cell::RefMut;
+use std::hash::Hasher;
 
 use log::trace;
 
-use crate::hashing::hash_str;
+use crate::hashing::{hash_str, DeterministicHasher};
 use crate::rand::distr::uniform::{SampleRange, SampleUniform};
 use crate::rand::distr::weighted::{Weight, WeightedIndex};
 use crate::rand::distr::Distribution;
-use crate::rand::{Rng, SeedableRng};
+use crate::rand::{Rng, RngCore, SeedableRng};
 use crate::random::{RngHolder, RngPlugin};
 use crate::{Context, ContextBase, RngId};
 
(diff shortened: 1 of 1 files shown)


RobertJacobsonCDC (approved):
This looks pretty reasonable to me. I like that it avoids requiring the RNG implement `Hash` as well.

PR #980, opened 2026-06-25
docs: Random network example

swo:
Minimal example of a randomly-generated network. Shows how to use an external graph-generating package to build connection networks.

Some open questions I think this PR poses:

1. Does this example belong here? How many examples do we want in this repo? One solution I could see, is having ~3 very well-curated examples here in this repo, and then another, more free-for-all ixa-examples repo.
2. Do we want to allow users to access rngs defined by `define_rng!`?. Right now, [`get_rng()`](https://github.com/CDCgov/ixa/blob/e7bddf055f422469b9087a8547e5cda8c0144b3f/src/random/context_ext.rs#L17) is private, so users can't directly access the defined rngs or their states. In this example, I need to call an external-to-ixa function to generate the graph, and I might have wanted to use the rng state to seed the generation of that graph, rather than passing around a seed as if it were a separate parameter.


k88hudson-cfa:
Let's close this for now in favour or some of the new work in comparison models etc.

PR #981, opened 2026-06-25
feat: Passive plans

### LINKED ISSUE #916: Distinguish periodic plans from other plans in the scheduler
Normally simulation ends when the event queue is empty. However, periodic plans reschedule themselves indefinitely. Consider for example the value change counter. The periodic handler currently does this after each report:

```rust
if context.remaining_plan_count() == 0 {
    return;
}

context.add_plan_with_phase(next_time, ...);
```

That works for a single counter if the only remaining plans are “real” simulation plans. But with multiple periodic counters, each counter’s already-scheduled report is counted by `remaining_plan_count()`.

So at time `t`:

1. Counter A reports.
2. remaining_plan_count() sees Counter B’s report still queued.
3. Counter A schedules its next report at t + period.
4. Counter B reports.
5. remaining_plan_count() now sees Counter A’s next report queued.
6. Counter B schedules its next report.

The periodic report plans keep each other alive forever.

I see two solutions:

1. Document this behavior as expected. Client code will need to explicitly add a shutdown plan.
2. Distinguish between periodic plans and nonperiodic plans in the scheduler. Then either:
  - change periodic plans to only reschedule themselves if there are scheduled nonperiodic plans, or
  - end the simulation when there are no more nonperiodic plans scheduled.

RobertJacobsonCDC:
The event loop includes a built-in stop condition in which "normal shutdown" is triggered if the regular plan queue is exhausted. This PR introduces a distinction between "active" and "passive" plans and modifies this stop condition so that normal shutdown is triggered if there are no more _active_ plans in the regular plan queue.

This is motivated by the fact that self-rescheduling periodic plans will always keep the event loop alive indefinitely, making the built-in "exhausted plan queue" stop condition unusable for all but the simplest toy models.

**Out of scope:**
- We should add an Ixa Book topic chapter on "Execution and Shutdown" in a separate PR.
- Shutdown semantics, in particular what happens when "normal shutdown" is triggered, belongs to #976 upon which this PR is stacked.
- We don't address primitives / convenience methods for constructing self-rescheduling plans directly in this PR except to modify the two existing mechanisms so they create passive plans. Whether we want better helpers or what they would be is out of scope for this PR.

## Public API

- Existing `add_plan` and `add_plan_with_phase` APIs remain _active_—they keep the simulation timeline alive. 
- New `add_passive_plan` and `add_passive_plan_with_phase` APIs schedule normal timed _passive_ plans that participate in the same time, phase, and FIFO ordering, but do not prevent shutdown once _active_ work is exhausted.
- `add_periodic_plan_with_phase` is now _passive plan only_, as it's plans are by definition perpetually self-rescheduling.
- Value-change count handlers registered with `track_periodic_value_change_counts` are implicitly self-rescheduling and are thus also now passive.

---

This PR is stacked on top of #976.

### DIFF
--- a/docs/book/src/topics/reports.md
+++ b/docs/book/src/topics/reports.md
@@ -409,6 +409,11 @@ pub fn init(context: &mut Context) {
 }
 ```
 
+Periodic plans are passive. They reschedule themselves after each report, but
+they do not keep the simulation running after active plans are exhausted. This
+means a periodic report can run at the final active simulation time, while later
+periodic report callbacks remain queued unless more active work is scheduled.
+
 The implementation of `write_aggregate_sir_report_item` is straightforward: We
 fetch the values from the data plugin, construct an instance of
 `AggregateSIRReportItem`, and "send" it to the report.
--- a/docs/book/src/topics/reports.md
+++ b/docs/book/src/topics/reports.md
@@ -410,9 +410,10 @@ pub fn init(context: &mut Context) {
 ```
 
 Periodic plans are passive. They reschedule themselves after each report, but
-they do not keep the simulation running after active plans are exhausted. This
-means a periodic report can run at the final active simulation time, while later
-periodic report callbacks remain queued unless more active work is scheduled.
+they do not keep the simulation running after non-passive plans are exhausted.
+This means a periodic report can run at the final simulation time,
+while later periodic report callbacks remain queued unless more work
+is scheduled.
 
 The implementation of `write_aggregate_sir_report_item` is straightforward: We
 fetch the values from the data plugin, construct an instance of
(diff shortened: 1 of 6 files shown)


k88hudson-cfa (approved):
In our discussion we decided it would be better to have the notion of `normal` vs passive plans instead of active being a special state 

I think it's a good call to make periodic plans passive without intervention from the user; 

I would probably recommend we add an API for shutdown to ensure that it is scheduled via passive plan and mostly passive plans are are internal mechanism, but it's ok to expose these methods in case they are needed (Filed #984)

k88hudson-cfa (changes requested):
Changing  `add_plan_with_phase_and_activity` is probably the only thing that needs to change externally to be `is_passive` instead

k88hudson-cfa (approved):
Thanks!

Issue #984, opened 2026-06-26
Add schedule shutdown helper method


k88hudson-cfa:
This would just schedule a shutdown in a **passive** plan to shutdown at some time t, and would help avoid footguns 

`context.schedule_shutdown(MAX_TIME)`

This depends on passive plans landing

PR #986, opened 2026-06-26
docs: Simulation execution and shutdown chapter of The Ixa Book

RobertJacobsonCDC:
A new chapter of the Ixa Book that covers everything related to simulation execution and showdown. [GitHub's partial rendering of it is here](https://github.com/CDCgov/ixa/blob/28415e6554d14dabd3ee98ebed7276c6b434fd54/docs/book/src/topics/execution-and-shutdown.md).

1. Execution and Shutdown
   1. Syntax Summary
   2. The Execution Lifecycle
      1. Beginning Execution
      2. During Normal Event-loop Execution
      3. Ending Execution
         1. Automatic Shutdown
         2. Normal Shutdown
            1. What Runs During Normal Shutdown
            2. Current-Time Plans Across Phases
            3. Shutdown-Time Plans
            4. What Remains Queued After Shutdown
         3. Aborting Execution
         4. Common Shutdown Patterns
   3. Execution Outside of the Event-loop
      1. Executing Again After `execute()` Returns
      2. Single-Step Execution

To Do Before Merging:

- [ ] Update to acknowledge the new `triggers` module (about to be merged, #962).
- [ ] Update to reflect decision to exclude "active plan" terminology from user-facing API and docs. We only have (regular) plans and "passive plans."
- [ ] Update to include "schedule shutdown" helper of #984 if it lands first.


RobertJacobsonCDC:
Clarify that there are _three_ work queues:

1. The callback queue
2. The plan queue, sometimes called the timeline
3. The shutdown-time plan queue

RobertJacobsonCDC:
We changed the language of the "passive plans" implementation to avoid calling non-passive plans "active." We just have plans, and then we have a special kind of "passive plan." This is just a language change, not a change in underlying functionality.

This new chapter needs to be updated to reflect this new language. Looks like it also needs to be rebased.

RobertJacobsonCDC:
~~Wait for #999.~~ Edit: We decided not to implement #999.

PR #987, opened 2026-06-26
docs: fix sample typo

bbbruce:
Need `&mut rng` not just `&rng`


bbbruce:
@RobertJacobsonCDC - now I think I may be wrong - think it had to do with the way I had my code setup at the time - so please close if so

RobertJacobsonCDC (approved):

RobertJacobsonCDC:
I prematurely clicked "approve," but actually this change is incorrect. A working example is in the unit test `sampler_function_closure_capture` in the file `src/random/context_ext.rs` around like 198. So in this case you don't need the `& mut`.

PR #995, opened 2026-07-02
chore: Make plugin context traits dyn-compatible

### LINKED ISSUE #919: Investigate why `PluginContext` isn't dyn compatible.
This issue is to audit the various `Context` trait extensions for [dyn compatibility](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility).

Apparently PluginContext isn't dyn compatible (can't be used as a trait object). We of course only have one type that implements these trait extensions, namely `Context`. It's annoying to have to make the type that implements `PluginContext` a generic type in client code. It'd be much nicer to just deal with a trait object. Same for it's supertraits:

```rust
PluginContext:
    ContextBase
    + ContextRandomExt
    + ContextReportExt
    + ContextNetworkExt
    + ContextGlobalPropertiesExt
    + ContextEntitiesExt
```

If it helps, we can always assume `Sized` or anything else we know `Context` also implements.

cdc-as81:
## Summary

- Make `PluginContext` dyn-compatible by moving `Self: Sized` bounds onto non-object-safe typed helper methods.
- Add tests proving `dyn PluginContext` and dyn plugin extension traits compile for object-safe behavior.
- Update related extension traits and parameter-loading helpers to preserve concrete/generic typed APIs.

### DIFF
--- a/examples/parameter-loading/parameters_loader.rs
+++ b/examples/parameter-loading/parameters_loader.rs
@@ -17,12 +17,18 @@ pub struct Parameters {
 define_global_property!(ParametersKey, Parameters);
 
 pub trait ParametersExt: PluginContext {
-    fn init_parameters(&mut self, file_path: &Path) -> Result<(), IxaError> {
+    fn init_parameters(&mut self, file_path: &Path) -> Result<(), IxaError>
+    where
+        Self: Sized,
+    {
         let parameters_json = self.load_parameters_from_json::<Parameters>(file_path)?;
         self.set_global_property_value(ParametersKey, parameters_json)?;
         Ok(())
     }
-    fn get_parameters(&self) -> &Parameters {
+    fn get_parameters(&self) -> &Parameters
+    where
+        Self: Sized,
+    {
         self.get_global_property_value(ParametersKey)
             .as_ref()
             .unwrap()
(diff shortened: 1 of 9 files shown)


RobertJacobsonCDC (commented):
Wow, this is really eye-opening. What this is demonstrating is that for the `PluginContext` hierarchy, the only methods still callable on a dyn trait object are just three object-safe methods left on `ContextBase`:

- `cancel_plan(&mut self, plan_id: &PlanId)`
- `get_current_time(&self) -> f64`
- `get_execution_statistics(&mut self) -> ExecutionStatistics`

So practically, `dyn PluginContext` is usable only for cancellation, current time, and execution statistics. Which makes the dyn trait object almost completely useless. Yikes!

The underlying reason boils down to our extensive use of generic parameters on methods. Any method that has to deal with an `Entity` type, a `Property` type, an `IxaEvent` type, or a callable type like `impl FnOnce(&mut Context) + 'static` is generic over those types, and that's pretty much everything.

The goal of this issue was to enable client code to use `PluginContext` (or its individual component trait extensions) as a dyn trait object, but since we learned that such a dyn trait object just isn't useful for _any_ of the trait extensions beyond those three methods inherited from `ContextBase`, I vote that we close this PR without merging and close the associated issue with, "Won't fix."

k88hudson-cfa:
I agree closing is the right call here – we also might want to rethink/reconsider `PluginContext` altogether if it's causing more complexity than it's saving

PR #998, opened 2026-07-09
feat: Deprecate the plan module

### LINKED ISSUE #977: Deprecate the `plan` module
The public `plan` module implements the timeline, a priority queue of scheduled plans. The only public item in this module is `PlanId`. Instead of having an entire `pub` module for a single item, let's re-export `PlanId` either from crate root or from `prelude`. Then make the `plan` module private in the next major release.

In #976 I actually rename `plan.rs` to `plan_queue.rs` to better represent its present-day purpose, but I created a dummy `plan` module through which I re-export `PlanId` to be nonbreaking. If/when this lands, we should remove the compatibility kludge in v3.0.

cdc-as81:
## Summary

  - Re-export `PlanId` from the `ixa` crate root.
  - Deprecate `ixa::plan::PlanId` in favor of `ixa::PlanId`.
  - Update the births/deaths example, README, rustdoc, and changelog.
  - Retain legacy modules for 2.x; their removal remains a planned 3.0 breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Deprecated
+
+- `ixa::plan::PlanId`; use `ixa::PlanId` instead ([#977](https://github.com/CDCgov/ixa/issues/977))
+
 ## [2.1.0](https://github.com/CDCgov/ixa/compare/ixa-v2.0.0...ixa-v2.1.0) - 2026-06-29
 
 ### Added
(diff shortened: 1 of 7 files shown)


RobertJacobsonCDC:
This seems reasonable to me. The only question I have is, do we want it in the prelude as well? Right now `EntityId` is in the prelude. We also introduced `EventListenerId<E>` recently in order to support cancelation of event subscribers. It seems to me both `PlanId` and `EventListenerId` should be in the prelude if `EntityId` is.

Anyone have opinions on this?

RobertJacobsonCDC (approved):
LGTM

PR #1004, opened 2026-07-13
chore: add startup-time plans

### LINKED ISSUE #999: Add `add_startup_plan` to be functionally symmetric to `add_shutdown_plan`
# Add `add_startup_plan` to be functionally symmetric to `add_shutdown_plan`

Now that we allow [negative simulation time](https://ixa.rs/book/topics/burn-in-and-negative-time.html), the "first" `current_time` is not guaranteed to be time `0.0`. Client code might want to schedule plans to run "at startup," whatever that time might be, in initialization code that is potentially far away from whereever the startup time is determined.

We already have `add_shutdown_plan` / `add_shutdown_plan_with_phase` that allows you to schedule a plan for shutdown time, whenever that time might be. From the user perspective, `add_startup_plan` would be functionally symmetric: it schedules a plan to run at simulation start, whenever that time might be.

I think the best design is to:

-  modify `PlanQueue` to have a field `PlanQueue::startup_queue: BinaryHeap<PlanSchedule>` to be symmetric to the existing `shutdown_queue` field, and add corresponding `add_startup_plan` and `pop_next_startup` methods. 
- expand the event loop state machine currently implemented in `Context::shutdown_status: ShutdownStatus` to also capture startup time. Maybe rename this field and its type to `Context::event_loop_status: EventLoopStatus`, and change the names of its variants while adding a new variant `EventLoopStatus::StartUpTimePlans` (and possibly `EventLoopStatus::NotStarted`). 
- modify the `match` in `Context::execute_single_step` to burn down the startup queue in the `PlanQueue` during the `EventLoopStatus::StartUpTimePlans` event loop state, transitioning to the normal event loop execution state when the startup-time plan queue is exhausted.

If #986 lands before this, making appropriate modifications and additions to the "Execution and Shutdown" chapter of The Ixa Book should also be in scope.

cdc-as81:
## Summary

  This PR adds startup-time plans: one-shot lifecycle work that runs at the
  effective simulation start time—either the configured start time or `0.0`—
  before regular plans.

  - Add `Context::add_startup_plan` and
    `Context::add_startup_plan_with_phase`.
  - Expose startup-plan scheduling through `ContextBase` and `PluginContext`.
  - Expand the event-loop state machine to model startup, regular execution,
    shutdown, and abort/resume transitions explicitly.
  - Add a dedicated startup queue with shared plan IDs, cancellation, and
    profiling support.
  - Update the `Context` API documentation to describe the full simulation
    lifecycle.

## Behavior

  Startup-time plans:

  - Run once at the effective simulation start time before any regular plan.
  - Follow `ExecutionPhase` ordering, then insertion order within each phase.
  - Yield to queued callbacks, which retain priority over plan selection.
  - May schedule additional startup-time plans while startup is still active.
  - Can be canceled through their `PlanId`.
  - Resume on a later `execute` call when startup is interrupted by `abort`.
  - Finish draining before shutdown proceeds when shutdown is requested during
    startup.
  - Panic if added after startup has completed.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -117,8 +117,8 @@ enum ShutdownStatus {
 ///
 /// Provides core simulation services including
 /// * Maintaining a notion of time
-/// * Scheduling events to occur at some point in the future and executing them
-///   at that time
+/// * Scheduling regular work on the simulation timeline
+/// * Scheduling work for the startup and shutdown lifecycle stages
 /// * Holding data that can be accessed by simulation modules
 ///
 /// Simulations are constructed out of a series of interacting modules that
@@ -126,14 +126,16 @@ enum ShutdownStatus {
 /// store data in the simulation using the [`DataPlugin`] trait that allows them
 /// to retrieve data by type.
 ///
-/// The future event list of the simulation is a queue of `Callback` objects -
-/// called `plans` - that will assume control of the [`Context`] at a future point
-/// in time and execute the logic in the associated `FnOnce(&mut Context)`
-/// closure. Modules can add plans to this queue through the [`Context`].
+/// The future event list of the simulation is a queue of regular plans that
+/// assume control of the [`Context`] at a specified simulation time and execute
+/// the logic in an associated `FnOnce(&mut Context)` closure. Modules can also
+/// schedule startup-time plans, which run once at the effective simulation start
+/// time before regular plans, and shutdown-time plans, which run after regular
+/// work for an execution pass is exhausted.
 ///
-/// The simulation also has a separate callback mechanism. Callbacks
-/// fire before the next timed event (even if it is scheduled for the
-/// current time). This allows modules to schedule actions for immediate
+/// The simulation also has a separate callback mechanism. Callbacks fire before
+/// the next plan selected for execution, including startup-time, regular, and
+/// shutdown-time plans. This allows modules to schedule actions for immediate
 /// execution but outside of the current iteration of the event loop.
 ///
 /// Modules can also emit 'events' that other modules can subscribe to handle by
@@ -151,6 +153,7 @@ pub struct Context {
     current_time: Option<f64>,
     start_time: Option<f64>,
     shutdown_status: ShutdownStatus,
+    startup_complete: bool,
     execution_profiler: ExecutionProfilingCollector,
     pub(crate) print_execution_statistics: bool,
 }
@@ -180,6 +183,7 @@ impl Context {
             current_time: None,
             start_time: None,
             shutdown_status: ShutdownStatus::None,
+            startup_complete: false,
             execution_profiler: ExecutionProfilingCollector::new(),
             print_execution_statistics: false,
         }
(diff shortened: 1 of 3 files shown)


src/plan_queue.rs
```diff
@@ -100,12 +123,25 @@ impl PlanQueue {
         phase: ExecutionPhase,
     ) -> PlanId {
         trace!("adding shutdown-time plan");
+        self.add_distinguished_plan(callback, phase, DistinguishedQueue::Shutdown)
+    }
+
+    fn add_distinguished_plan(
```
RobertJacobsonCDC: No, don't do this. This is strictly worse than inlining the logic of this method.

src/plan_queue.rs
```diff
@@ -225,30 +262,51 @@ impl PlanQueue {
...
     ///
     /// Returns the next shutdown-time plan if it exists or else `None` if the
     /// shutdown-time queue is empty.
     pub(crate) fn pop_next_shutdown(&mut self) -> Option<Plan> {
         trace!("getting next shutdown-time plan");
-        std::iter::from_fn(|| self.shutdown_queue.pop()).find_map(|entry| {
+        self.pop_next_distinguished(DistinguishedQueue::Shutdown)
+    }
+
+    fn pop_next_distinguished(&mut self, queue: DistinguishedQueue) -> Option<Plan> {
```
RobertJacobsonCDC: Also this. This is strictly worse than inlining this logic.

src/context.rs
```diff
@@ -175,6 +178,7 @@ impl Context {
             current_time: None,
             start_time: None,
             shutdown_status: ShutdownStatus::None,
+            startup_complete: false,
```
RobertJacobsonCDC: Can you explain why you added a distinct field instead of expanding the event loop states as described in the issue?

RobertJacobsonCDC (changes requested):
The issue described how to expand the event loop state machine in a way that expands it to startup time. This implementation uses a separate state tracking mechanism that obfuscates which states and transitions are possible. If there is a reason, I need an explanation.

Also, don't let the LLM factor out single lines of code into useless helpers. Introducing `DistinguishedQueue`, `add_distinguished_plan`, and `pop_next_distinguished` is pretty egregious.

src/plan_queue.rs
cdc-as81: Agreed.
I removed DistinguishedQueue and add_distinguished_plan. add_startup_plan and add_shutdown_plan now inline their queue-specific insertion logic, making the target queue and bookkeeping explicit.

src/plan_queue.rs
cdc-as81: Agreed. I removed pop_next_distinguished. pop_next_startup and pop_next_shutdown now each pop their respective heap and skip canceled entries directly

src/context.rs
cdc-as81: Agreed with the suggested design. I removed the separate startup_complete field and replaced the shutdown-only status with a unified EventLoopStatus. Startup, regular execution, shutdown, and abort/resume behavior are now represented by explicit states and transitions in the event-loop state machine.

cdc-as81:
> The issue described how to expand the event loop state machine in a way that expands it to startup time. This implementation uses a separate state tracking mechanism that obfuscates which states and transitions are possible. If there is a reason, I need an explanation.
> 
> Also, don't let the LLM factor out single lines of code into useless helpers. Introducing `DistinguishedQueue`, `add_distinguished_plan`, and `pop_next_distinguished` is pretty egregious.

The separate startup-progress field is gone, and the lifecycle is now represented by a single explicit EventLoopStatus state machine. I also removed DistinguishedQueue, add_distinguished_plan, and pop_next_distinguished, and inlined the queue-specific logic. I added tests for abort and shutdown during startup, preserving pending shutdown across an abort, and single-step state transitions.

RobertJacobsonCDC:
A diagram depicting the event loop state machine proposed by this implementation. 

EventLoopStateMachine

RobertJacobsonCDC:
After discussion with the dev team, we decided not to merge this feature.

Rationale: 

The primary use case is to decouple scheduling startup-time plans from the specific concrete simulation start time, which might not be time 0.0. In other words, in the presence of negative time, using this feature users could schedule plans at startup without knowing when the specific startup time is.

But negative start time is already considered a niche use case, so it was thought that burdening the modeler with having to know that concrete simulation start time was not too onerous—if they are using a negative start time, they already know what that time is, and so they can schedule plans at that concrete time if they need to.

In fact, negative start time is really mostly useful when you don't have ergonomic triggers that trigger on when you want the simulation proper to begin. We now how a suite of triggers, and so negative time is less useful as a mechanism to implement simulation burn-in. (Also, we are implementing bulk population import / export, so a "burned in" population can be serialized more easily.)

It might be worth investigating how much code complexity and maintenance burden supporting the negative time feature costs. (My memory is that it isn't much.) If it is significant, we can deprecate it. For now, let's leave it as it is.

PR #1005, opened 2026-07-13
feat: Add on_unsubscribe hook to IxaEvent

### LINKED ISSUE #1003: Add `on_unsubscribe` to `IxaEvent`.
When `Context::subscribe_to_event<E: IxaEvent>` is called, the `E::on_subscribe` method is called after the context registers the subscriber. We recently added a mechanism to unsubscribe a listener from an `IxaEvent` with `Context::unsubscribe_from_event`, but we didn't think to add a method `IxaEvent::on_unsubscribe` to be symmetric. When `Context::unsubscribe_from_event<E: IxaEvent>` is called, `E::on_unsubscribe` should be called after a subscriber is unregistered. As with `IxaEvent::on_subscribe`, the trait should supply a default trivial implementation for `IxaEvent::on_unsubscribe` for implementors that don't use it.

So far we haven't used this method for anything, but I am considering at least one use case: caching the fact that a particular `IxaEvent` has any subscribers in order to avoid constructing the concrete `IxaEvent` and calling `context.emit_event` in hot loops. (Benchmarking is required to determine if this is worth it.)

Out of scope for now, but for Ixa v3.0, we should consider:
1. Removing `on_subscribe` and `on_unsubscribe` altogether if we still don't have a good use case at that time.
2. Or changing their signatures to also accept the `EventListenerId<E>` value of the added/removed subscriber. That way an `IxaEvent` can actually change the subscription status based on its own logic if it wants to.

cdc-as81:
## Summary

  - Add a default `IxaEvent::on_unsubscribe` lifecycle hook.
  - Invoke the hook after a listener is successfully removed.
  - Avoid invoking the hook for unknown or previously removed listeners.
  - Add tests covering hook timing, multiple listeners, and unsuccessful removals.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -54,8 +54,13 @@ struct EventHandlerRegistration<E: IxaEvent> {
 }
 
 pub trait IxaEvent: Copy + 'static {
-    /// Called every time `context.subscribe_to_event` is called with this event
+    /// Called after [`Context::subscribe_to_event`] registers a listener for
+    /// this event type.
     fn on_subscribe(_context: &mut Context) {}
+
+    /// Called after [`Context::unsubscribe_from_event`] successfully removes a
+    /// listener for this event type.
+    fn on_unsubscribe(_context: &mut Context) {}
 }
 
 /// An enum to indicate the phase for plans at a given time.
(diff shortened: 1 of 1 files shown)


RobertJacobsonCDC (approved):
Yep, perfect! Thanks!

PR #1011, opened 2026-07-16
fix: handle filtered criterion summaries

### LINKED ISSUE #1010: `sample_single_excluding` panics when you run a named benchmark group
Suppose you want to run a specific named benchmark group, like `counts`, say:

```bash
cargo bench -p ixa-bench counts -- --save-baseline counts_baseline
```

You will get output that ends with something like:

```text
     Running criterion/sample_single_excluding.rs (target/release/deps/sample_single_excluding-0b48071fd02d473f)
Gnuplot not found, using plotters backend

=== Strategy comparison: sample_single_excluding (ns/sample) ===
         n     rejection     iteration          auto      winner

thread 'main' (44590844) panicked at ixa-bench/criterion/sample_single_excluding.rs:115:24:
no entry found for key
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: bench failed, to rerun pass `-p ixa-bench --bench sample_single_excluding`
[bench:create] ERROR task failed
```

What I think is happening is, the criterion harness has a filtering mechanism that ends up "running" all groups, but individual groups choose to run or not based on the filter. Because the `sample_single_excluding` doesn't actually run, it doesn't generate any data, but that group still tries to print a report to the console and tries to fetch a key that was never inserted into its accumulated results data structure.

This seems to only occur when you are either creating or comparing a baseline, so the `-- --save-baseline` flag (or `-- --baseline`) is required to trigger this behavior. The following doesn't seem to be affected, for example:

```bash
cargo bench -p ixa-bench --bench counts
```

cdc-as81:
## Summary

  - Prevent `sample_single_excluding` from panicking when Criterion filters skip benchmarks.
  - Print comparison summaries only for slice sizes with results from all three strategies.
  - Skip the summary when no complete comparison rows are available.
  - Add regression tests for empty, single-strategy, and partially complete results.


RobertJacobsonCDC (changes requested):
This is kind of a nit pick, but can we put everything in the single `ixa-bench/criterion/sample_single_excluding.rs` file instead of having a new file `ixa-bench/criterion/sample_single_excluding_summary.rs` for the auxiliary types that are only used in `sample_single_excluding.rs` and a separate file `ixa-bench/tests/sample_single_excluding_summary.rs` just for tests? If we really need tests, just make it a `tests` submodule in `ixa-bench/criterion/sample_single_excluding.rs` like we normally do.

Also, do we really need the tests for this? I'd be tempted to leave them out, personally. But it's up to you.

RobertJacobsonCDC (approved):
LGTM.

PR #1020, opened 2026-07-27
feat: add whole-population CSV import and export

### LINKED ISSUE #920: Implement population export/import
Would it be useful to have `Context` be cloneable? We'd have to think through the consequences for things like data plugins.

The use cases that make the most sense are testing and cloning an initialized `Context` instance for, say, calibration and experiment runs.

To make the second category of use cases feasible, we might think about a global RNG reseeding mechanism.

cdc-as81:
## Summary

  Adds support for exporting and importing complete entity populations as a
  versioned directory of CSV files.

  - Adds `ContextPopulationExt` with `export_population` and `import_population`.
  - Preserves entity IDs and all persistable, non-derived property values.
  - Recomputes derived properties and rebuilds configured indexes during import.
  - Emits `EntityCreatedEvent` for imported entities.
  - Validates the complete manifest, schema, files, IDs, and property values before
    mutating the target context.
  - Automatically enables persistence for compatible `define_property!` types.
  - Adds `persist = true` support for properties declared with `impl_property!`.
  - Documents the file format, supported scalar values, and persistence behavior.

  Population exports intentionally exclude networks, global properties, RNG state,
  plans, simulation time, plugins, event handlers, and indexes.

### DIFF
--- a/.github/workflows/release-plz-pr.yaml
+++ b/.github/workflows/release-plz-pr.yaml
@@ -78,7 +78,7 @@ jobs:
         run: |
           rustup toolchain install
       - name: Run release-plz
-        uses: release-plz/action@e8792575c7f2366cf6ff3ccc33ead9ace5b691c7 # v0.5.130
+        uses: release-plz/action@2eb1d8bcb770b4c48ccfaad919734b38b51958c9 # v0.5.131
         with:
           command: release-pr
         env:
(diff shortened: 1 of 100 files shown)


RobertJacobsonCDC:
This was a very useful prototype. I think the result of our conversation about it is that this isn't the implementation we want. Let me prototype an alternative. The big differences are:

- use a binary serialization, which should be a little more efficient and apply to a larger family of property types
- serialize _everything_ (all properties for all entities), which gets rid of the need to have a registration infrastructure that selects which properties participate in population persistence
- put a universal trait constraint on `Property<E>`, requiring every property to be serializable

PR #1022, opened 2026-07-27
fix: seal ContextBase trait

### LINKED ISSUE #1019: Seal `ContextBase` trait
This is not meant to be externally implemented, so we might as well seal the `ContextBase` trait.

cdc-as81:
## Summary

  Seal `ContextBase` so downstream crates can use it as a trait bound but cannot implement it.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -778,7 +778,13 @@ impl Context {
     }
 }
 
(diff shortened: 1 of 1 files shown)


RobertJacobsonCDC:
Decided not to implement.

Issue #1023, opened 2026-07-28
Should we align limits on length of initialization list to total number of properties allowed?


RobertJacobsonCDC:
We implement `PropertyList<E>` (or `InitializationList<E>`) for tuples up to length 20. However, we do not have a hard cap on the number of properties an entity can have. As a consequence, if we have an entity with *more* than 20 properties, there is currently no way to initialize more than 20 of those properties for a new entity. Some alternatives:

1. We leave it this way. We assume >20 properties is an obscure use case, and client code can just deal with the limitation on length of initialization list themselves.
2. We give a hard limit to the number of allowed properties one can define on an entity. 
  - This is convenient for certain internals implementation (imagine storing a `u32` bit mask that holds some boolean property for all of the properties of an entity).
  - Aligns the hard cap on initialization length to the cap on number of allowed properties. 
3. Provide a mechanism to initialize an arbitrary number of properties. (This should be easy to do now that we require the `with!` syntax.)

Note that we don't have the same limitation for queries, because you can always construct a query of more than 20 properties by AND'ing multiple queries—internally it constructs the same data structure.

k88hudson-cfa:
Let's revisit if we start hitting this limit

PR #1025, opened 2026-07-29
feat: Emit `PropertyInitializedEvent` in `add_event`

### LINKED ISSUE #1000: Emit `PropertyInitializedEvent<E, P>` for non-default property values on entity creation
# Emit `PropertyInitializedEvent<E, P>` for non-default property values on entity creation

Motivated by #991, this feature would make non default initial property values observable in client code.

Add a `PropertyInitializedEvent<E, P>` that is emitted for each non-default property value that a new entity is created with (e.g. via `context.add_entity(with!(Person, value1, value2, ...))`). This gives subscribers a way to react to the initial property values of newly created entities, analogous to how `PropertyChangeEvent<E, P>` lets them react to changes on existing entities.

## Event semantics

Each `PropertyInitializedEvent<E, P>` instance holds:

- a copy of the non-default property value, and
- the `EntityId` of the newly created entity.

For each subscriber to `PropertyInitializedEvent<E, P>`, a callback delivering the event is queued on the callback queue.

**Only non-default values trigger the event.** If an entity is created with a property left at its default, no event is emitted for that property.

**Only non-derived properties emit the event** (see below).

## Background: how property values get set today

There are two circumstances in which a property value is set, and they currently use different execution paths:

- **Setting a property on an existing entity** — `context.set_property(entity_id, property_value)`. This has a fast path and a slow path. On the fast path (no change-event subscribers, no value change counters, and no index — checked for the property *and* all its dependents), the value is written directly into backing storage without recording the previous value. On the slow path, the previous value is recorded, storage is updated, indexes and value counts are updated, and callbacks are queued for all event subscribers.
- **Creating a new entity** — `add_entity` calls `PropertyList::set_values_for_new_entity`, which bypasses `set_property` entirely. The value is written directly to storage (if it differs from the default), and then `add_entity` updates any indexes and value change counts.

This feature adds event emission to the second path.

## Optional refactor (include with this task)

The entity-creation path (`set_values_for_new_entity`) and the `set_property` path duplicate similar logic (writing storage, updating indexes/counts, and now emitting events). As part of implementing this feature, evaluate whether it's worth refactoring the creation path so that it mirrors the `set_property` internals.

One candidate shape: a private `context.initialize_property` method that sets the value, updates indexes, and emits the `PropertyInitializedEvent<E, P>`. Because all concrete property types are only known within the `PropertyList` trait, `initialize_property` would be invoked through that trait.

This refactor is optional but is bundled into this task because anyone implementing the feature will be working in exactly this code anyway. Whether it's actually worthwhile won't be clear until implementation is underway, so treat it as an evaluate-and-decide step rather than a requirement.

## Derived properties: out of scope for now

Do **not** emit `PropertyInitializedEvent<E, P>` for dependent derived properties.

The reasoning: derived properties have no true default value, since they're computed from the properties they depend on. They do have a *de facto* initial value — the value computed from the defaults of their dependencies — but we can't cheaply tell whether the value computed from a non-default dependency actually differs from that de facto initial value. Emitting the event only sometimes, for a value that may or may not match what the entity "would have had" under all-default inputs, would be inconsistent and confusing. For this reason, **only non-derived properties emit the event.** Revisiting derived properties can be a follow-up if a use case emerges.

## Performance requirement

Regardless of whether the optional refactor is done, we must measure `add_entity` performance in the case where non-default property values *are* initialized but there are **no subscribers** to the corresponding `PropertyInitializedEvent<E, P>`. The goal is to confirm there is no performance penalty for adding this feature when it isn't being used.

We should also verify there's no regression for the default-value case, though the risk there is lower.

## Acceptance criteria

- `PropertyInitializedEvent<E, P>` is emitted for each non-default, non-derived property value supplied at entity creation, carrying the value and the new `EntityId`.
- No event is emitted for default values or for derived properties.
- Subscribers receive queued callbacks consistent with existing event-delivery behavior.
- A benchmark confirms no performance penalty for `add_entity` with non-default initialized values and no subscribers.

RobertJacobsonCDC:
Implements `PropertyInitializedEvent<E, P>` (#1000). 

To make the implementation zero cost, we cache the fact that a property has `PropertyInitializedEvent` subscribers in a bit mask owned by `PropertyStore<E>`. This puts a hard cap on the number of properties an entity can have of 32. If we ever encounter a use case for more than 32 properties, we can increase this cap to 64 or 128, or we can find an alternative that eliminates the cap—but I figure we can cross that road if and when we come to it. (See also #1023.)

While I was at it, I also added a flag to cache the fact that there exist `EntityCreatedEvent` subscribers, which speeds up `add_entity` by a handful of percentage points. It's not huge, but the implementation is also pretty simple, so I figure it pays for itself.

### DIFF
--- a/src/data_structures/bit_set.rs
+++ b/src/data_structures/bit_set.rs
@@ -1,0 +1,317 @@
+//! A semi-dynamic bit mask data structure the size of which is fixed at construction.
+//!
+//! Supports very efficient "is empty" check, `get`, and `set` operations. A `BitSet` allocates all
+//! of its storage up front. The storage is chunked into 64 bit words, which means allocated
+//! capacity can be greater than the value of `bit_count` the `BitSet` was constructed with. For
+//! simplicity, we allow the entire capacity to be used (with `set` and `get`). As a consequence,
+//! the actual value of `bit_count` the `BitSet` was constructed with is not recoverable after
+//! construction.
+//!
+//! ## Implementation
+//!
+//! Storage is inline for <=64 bits, which makes access and cloning very cheap for the
+//! overwhelmingly common case. Internally we use a `set_count` to make `is_empty` very fast
+//! independent of whether or not storage is inline.
+
+pub struct BitSet {
+    /// 64 bits inline or heap allocated spillover.
+    storage: Storage,
+    /// Count of set bits, for O(1) empty check. Maintained in mutation operations.
+    set_count: u32,
+}
+
+enum Storage {
+    /// Inline storage for bit counts <= 64. No heap allocation; `Clone` is a register copy.
+    Inline(u64),
+    /// Heap storage for bit counts > 64. Fixed length, allocated once.
+    Heap(Box<[u64]>),
+}
+
+impl BitSet {
+    pub fn new(bit_count: usize) -> Self {
+        let storage = if bit_count <= 64 {
+            Storage::Inline(0)
+        } else {
+            let num_words = bit_count.div_ceil(64);
+            Storage::Heap(vec![0u64; num_words].into_boxed_slice())
+        };
+        Self {
+            storage,
+            set_count: 0,
+        }
+    }
+
+    #[inline(always)]
+    pub fn is_empty(&self) -> bool {
+        self.set_count == 0
+    }
+
+    /// Clones this `BitSet` if it contains any set bits.
+    #[inline(always)]
+    pub fn clone_if_nonempty(&self) -> Option<Self> {
+        if self.is_empty() {
+            None
+        } else {
+            Some(self.clone())
+        }
+    }
+
+    /// Returns the total number of bits that can be stored in this `BitSet`.
+    #[inline(always)]
+    pub fn capacity(&self) -> usize {
+        match &self.storage {
+            Storage::Inline(_) => 64,
+            Storage::Heap(words) => words.len() * 64,
+        }
+    }
+
+    /// Returns `true` if the `n`th bit is set, `false` otherwise.
+    #[inline(always)]
+    pub fn get(&self, n: usize) -> bool {
+        match &self.storage {
+            Storage::Inline(word) => (word >> n) & 1 != 0,
+            Storage::Heap(words) => {
+                let word = n >> 6; // n / 64
+                let bit = n & 63; // n % 64
+                (words[word] >> bit) & 1 != 0
+            }
+        }
+    }
+
+    /// Sets the `n`th bit.
+    #[inline(always)]
+    pub fn set(&mut self, n: usize) {
+        match &mut self.storage {
+            Storage::Inline(word) => {
+                let mask = 1u64 << n;
+                if *word & mask == 0 {
+                    *word |= mask;
+                    self.set_count += 1;
+                }
+            }
+            Storage::Heap(words) => {
+                let idx = n >> 6;
+                let mask = 1u64 << (n & 63);
+                let prev = words[idx];
+                if prev & mask == 0 {
+                    words[idx] = prev | mask;
+                    self.set_count += 1;
+                }
+            }
+        }
+    }
+
+    /// Clears the `n`th bit.
+    #[inline(always)]
+    pub fn reset(&mut self, n: usize) {
+        match &mut self.storage {
+            Storage::Inline(word) => {
+                let mask = 1u64 << n;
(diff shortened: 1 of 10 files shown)


RobertJacobsonCDC:
> To make the implementation zero cost, we cache the fact that a property has PropertyInitializedEvent subscribers in a bit mask owned by PropertyStore<E>. This puts a hard cap on the number of properties an entity can have of 32. If we ever encounter a use case for more than 32 properties, we can increase this cap to 64 or 128, or we can find an alternative that eliminates the cap—but I figure we can cross that road if and when we come to it. (See also https://github.com/CDCgov/ixa/issues/1023.)

Feedback from discussion:

- A hard cap of 32 seems too restrictive.
- Let's benchmark an unrestricted implementation. There's some reasonable skepticism that a single pointer chase would really matter for performance in this use case.

RobertJacobsonCDC:
It turns out that we make a copy (or clone) of the bit mask cache of which properties have `PropertyInitializedEvent` subscribers in order to avoid a double borrow during mutation. I implemented a `BitSet` data structure with 64 bits of inline storage with heap spillover. It performs very close to a raw bit mask when the number of properties is <= 64. For more than 64 properties, we incur a pointer dereference, but only if the `BitSet` is not completely empty, which is acceptable.

The price we pay for removing the hard cap on number of properties is the extra code of the `BitSet` implementation, but actually this is a very simple and easy to understand data structure, and the fact that it has a nice API makes the sites of its _usage_ easier to understand and reason about.

src/macros/property_impl.rs
```diff
@@ -194,6 +197,8 @@ impl_property!(
 /// ### Notes
 ///
 /// - By default, the generated type derives `Debug`, `PartialEq`, `Eq`, `Hash`, `Clone`, and `Copy`.
+/// - An entity may have at most 32 registered properties, including explicit, constant, derived,
```
k88hudson-cfa: This is not true anymore, right?

src/entity/context_extension.rs
```diff
@@ -322,11 +323,18 @@ impl ContextEntitiesExt for Context {
...
+
+        // We simultaneously fetch the index count and set of property initialized event
+        // subscriptions from the entity store without retaining a borrow of the store.
+        let (index_count, property_initialized_event_subscriptions) = {
+            let property_store = self.entity_store.get_property_store::<E>();
+            (
+                property_store.index_new_entity_fns.len(),
+                property_store
+                    .property_initialized_event_subscriptions
+                    .clone(),
```
k88hudson-cfa: Couldn't you conditionally clone on when `property_initialized_event_subscriptions` being non-empty (like just move it inside the later check?)

src/data_structures/bit_set.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+    }
+
+    /// Avoids bounds check.
+    ///
+    /// # Safety
+    /// Use this if you can guarantee `n` is within bounds.
+    #[inline(always)]
+    pub unsafe fn get_unchecked(&self, n: usize) -> bool {
```
k88hudson-cfa: It doesn't look like you're using this anywhere, can we just remove it?

k88hudson-cfa (approved):
Looks good with nits fixed

src/entity/context_extension.rs
RobertJacobsonCDC: This clone is so cheap that it can't be measured.

src/data_structures/bit_set.rs
RobertJacobsonCDC: Benchmarking shows the performance difference is in the noise. One of the interesting performance lessons for me in the last 8-9 years working with Rust is how Rust's bounds checking is virtually free. The compiler can elide the bounds check in a surprisingly large number of cases.

src/macros/property_impl.rs
RobertJacobsonCDC: Yep, looks like I accidentally left this language in a couple of places. Fixed.

src/entity/context_extension.rs
RobertJacobsonCDC: Well... we could have a `BitSet::clone_if_nonempty() -> Option<BitSet>` and then later do `if let Some(subscriptions) = subscriptions {...}`. Maybe that is better. Ok, fine, you've convinced me.

PR #1028, opened 2026-07-30
feat: Added "internals" MDBook, Architectural Decision Records

### LINKED ISSUE #570: Implementation documentation

RobertJacobsonCDC:
This PR adds an "internals" MDBook as a home for:

- Architectural Decision Records (ADRs): blog-like descriptions of design decisions. These are historical snapshots that capture the design at the time of acceptance, so they do not need to be updated and maintained as the code evolves.
- Research results: similar to ADRs, but not connected to internal implementation. These are research tasks that are useful to record but that are not connected to a PR or feature implementation. 
- Developer guides: developer facing documentation. Unlike the other two categories, these are meant to be maintained and kept in sync with the codebases present implementation. Right now its just a quick guide to running benchmarks and benchmark comparisons. 

The determination for inclusion is "useful."


docs/internals/.gitignore
```diff
@@ -0,0 +1,15 @@
+## Project Specific ##
```
k88hudson-cfa, line 1: Is all of this actually needed?

docs/internals/src/adrs/0007-introduce-composable-triggers.md
```diff
@@ -0,0 +1,277 @@
...
+- [Criterion/specification split commit `c8c91df`](https://github.com/CDCgov/ixa/commit/c8c91df65bfb290f352e3c52f85c429c6329a2ed)
+- [Toggling trigger commit `97fc3f7`](https://github.com/CDCgov/ixa/commit/97fc3f7dd15b0353e6b3384cc95a7cc3cbfffae7)
+- [Aligned builder commit `932066b`](https://github.com/CDCgov/ixa/commit/932066bb7a13aa60272000d50f1e2ac1c32f3566)
+- [Periodic time criterion commit `bf31a24`](https://github.com/CDCgov/ixa/commit/bf31a24ae765ca77c0a069378dd8670f859c7298)
+- [Retained aggregate feature commit `8331ecc`](https://github.com/CDCgov/ixa/commit/8331eccaf82d181b1dd27296b282c644771162c5)
+- [Merged commit `e7bddf0`](https://github.com/CDCgov/ixa/commit/e7bddf055f422469b9087a8547e5cda8c0144b3f)
+
+The trigger files in `8331ecc` and `e7bddf0` are identical; their complete
+trees differ because their respective bases contain unrelated context changes.
+The later commits `a422158` and `d5644de` correspond to the same documentation
```
k88hudson-cfa, line 275: 8331ecc and a422158 don't exist on GitHub, are these maybe local commits?

docs/internals/src/research/0001-potential-property-index-architectures.md
```diff
@@ -0,0 +1,92 @@
...
+- The `ValueCountIndex` itself is a specific kind of aggregator, one that counts how many entities have each value for a property. We can imagine other kinds of aggregation.
+- We have a whole `Network` module implementing a directed graph system that can be used, for example, to implement contact networks. But this same functionality could be replaced with a particular kind of index. Imagine we have a `Person` entity and a ZST naming the type of network that we'll generically call `Network`. Then we can define another entity `Edges&lt;Person, NetworkType&gt;` with a property `ToPerson`. If we implement a "dense" index in which the keys are `EdgeID` values that are used to just index into a vector and the values are stored in a vector, say, then a "Network" is just an `Edges&lt;Person, NetworkType&gt;` with a property `ToPerson` having just such an index. We'd just identify each `EdgeId` with the `PersonId` the edge originates _from_ and the value of `ToPerson` as the `PersonId` the edge terminates at. (This only models 1-to-many relationships but could easily be modified to many-to-many relationships.)
```
k88hudson-cfa, line 16: It seems really annoying that you can't use < / > in inline code, is there not some mdbook setting that handles this?

docs/internals/src/adrs/0002-represent-query-results-with-entity-set.md
```diff
@@ -0,0 +1,161 @@
+# ADR-0002: Represent query results with `EntitySet`
+
+| Field | Value |
+| --- | --- |
+| Decision date | 2026-02-27 (merge of PR #786) |
+| Recorded | 2026-07-29 |
+| Status | Accepted |
+| GitHub issue | [#746](https://github.com/CDCgov/ixa/issues/746) |
+| Pull requests | [#745](https://github.com/CDCgov/ixa/pull/745), [#786](https://github.com/CDCgov/ixa/pull/786) |
+| Feature branches | `RobertJacobsonCDC_746_entity_set`; precursor `RobertJacobsonCDC_v2.0.0_rename_query_result_iterator` |
```
k88hudson-cfa, line 10: I would personally omit the branch names since they tend to get deleted/cleaned up over time; the PR is the right artifact IMO

docs/internals/adr-authoring-guide.md
```diff
@@ -0,0 +1,321 @@
+# ADR Authoring Session Guide
```
k88hudson-cfa, line 1: I think this would be fine to include in a .codex/skills directory if you want, some comments: 

- You could take the ADR number as input to the skill, as well as any additional documents (e.g. notes)
- It's pretty specific to your own local file directory, I would clean that up. I would also instruct it to refer to notes, commits, etc. only on the remote repository not the user's local machine 
- I think it's more verbose than necessary in a lot of places, e.g. "It is not responsible for selecting the next ADR, renumbering the inventory, integrating navigation, or disposing of source files. This narrow ownership allows several ADRs to be authored concurrently without conflicting edits.", I would try to cut it down
- I think the parallelism approach here  is unnecessarily complicated and can be much simpler. Just like instruct the agent to use a worktree with a branch naming convention for the adr, and tell it to avoid making changes outside of new file deliverables. 

For the validation loop:
- I don't really get the purpose of the summary for interactive review, don't you want to just review the text in full? 
- I would have an explicit check for the correctness of commit hashes, PRs, dates to make sure they are accurate/not hallucinated and exist remotely

k88hudson-cfa (changes requested):
I added some comments on your proposed skill and some minor things in line.

Overall, there are quite a few references to your own local Notes directory in ADRs that I would do a pass to remove or reframe

docs/internals/src/research/0001-potential-property-index-architectures.md
```diff
@@ -0,0 +1,92 @@
+# Research-0001: Potential property index architectures
```
k88hudson-cfa, line 1: Do we need to keep these

docs/internals/src/research/0001-potential-property-index-architectures.md
RobertJacobsonCDC: Ach, yeah, the fix is actually to just escape the first `<`, which we do everywhere else. This is AI slop leaking.

docs/internals/src/adrs/0002-represent-query-results-with-entity-set.md
RobertJacobsonCDC: And as with the commit numbers they are sometimes local, which isn't exactly useful.

docs/internals/adr-authoring-guide.md
RobertJacobsonCDC: I go back and forth about whether it's a good idea to commit a skill. Maybe we just omit this altogether? Or just have an empty template?

docs/internals/src/research/0001-potential-property-index-architectures.md
RobertJacobsonCDC: Arguably no. Are you suggesting we eliminate all the research items? If there is research we did to determine whether we should adopt a particular design, and the result was we shouldn't adopt the design, then that could be captured in a proper ADR. Other research artifacts are not as well justified.

docs/internals/.gitignore
RobertJacobsonCDC: No, it shouldn't be there. I'll just add the output directory to the top level gitignore.

docs/internals/src/adrs/0007-introduce-composable-triggers.md
RobertJacobsonCDC: Yeah, looks like it. Mentioning a commit hash at all is of dubious utility anyway, I think. I'll do another pass over the ADR content itself to evaluate usefulness.

PR #1029, opened 2026-07-31
chore: harden unwrap and panic handling

### LINKED ISSUE #979: Have a more consistent treatment of internal asserts and errors
This issue is about developing a more consistent approach to doing internal checks that only fail if there is a bug in Ixa core itself and the messages that are reported if those checks fail.

We have a lot of internal uses of:
- `assert!` (and equivalent)
- `debug_assert!`
- `.expect(...)`
- unguarded uses of `unwrap`

Many of these have messages for the panic cases.

Some issues and questions:
- Most panic messages don't tell the user that the panic represents a bug in Ixa internals rather than a client-side problem. They should.
- We should have a clearer understanding of when to use an `assert!` versus a `debug_assert!` versus naked `unwrap`.
  -  use `unwrap` for infallible unwrapping
  - use `assert!` for checks that should be run even in production builds
  - use `debug_assert!` for checks that should be done in debug/dev builds but not in production
- Most uses of naked `unwrap` should probably have a code comment justifying it. We should consider adopting an lint for this.
- When should we promote an `assert!` / `debug_assert!` to a naked `unwrap!` (if ever)?
  - Which existing instances of `assert!` / `debug_assert!` can we justify promoting to `unwrap!`?
  - How can we track maturation of assertion instances? (If we decide to "mature" assertions at all.)

cdc-as81:
## Summary

- Improve panic diagnostics by distinguishing internal Ixa invariants from client-triggerable failures.
- Document public panic contracts across context, entity, random, reporting, and trigger APIs.
- Preserve locally infallible `unwrap()` calls with justification instead of enabling a workspace-wide `clippy::unwrap_used` lint.
- Clarify weighted-sampling preconditions and add regression coverage for invalid inputs.
- Record the review triage and resolutions in `comments.md`.

## Testing

- `mise run prepush`


examples/basic-infection/src/infection_manager.rs
```diff
@@ -14,8 +14,9 @@ define_rng!(InfectionRng);
 
 fn schedule_recovery(context: &mut Context, person_id: PersonId) {
     trace!("Scheduling recovery");
-    let infection_duration =
-        context.sample_distr(InfectionRng, Exp::new(1.0 / INFECTION_DURATION).unwrap());
+    let recovery_distribution = Exp::new(1.0 / INFECTION_DURATION)
+        .expect("basic-infection requires a positive infection duration");
```
RobertJacobsonCDC: This is a good example of when a naked `unwrap` is well justified and `clippy::unwrap_used` is annoying:
- `INFECTION_DURATION` is a constant we have complete control over, so we can guarantee this call is infallible.
- This is example code. There doesn't seem to be any pedagogical benefit to handling the (provably impossible) error case. It just gets in the way.
- An `#[allow(clippy::unwrap_used)]` annotation would introduce noise.

examples/basic-infection/main.rs
```diff
@@ -1,10 +1,10 @@
 use ixa::runner::run_with_args;
 use ixa_example_basic_infection::initialize;
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>> {
```
RobertJacobsonCDC: I don't understand how this is better than calling `unwrap`.

src/entity/entity_set/entity_set_iterator.rs
```diff
@@ -65,19 +65,30 @@ impl<'a, E: Entity> EntitySetIteratorInner<'a, E> {
...
 
                 if sets.iter().all(EntitySet::is_source_leaf) {
                     // `EntitySet::Intersection` stores operands sorted ascending by `cost_hint`.
                     // Keep that order so membership checks short-circuit early on small filters.
                     let mut set_iter = sets.into_iter();
 
-                    let first = set_iter.next().unwrap().into_source_leaf().unwrap();
+                    let first = set_iter
+                        .next()
+                        .expect("Ixa internal error: nonempty intersection has no first set")
```
RobertJacobsonCDC: If we can prove the `unwrap` is infallible (nonempty by construction), we should just call `unwrap`.

src/entity/context_extension.rs
```diff
@@ -199,6 +218,10 @@ pub trait ContextEntitiesExt {
     ///     },
     /// );
     /// ```
+    ///
+    /// # Panics
+    ///
+    /// Panics if `period` is not finite and strictly positive.
```
RobertJacobsonCDC: This wording is awkward. Better:

> Panics if `period` is not a strictly positive finite number.

src/entity/context_extension.rs
```diff
@@ -360,7 +383,7 @@ impl ContextEntitiesExt for Context {
         entity_id: EntityId<E>,
         property_value: P,
     ) {
-        debug_assert!(!P::is_derived(), "cannot set a derived property");
+        assert!(!P::is_derived(), "cannot set a derived property");
```
RobertJacobsonCDC: What is the reasoning here?

src/entity/entity.rs
```diff
@@ -87,7 +87,7 @@ pub trait Entity: Any + Default {
     #[must_use]
     fn name() -> &'static str {
         let full = std::any::type_name::<Self>();
-        full.rsplit("::").next().unwrap()
+        full.rsplit("::").next().unwrap_or(full)
```
RobertJacobsonCDC: What is the reasoning here?

src/entity/entity_store.rs
```diff
@@ -63,7 +65,9 @@ pub fn register_property_with_entity(
     property_type_id: TypeId,
     required: bool,
 ) {
-    let mut builder = ENTITY_METADATA_BUILDER.lock().unwrap();
+    let mut builder = ENTITY_METADATA_BUILDER
+        .lock()
+        .expect("Ixa internal error: entity metadata registry lock is poisoned");
```
RobertJacobsonCDC: This lock is infallible in the absence of a panic. In the presence of a panic, the program crashes, and this function is never called.

RobertJacobsonCDC (changes requested):
The interesting part of Issue #979 is the policy questions. This PR:

- Appears to ignore the policy question, "Do we have a process for "maturing" internal checks through  `assert!` / `debug_assert!` to a naked `unwrap!`?"
- Rejects the criteria proposed in #979 for when to use `unwrap`, `assert!`, and `debug_assert!`, in favor of an alternative criteria, but it doesn't discuss the reasoning.
- Chooses to adopt the `clippy::unwrap_used` lint (which #979 says we should consider) but doesn't really discuss the consequences. I think this lint is actually rather problemmatic, because it disallows a naked `unwrap` on _every_ `add_entity`, `sample_entity`, `get_global_property`, and others, even though local reasoning can often guaranteed these calls are infallible. Could we enable the lint for ixa core internals but disable it for all example code? Or are there reasons why it is too heavy handed even in ixa core internals?



I don't like the changes to the examples at all. The changes to the `births-deaths` example in particular feel overwrought. This is meant to be an illustrative simple example, not hardened production code. Principled error handling makes sense when it illustrates a best practice we want to model for the reader. Otherwise, it just gets in the way. In some cases it illustrates anti-patterns, such as calling `expect` on `Exp::new(FOI)`, where `FOI` is a known constant defined at the top of `examples/basic-infection/src/lib.rs`.



There _are_ some cases in internal implementation where an `unwrap` is provably infallible. Part of the  purpose of #979 is to document these cases (presumably with a code comment), not to rewrite them to use `expect` or similar. There are many cases in `src/entity/entity_set/entity_set_iterator.rs`, for example. This PR inexplicably changes

```rust
if sets.len() == 1 {
    return Self::from_entity_set(sets.pop().unwrap());
}
```

to

```rust
if sets.len() == 1 {
    return Self::from_entity_set(
        sets.pop()
            .expect("Ixa internal error: singleton entity set vector is empty"),
    );
}
```

This rewrite is clearly strictly worse.



I do like having an "Internal Ixa Error:" prefix for errors that only happen if there is a bug in Ixa itself.



Maybe instead of this PR we discuss the policy questions in a dev meeting.



A safe set of tasks we could do independent of the policy discussion might be:

- Catalog all cases of `unwrap` or `expect` that are actually provably infallible from local reasoning. If they are provably infallible, a naked `unwrap` is justified; `expect` calls can be replaced.
- For all cases where a panic indicates an internal bug in ixa, where we have a panic message, use an "Internal Ixa Error:" (or similar) prefix in the panic message.

src/log/standard_logger.rs
```diff
@@ -47,7 +47,10 @@ impl LogConfiguration {
 
             None => {
                 // The global logger has not yet been initialized
-                self.root_handle = Some(log4rs::init_config(new_config).unwrap());
+                self.root_handle = Some(
+                    log4rs::init_config(new_config)
+                        .expect("failed to initialize the standard logger configuration"),
```
RobertJacobsonCDC: Is this an "Ixa internal error:" situation?

src/random/context_ext.rs
```diff
@@ -13,12 +13,17 @@ use crate::random::{RngHolder, RngPlugin};
 use crate::{Context, ContextBase, RngId};
 
 /// Gets a mutable reference to the random number generator associated with the given
-/// [`RngId`]. If the Rng has not been used before, one will be created with the base seed
-/// you defined in `init`. Note that this will panic if `init` was not called yet.
+/// [`RngId`]. If the RNG has not been used before, one will be created with the current
+/// base seed, which is zero until configured by [`ContextRandomExt::init_random`].
```
RobertJacobsonCDC: > which is zero until configured by [`ContextRandomExt::init_random`].

This wording feels awkward to me, especially the word "until." Here's a suggestion:

```rust
/// Gets a mutable reference to the random number generator associated with the given
/// [`RngId`]. The RNG is created lazily: the first time a given [`RngId`] is used, its
/// generator is initialized from the current base seed. The base seed is `0` unless it
/// has been set with [`ContextRandomExt::init_random`].
```

src/random/context_ext.rs
```diff
@@ -150,7 +175,8 @@ pub trait ContextRandomExt: ContextBase {
     ///
     /// # Panics
     ///
-    /// Panics if the converted probability is outside `[0.0, 1.0]` or is NaN.
+    /// Panics if RNG state is already borrowed or if the converted probability
+    /// is outside `[0.0, 1.0]` or is NaN.
```
RobertJacobsonCDC: > ...or if the converted probability is outside `[0.0, 1.0]` or is NaN.

I suggest the alternative wording:

> ... or if the provided probability `p` is not in the interval `0.0..=1.0`. 

This names the probability and uses Rust interval notation so the interpretation as a closed interval is unambiguous. I don't think we need to acknowledge that it gets converted or that `NaN` is not in the interval, as that seems obvious to me.

src/random/context_ext.rs
```diff
@@ -175,7 +205,8 @@ pub trait ContextRandomExt: ContextBase {
             + PartialOrd
             + Weight,
     {
-        let index = WeightedIndex::new(weights).unwrap();
+        let index = WeightedIndex::new(weights)
+            .expect("weights must contain at least one valid, positive weight");
```
RobertJacobsonCDC: Actually, the "Panics" doc comment and this `expect` message are not correct or complete. The [rand docs](https://docs.rs/rand/0.9.5/rand/distr/weighted/struct.WeightedIndex.html#method.new) list the error cases:
- [`Error::InvalidInput`](https://docs.rs/rand/0.9.5/rand/distr/weighted/enum.Error.html#variant.InvalidInput) when the iterator `weights` is empty.
- [`Error::InvalidWeight`](https://docs.rs/rand/0.9.5/rand/distr/weighted/enum.Error.html#variant.InvalidWeight) when a weight is not-a-number or negative.
- [`Error::InsufficientNonZero`](https://docs.rs/rand/0.9.5/rand/distr/weighted/enum.Error.html#variant.InsufficientNonZero) when the sum of all weights is zero.
- [`Error::Overflow`](https://docs.rs/rand/0.9.5/rand/distr/weighted/enum.Error.html#variant.Overflow) when the sum of all weights overflows.

We probably don't need to repeat all of this in its entirety. Maybe we point to these docs for users who want complete details? I think the important points are that the weights need to be nonempty nonnegative numbers at least one of which is positive. Of course, "at least one of which is positive" already implies nonempty. The prohibition against `NaN` and overflow of the sum seems obvious and overly technical. So maybe say something like:

```rust
/// Draws a random entry out of the list provided in `weights`
/// with the given weights using the generator associated with the
/// given [`RngId`]. The weights need to be a list of nonnegative
/// numbers at least one of which is positive.
///
/// # Panics
///
/// Panics if RNG state is already borrowed, or if `weights` is not a
/// list of nonnegative numbers at least one of which is positive.
```

Then the `expect` message might be:

```rust
.expect("weights must be a list of nonnegative numbers at least one of which is positive");
```

src/triggers/entity_count.rs
```diff
@@ -85,6 +85,11 @@ pub struct EntityCountTriggerEvent<E: Entity> {
...
 
 impl<E: Entity> EntityCountTrigger<E> {
+    /// Creates a trigger that fires when the entity count reaches `threshold`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `threshold` is zero.
     #[must_use]
     pub fn increases_to(threshold: usize) -> Self {
         assert!(threshold > 0, "threshold must be greater than 0");
```
RobertJacobsonCDC, line 96: Should we have a "Panics" doc in cases where we are asserting a requirement? ...probably. But it feels more natural to include the condition in the docs:

> Creates a trigger that fires when the entity count reaches `threshold`. The `threshold` must be positive.

src/context.rs
```diff
@@ -210,14 +210,15 @@ impl Context {
         self.next_event_listener_id = self
             .next_event_listener_id
             .checked_add(1)
-            .unwrap_or_else(|| panic!("event listener id overflow"));
+            .unwrap_or_else(|| panic!("Ixa internal error: event listener id overflow"));
```
RobertJacobsonCDC: We don't need to construct anything, so this should just be `.expect("Ixa internal error: event listener id overflow");`. I wonder why Clippy didn't flag this?

src/context.rs
```diff
@@ -551,12 +555,18 @@ impl Context {
     /// to construct a new data container and store it in the [`Context`].
     ///
     /// Returns a mutable reference to the data container
+    ///
+    /// # Panics
+    ///
+    /// Panics if `T` was not defined with [`define_data_plugin!`](crate::define_data_plugin)
+    /// or its manually implemented registry metadata is inconsistent.
```
RobertJacobsonCDC: Data plugins should never manually implement `DataPlugin`. If that isn't in our docs somewhere, it should be. In any case, we shouldn't imply that `DataPlugin` can be manually implemented here.

The same goes for similar wording on other methods of `DataPlugin` and on `GlobalParameter` in `src/global_properties.rs`.

Actually, it occurs to me that in this case I think we _only_ panic if there is an internal Ixa error, which only happens if there is a bug in Ixa internals. In such cases, I don't think we actually need a "Panics" section of the doc comment, because the panic should never actually happen.

src/hashing.rs
```diff
@@ -91,10 +91,13 @@ impl<const N: usize> rkyv::ser::Positional for EqualityBufferWriter<N> {
 
 impl<const N: usize, E: rkyv::rancor::Source> rkyv::ser::Writer<E> for EqualityBufferWriter<N> {
     fn write(&mut self, bytes: &[u8]) -> Result<(), E> {
-        let end = self.pos + bytes.len();
+        let end = self
+            .pos
+            .checked_add(bytes.len())
+            .expect("Ixa internal error: serialized buffer position overflowed usize");
```
RobertJacobsonCDC: We do not need a `checked_add` here. If we are overflowing here, that means we are addressing more than 18,446,744,073,709,551,616 bytes, which just cannot happen.

RobertJacobsonCDC (changes requested):
This PR I think contributes significantly to our code quality.

- Users are now explicitly told when a panic is caused by an internal Ixa bug instead of a bug in client code ("Ixa internal error:").
- Several panics that can be triggered by client code now how meaningful user-facing messages (e.g. in `ContextRandomExt`, `ContextReportExt`).
- Many instances of `unwrap` now have justifications in code comments.
- User facing API documentation is now much more complete with respect to panic documentation.



Before merging, let's do the following:

- Audit for cases where a panic can only happen if there is an internal Ixa error. In these cases, I think we don't actually need a "Panics" section in the docs, because the panic should never happen, and even if it does happen, it's not something users have control over.
- The issue with `sample_weighted`
- We shouldn't imply the `DataPlugin` or `GlobalProperty` traits can be manually implemented.

examples/basic-infection/src/infection_manager.rs
cdc-as81: Addressed in 340edee. The example uses the locally infallible `unwrap()` again, and the workspace-wide `clippy::unwrap_used` lint is not enabled.

src/entity/entity_set/entity_set_iterator.rs
cdc-as81: Addressed in 340edee. The locally justified `unwrap()` calls were restored in the entity-set iterator.

examples/basic-infection/main.rs
cdc-as81: Addressed in 340edee. The example has the original simple `fn main()` form with `run_with_args(...).unwrap()`.

src/entity/context_extension.rs
cdc-as81: Addressed in 340edee. The documentation now says that `period` must be a strictly positive finite number.

src/entity/context_extension.rs
cdc-as81: Addressed in 340edee. The derived-property invariant uses `debug_assert!` again.

src/entity/entity.rs
cdc-as81: Addressed in 340edee. Restored `unwrap()` and added a local proof comment that splitting any string yields at least one segment.

src/log/standard_logger.rs
cdc-as81: Addressed in 340edee. This is not necessarily an internal Ixa error because `log4rs::init_config` can fail when another component has installed the process-global logger. The message now explains that likely conflict without using the internal-error prefix.

src/context.rs
cdc-as81: Addressed in 340edee. Replaced `unwrap_or_else(|| panic!(...))` with `.expect("Ixa internal error: event listener id overflow")`.

src/triggers/entity_count.rs
cdc-as81: Addressed in 340edee. The main description now says that `threshold` must be positive, and the `# Panics` section remains for the public precondition.

src/random/context_ext.rs
cdc-as81: Addressed in 340edee. The docs and `expect` message now require nonnegative weights with at least one positive weight. Added regression tests for empty, negative, all-zero, and NaN inputs.

src/context.rs
cdc-as81: Addressed in 340edee. Removed wording that implied manual `DataPlugin` or `GlobalProperty` implementations and added a macro-only warning to the `DataPlugin` trait docs. The concise panic sections remain because the public, unsealed traits still permit unsupported manual implementations to reach those failures.

src/random/context_ext.rs
cdc-as81: Addressed in 340edee. Adopted the suggested lazy-RNG wording and applied it consistently to `get_rng`, `sample`, and `sample_distr`.

src/hashing.rs
cdc-as81: Addressed in 340edee. Removed `checked_add`, restored ordinary addition, and retained the following fixed-buffer bounds assertion.

src/random/context_ext.rs
cdc-as81: Addressed in 340edee. The panic documentation now uses the interval `0.0..=1.0`; the earlier conversion note remains to document the generic input contract.

src/entity/entity_store.rs
cdc-as81: Addressed in 340edee. Restored `ENTITY_METADATA_BUILDER.lock().unwrap()`.

cdc-as81:
Follow-up to the [first change-request review](https://github.com/CDCgov/ixa/pull/1029#pullrequestreview-4856760440): the workspace-wide `clippy::unwrap_used` lint and example rewrites are not part of the final PR. Locally infallible unwraps were restored and justified, while internal invariant failures retain contextual `Ixa internal error:` messages. I replied to each inline thread with its specific resolution.

cdc-as81:
Follow-up to the [second change-request review](https://github.com/CDCgov/ixa/pull/1029#pullrequestreview-4963349906): audited the added panic documentation, corrected `sample_weighted` docs and diagnostics, added invalid-weight regression tests, removed wording that implied manual `DataPlugin`/`GlobalProperty` implementation, and addressed the remaining inline wording and simplification requests. I replied to each inline thread with details.

RobertJacobsonCDC (approved):
LGTM!

PR #1033, opened 2026-08-04
feat: Add query profiling feature

### LINKED ISSUE #705: Add query tracking to profiler

RobertJacobsonCDC:
This PR builds upon the work of #966 and supersedes it. In individual commit order, the major changes relative to #966 are:

- It uses RAII guards instead of the `query_profiling_scope.measure(| | { /* thing to measure */ })` pattern (and removes the factored-out helpers that supported this pattern). This simplifies code structure, keeps the feature less invasive on non-"profiling" feature builds, and makes the code more panic-safe.
- It reuses the canonical query-shape resolution already used to locate query indexes, removing the redundant `&str`-keyed registry. It introduces `QueryIdentityId` as a handle to a single (profiler, query shape) identity, the key needed to accumulate the profiling data in `Context`-specific storage.
- Benchmarking uncovered performance regressions even when the "profiling" feature is disabled. The reasons and fixes are:
  - We accidentally clobbered a fast path for empty, singleton, and "all" queries, which I restored.
  - The surrounding code changes made it harder for Rust / LLVM to inline some functions for mysterious reasons. Adding a few `#[inline]` annotations fixed it.

I also rebased onto `main` and cleaned up some LLM-flavored structural and stylistic issues. This PR does not otherwise change the public semantics of the feature implemented in #966. 

A copy+paste of the feature description from #966 follows.

---

## Summary

  Adds query profiling under the `profiling` feature, with timing statistics aggregated per `Context` and query shape.

  - Records timings across eager query APIs, lazy `EntitySetIterator` execution, and `EntitySet` terminal and sampling operations.
  - Groups equivalent queries by entity type and unordered property names; property order and values do not affect query identity.
  - Reports timings in the profiling console output and the `query_timings` array in profiling JSON.
  - Keeps profiling-only types and helpers out of builds where the feature is disabled.

  ## Execution semantics

  `Count` represents **consumed query executions**:

  - Each eager query operation or `EntitySet` terminal operation records one execution.
  - An iterator records one execution after performing query work, regardless of whether it is consumed through `next`, `count`, `nth`, `for_each`, or `fold`.
  - Creating and dropping an unused iterator records nothing.
  - Dropping a partially consumed iterator records one execution containing only the work completed before the drop.
  - Repeated `next` calls after exhaustion do not create additional executions.
  - User callback time in `for_each` and `fold` is excluded.

  Iterator timing is accumulated with RAII and committed once on exhaustion or drop. This gives `count`, `min`, and `max` a consistent unit: one consumed query execution.

  ## Query identity and set algebra

  A query identity consists of the entity type and its sorted property names. For example, queries for `Age(10), Alive(true)` and `Alive(false), Age(20)` share a timing row.

  Set algebra uses a single-unique-origin attribution policy:

  - A profiled set combined with an unprofiled set retains the profiled identity.
  - Two profiled sets with the same identity retain that identity.
  - Two profiled sets with different identities clear profiling attribution rather than charging the composed work to an arbitrary query.

  ## Example console output

  ```text
Query                       Count       Total       Min         Max
------------------------------------------------------------------
Person: (Age, County)           3  42us 100ns  8us 50ns  20us 400ns
Person: (Age)                   1   7us 300ns  7us 300ns   7us 300ns
  ```

  ## Example JSON output

  ```json
{
  "query_timings": [
    {
      "query": "Person: (Age, County)",
      "count": 3,
      "total": 0.0000421,
      "min": 0.00000805,
      "max": 0.0000204
    },
    {
      "query": "Person: (Age)",
      "count": 1,
      "total": 0.0000073,
      "min": 0.0000073,
      "max": 0.0000073
    }
  ]
}
  ```

### DIFF
--- a/docs/book/src/topics/profiling-module.md
+++ b/docs/book/src/topics/profiling-module.md
@@ -118,6 +118,27 @@ See `examples/profiling` in the repository for a more complete example,
 including configuring `report_options()` to control the output directory, file
 prefix, and overwrite behavior.
 
+## Query timings
+
+When the `profiling` feature is enabled, context query APIs automatically
+aggregate query execution times by query shape. A query shape consists of the
+entity type and its unordered property names; property values do not affect the
+identity. For example, queries for `Age(10), Alive(true)` and
+`Alive(false), Age(20)` share one timing row.
+
+The `Count` column reports consumed query executions. Each eager query method or
+`EntitySet` terminal operation counts once. An iterator counts once when it
+performs query work, whether it is consumed through `next`, `count`, `for_each`,
+or `fold`. Creating and dropping an unused iterator does not increment the
+count. For partially consumed iterators, only work completed before the iterator
+is dropped contributes to the recorded execution time. User callback time in
+`for_each` and `fold` is excluded.
+
+Set algebra retains a query profile when all profiled operands have the same
+query identity, including when a profiled set is combined with an unprofiled
+set. Combining sets with different query identities clears the profile from the
+result so the composed work is not attributed to an arbitrary query.
+
 ## Writing JSON output
 
 `ProfilingContextExt::write_profiling_data()` writes a pretty JSON file to:
@@ -131,6 +152,8 @@ prefix, overwrite). The JSON includes:
 - `execution_statistics`
 - `named_counts`
 - `named_spans`
+- `query_timings`, with query label, observation count, total duration,
+  minimum duration, and maximum duration
 - `computed_statistics`
 
 Example:
--- a/integration-tests/ixa-profiling-tests/Cargo.toml
+++ b/integration-tests/ixa-profiling-tests/Cargo.toml
@@ -11,6 +11,7 @@ authors.workspace = true
 
 [dependencies]
 ixa = { path = "../../", features = ["profiling"] }
+serde.workspace = true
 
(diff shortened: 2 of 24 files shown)


RobertJacobsonCDC:
Performance has actually slightly improved even when the `profiling` feature is not enabled.

Some of my suggestions on #966 for improved user-guide language haven't been acted on yet. I should do that before merging.

RobertJacobsonCDC:
The two benchmark runs in CI are interesting, because the only difference between the two is a cosmetic code formatting change, but they show big swings. Benchmark locally to verify performance claims.

k88hudson-cfa (approved):
LGTM, let's try it out

PR #1044, opened 2026-08-12
fix: support profiling on wasm32

### LINKED ISSUE #1035: Add Wasm support for `profiling` feature
The implementation of the query profiling feature of #1033 does not support the `wasm32` target. The issue is a lack of Wasm support for the `Instant` type—Wasm needs to use `f64` instead. (We ran into this issue before with profiling.)

cdc-as81:
• ## Summary

  Fixes #1035 by enabling profiling on `wasm32-unknown-unknown`.

  ## Changes

  - Add a portable `ProfilingInstant` abstraction:
    - Uses `std::time::Instant` on native targets.
    - Uses `performance.now()` on WASM.
  - Use `ProfilingInstant` across query profiling and execution statistics.
  - Support timing in both browser windows and Web Workers.
  - Preserve the existing `get_high_res_time()` WASM export.
  - Print profiling tables to the browser console.
  - Enable profiling in the WASM integration-test crate.
  - Test query profiling in both the browser and a Web Worker.
  - Add a CI check for the WASM profiling feature combination.

### DIFF
--- a/src/execution_stats.rs
+++ b/src/execution_stats.rs
@@ -8,7 +8,7 @@ use serde_derive::Serialize;
 #[cfg(feature = "profiling")]
 use sysinfo::{Pid, ProcessRefreshKind, ProcessesToUpdate, System};
 #[cfg(target_arch = "wasm32")]
-use wasm_bindgen::prelude::*;
+use wasm_bindgen::{JsCast, JsValue};
 
 #[cfg(feature = "profiling")]
 use crate::profiling::QueryProfiler;
@@ -51,16 +51,6 @@ impl ProfilingInstant {
 
         Duration::from_secs_f64(elapsed_milliseconds / 1000.0)
     }
-}
-
-#[cfg(target_arch = "wasm32")]
-#[wasm_bindgen]
-#[must_use]
-/// Returns a high-resolution monotonic timestamp in milliseconds.
-///
-/// This works in both Window and Web Worker contexts.
-pub fn get_high_res_time() -> f64 {
-    ProfilingInstant::now().0
 }
 
 /// A container struct for computed final statistics.
--- a/src/execution_stats.rs
+++ b/src/execution_stats.rs
@@ -1,6 +1,4 @@
 use std::time::Duration;
-#[cfg(not(target_arch = "wasm32"))]
-use std::time::Instant;
 
 use humantime::format_duration;
 use log::info;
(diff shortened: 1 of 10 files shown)


src/profiling/display.rs
```diff
@@ -165,33 +168,41 @@ pub fn print_formatted_table(rows: &[Vec<String>]) {
...
+    }
+}
+
+#[cfg(feature = "profiling")]
+fn print_formatted_row(row: &[String], col_widths: &[usize]) {
+    let mut line = String::new();
+    for (i, cell) in row.iter().enumerate() {
+        if i == 0 {
+            write!(&mut line, "{:<width$} ", cell, width = col_widths[i] + 1)
+                .expect("writing to a String cannot fail");
```
RobertJacobsonCDC: > writing to a String cannot fail

Then why have the `expect`? Just `unwrap` this. (Same for below.)

src/profiling/mod.rs
```diff
@@ -168,8 +168,42 @@ mod file;
...
+#[cfg(all(feature = "profiling", not(target_arch = "wasm32")))]
+pub type ProfilingInstant = std::time::Instant;
+
+/// A monotonic timestamp used to measure profiling intervals.
+///
+/// Only `now()` and `elapsed()` are part of the portable API shared by native
+/// and WASM targets.
+#[cfg(all(feature = "profiling", target_arch = "wasm32"))]
+#[derive(Clone, Copy, Debug, PartialEq, PartialOrd)]
+pub struct ProfilingInstant(f64);
```
RobertJacobsonCDC: I like this. We should use it everywhere we need an `Instant` type. For example, we should use it for the `Instant`-like fields on `ExecutionProfilingCollector` in `src/execution_stats.rs`. I suspect this would clean up a lot of code in `src/execution_stats.rs`. There might be other places where we use conditional compilation to switch between `f64` and `Instant` types (logging maybe?). Let's do an audit and use this `ProfilingInstant` type in all those places where we really want an `Instant` type.

- I think once we do this, we can fold `get_high_res_time` directly into the `wasm32` implementation of `ProfilingInstant` instead of having it as a separate free function.
- If there are non profiling-related uses in our codebase, should we rename `ProfilingInstant` to just `Instant`? Then whenever we want an `Instant` type we use `ixa`'s `Instant` which is sometimes just `std::time::Instant` and sometimes our custom implementation depending on platform. This is probably subjective personal preference.

RobertJacobsonCDC (changes requested):
The `ProfilingInstant` type is a great idea. Let's expand this PR to use `ProfilingInstant` to clean up more conditionally compiled code across the codebase.

src/profiling/display.rs
cdc-as81: Updated both String formatting calls to use .unwrap() in 0fd0417.

src/profiling/mod.rs
cdc-as81: Addressed in 0fd0417. I audited timer usage and moved ProfilingInstant into execution_stats, then used it for both ExecutionProfilingCollector configurations and the timing tests. The performance.now() lookup now lives in the WASM ProfilingInstant implementation, and profiling re-exports the type. The only remaining direct std::time::Instant uses are native-only benchmarks, so I kept the profiling-specific name. get_high_res_time remains as a thin compatibility wrapper because it is an existing public wasm-bindgen export. Verified with the full Rust check, 593 profiling-enabled unit tests, 28 doctests, and all 5 browser and Web Worker WASM tests.

src/profiling/display.rs
```diff
@@ -165,33 +168,39 @@ pub fn print_formatted_table(rows: &[Vec<String>]) {
...
+    print_table_line(&line);
+}
+
+#[cfg(all(feature = "profiling", target_arch = "wasm32"))]
+fn print_table_line(line: &str) {
+    web_sys::console::log_1(&line.into());
+}
+
+#[cfg(all(feature = "profiling", not(target_arch = "wasm32")))]
+fn print_table_line(line: &str) {
```
RobertJacobsonCDC: I think there is a larger question about how and where we output execution stats and other diagnostic data of which this is a symptom. If this helper function was in the service of a larger strategy for how to output diagnostic data (which it isn't), it would be justified, but as this PR is currently implemented, it is a purely local concern, and I argue it should be inlined.

In general, put each `cfg` on the thing it actually controls.

The original fuses two unrelated conditions into one compound predicate, then duplicates that predicate across two definitions:

```rust
#[cfg(all(feature = "profiling", target_arch = "wasm32"))]
fn print_table_line(line: &str) { /* ... */ }

#[cfg(all(feature = "profiling", not(target_arch = "wasm32")))]
fn print_table_line(line: &str) { /* ... */ }
```

The two conditions aren't at the same level. Whether the function exists depends only on `profiling`. Which body it uses depends only on the platform. Jamming them together forces `profiling` to appear twice and makes the reader diff two `all(...)` clauses to notice they differ by one term.

Separate them:

```rust
#[cfg(feature = "profiling")]
fn print_table_line(line: &str) {
    #[cfg(target_arch = "wasm32")]
    web_sys::console::log_1(&line.into());

    #[cfg(not(target_arch = "wasm32"))]
    println!("{line}");
}
```

Now the outer `cfg` says the function only exists under profiling, and the inner ones pick the platform line. No repetition, and either condition can change without touching the other.

src/execution_stats.rs
```diff
@@ -15,21 +13,54 @@ use wasm_bindgen::prelude::*;
...
+
 #[cfg(target_arch = "wasm32")]
 #[wasm_bindgen]
 #[must_use]
-/// The `wasm` target does not support `std::time::Instant::now()`.
-/// Works in both Window and Web Worker contexts.
+/// Returns a high-resolution monotonic timestamp in milliseconds.
+///
+/// This works in both Window and Web Worker contexts.
 pub fn get_high_res_time() -> f64 {
```
RobertJacobsonCDC: This function isn't needed anymore.

src/execution_stats.rs
```diff
@@ -81,10 +106,7 @@ impl ExecutionProfilingCollector {
     #[must_use]
     pub fn new() -> ExecutionProfilingCollector {
         let process_id = sysinfo::get_current_pid().ok();
-        #[cfg(target_arch = "wasm32")]
-        let now = get_high_res_time();
-        #[cfg(not(target_arch = "wasm32"))]
-        let now = Instant::now();
+        let now = ProfilingInstant::now();
```
RobertJacobsonCDC, line 99: This is so much nicer!

RobertJacobsonCDC (changes requested):
I have just a couple of nits, but otherwise this looked good.

But it surfaced a larger question that I think we should address separately, namely: How do we print out things like execution stats and profiling information? For query profiling, you've chosen to output to `console.log` for Wasm and to the console otherwise, which is a reasonable choice. But elsewhere (e.g. `src/execution_stats.rs`) we only print to the console with `print!`/`println!`.

This is arguably the whole point of having a logging system in the first place—but so far we haven't employed the logging system consistently for this purpose.

In any case, I think that's a bigger question that is out of scope for this PR. I'll file a separate issue.

src/profiling/display.rs
RobertJacobsonCDC: Just to clarify, I think this function should just go away entirely in this PR. Just replace the two call sites with this code:

```rust
#[cfg(target_arch = "wasm32")]
web_sys::console::log_1(&line.into());
#[cfg(not(target_arch = "wasm32"))]
println!("{line}");
```

RobertJacobsonCDC (approved):
Merge when CI passes

